PR #1, opened 2020-09-08
Add minimal buildscript and test programs

drbenmorgan:
As a trivial starting point, this copies across the minimal build setup from Excalibur/emcuda but with the CMake min version bumped to 3.17. This [recent version](https://cmake.org/cmake/help/latest/release/3.17.html) is chosen as it provides very simple configure/use of the CUDA toolkit through the [FindCUDAToolkit](https://cmake.org/cmake/help/v3.17/module/FindCUDAToolkit.html#module:FindCUDAToolkit) module. 

The C++/CUDA standard is fixed at C++14 for now, but can be reviewed, along with any min requirements on CUDA version or compute capability.  

To check that this works a, for now arbitrary, `test` directory holds a couple of programs to exercise CPU/CUDA builds. Those can be tried locally, I haven't thought about what, if any, CI we'll use.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,0 +1,34 @@
+cmake_minimum_required(VERSION 3.17)
+project(Adept
+  VERSION 0.1.0
+  DESCRIPTION "Accelerated demonstrator of electromagnetic Particle Transport"
+  LANGUAGES C CXX CUDA)
+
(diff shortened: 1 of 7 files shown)


agheata (approved):
Good bootstrap!

PR #8, opened 2020-09-16
Added cmake dependencies.

agheata:
This adds VecCore and VecGeom as dependencies for the project. It adds also automatic recording and generation of a script to replay the last build command. The README file was updated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -2,6 +2,10 @@
 # SPDX-License-Identifier: Apache-2.0
 
 cmake_minimum_required(VERSION 3.18)
+
+# Record the command line invoking the cmake command. Replay with recmake_initial.sh. 
+include(cmake/RecordCmdLine.cmake)
+
 project(Adept
   VERSION 0.1.0
   DESCRIPTION "Accelerated demonstrator of electromagnetic Particle Transport"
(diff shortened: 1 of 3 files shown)


agheata:
Looks like the REUSE compliance check works properly, failed before adding the licence/copyright header :-)

graeme-a-stewart:
> 
> 
> Looks like the REUSE compliance check works properly, failed before adding the licence/copyright header :-)

As it should 😄

CMakeLists.txt
```diff
@@ -33,5 +37,20 @@ set(CMAKE_CUDA_EXTENSIONS OFF)
...
 
+#Find VecCore with correct backend
+set(VecCore_VERSION 0.5.2)
+set(VecCore_BACKEND CUDA)
+find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
+message(STATUS "Using VecCore version ${VecCore_VERSION}")
+set(ADEPT_LIBRARIES_EXTERNAL ${ADEPT_LIBRARIES_EXTERNAL} VecCore::VecCore)
+
+# Find VecGeom geometry headers library
+find_package(VecGeom REQUIRED)
```
drbenmorgan, line 48: Should we set a minimum VecGeom version as for VecCore? Could we use the latest 1.1.7 release?

CMakeLists.txt
```diff
@@ -33,5 +37,20 @@ set(CMAKE_CUDA_EXTENSIONS OFF)
...
+message(STATUS "Using VecGeom version ${VecGeom_VERSION}")
+# make sure we import VecGeom architecture flags
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${VECGEOM_CXX_FLAGS}")
+include_directories(${VECGEOM_INCLUDE_DIRS})
```
drbenmorgan, line 52: This can be removed for VecGeom >= 1.1.7, which exports most usage requirements in the `VecGeom::vecgeom` target (so dependent on what we choose for min VecGeom version.

drbenmorgan (commented):
I think it is worth setting the min VecGeom version to 1.1.7 unless there are good reasons to support older releases.

CMakeLists.txt
agheata: Will do...

drbenmorgan (approved):
Looks good to me @agheata!

PR #9, opened 2020-09-23
Added atomic ops type and BlockData concurrent container

agheata:
The templated type adept::Atomic_t dispatches std::atomic interfaces to the appropriate backend (CPU or CUDA). Atomic_t objects adopt pre-allocated memory, using the static function MakeInstanceAt, on a buffer of length sizeof(Atomic_t<Type>). Unit test: test_atomic.
This adds also a new concurrent container adept::BlockData. The block is templated on the data type, allocating the demanded number of elements in a pre-allocated buffer. The number of bytes to be allocated is returned by the static method BlockData<Type>::SizeOfInstance. The method SizeOfAlignAware should be used for allocating multiple blocks in the same buffer. The method NextElement gives write access to the next unused element, and the bracket operator can be used for read access. Unit test: test_track_block
A new concurrent queue adept::mpmc_bounded_queue was added. The queue is templated on the data type, to be created using its static MakeInstanceAt(int nelements, void *buffer). The number of elements must be a power of 2 (for fast index finding in the stored circular buffer) and the buffer size has to use SizeOfInstance (or SizeOfAlignAware for multiple contiguous queues). Unit test: test_queue

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,16 +42,13 @@ set(VecCore_VERSION 0.5.2)
 set(VecCore_BACKEND CUDA)
 find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
-add_definitions(${VecCore_CUDA_DEFINITIONS})
-set(ADEPT_LIBRARIES_EXTERNAL ${ADEPT_LIBRARIES_EXTERNAL} VecCore::VecCore)
 
 # Find VecGeom geometry headers library
 set(VecGeom_VERSION 1.1.7)
 find_package(VecGeom ${VecGeom_VERSION} REQUIRED)
 message(STATUS "Using VecGeom version ${VecGeom_VERSION}")
-# make sure we import VecGeom architecture flags
+# make sure we import VecGeom architecture flags - is this needed?
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${VECGEOM_CXX_FLAGS}")
-set(ADEPT_LIBRARIES_EXTERNAL ${ADEPT_LIBRARIES_EXTERNAL} ${VECGEOM_LIBRARIES})
 
 # Builds...
 add_subdirectory(test)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,7 @@ set(VecCore_VERSION 0.5.2)
 set(VecCore_BACKEND CUDA)
 find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
+add_definitions(${VecCore_CUDA_DEFINITIONS})
 set(ADEPT_LIBRARIES_EXTERNAL ${ADEPT_LIBRARIES_EXTERNAL} VecCore::VecCore)
 
 # Find VecGeom geometry headers library
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -14,6 +14,7 @@ project(Adept
 # - Include needed custom/core modules
 set(CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/cmake" ${CMAKE_MODULE_PATH})
 include(CMakeSettings)
+include(CTest)
 
 # - Core/C++/CUDA build and dependency settings
 # For single-mode generators, default to Optimized with Debug if nothing is specified
--- a/base/inc/AdePT/BlockData.h
+++ b/base/inc/AdePT/BlockData.h
@@ -118,7 +118,10 @@ public:
   {
     // Try to get a hole index if any
     int index = -1;
-    if (fHoles->dequeue(index)) return &fData[index];
+    if (fHoles->dequeue(index)) {
+      fNused--;
+      return &fData[index];
+    }
     index = fNbooked.fetch_add(1);
     if (index >= fCapacity) return nullptr;
     fNused++;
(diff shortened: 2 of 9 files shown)


agheata:
I had to do some minor fixes and force pushed, should be OK now for review.

agheata:
Profiting to add also a concurrent data block in the same PR

CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ set(VecCore_VERSION 0.5.2)
 set(VecCore_BACKEND CUDA)
 find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
+add_definitions(${VecCore_CUDA_DEFINITIONS})
 set(ADEPT_LIBRARIES_EXTERNAL ${ADEPT_LIBRARIES_EXTERNAL} VecCore::VecCore)
```
amadio, line 46: Some things don't work when going through variables. Since VecCore is a required dependency, I suggest using targets directly.

base/inc/AdePT/Utils.h
```diff
@@ -0,0 +1,34 @@
...
+#endif
+
+namespace adept {
+namespace utils {
+/**
+ * @brief Rounds up a value to upper aligned version
+ * @param value Value to round-up
+ */
+template <typename Type>
+VECCORE_ATT_HOST_DEVICE static Type round_up_align(Type value, size_t padding = 32)
```
amadio, line 20: Why the arbitrary default value of 32 for padding?

test/CMakeLists.txt
```diff
@@ -9,3 +10,11 @@ add_executable(fisher_price fisher_price.cpp)
 # Noddy example of particle processing with GPU
 add_executable(cufisher_price cufisher_price.cu)
 target_link_libraries(cufisher_price PRIVATE CUDA::curand)
+# Unit test for atomic ops
+add_executable(test_atomic test_atomic.cu)
+target_include_directories(test_atomic PUBLIC
+  ${VecCore_INCLUDE_DIRS}
```
amadio, line 16: This is not needed, just linking to VecCore::VecCore already takes care of this.

test/CMakeLists.txt
```diff
@@ -9,3 +10,11 @@ add_executable(fisher_price fisher_price.cpp)
...
+  ${VECGEOM_INCLUDE_DIR}
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
+  $<INSTALL_INTERFACE:base/inc>
```
amadio, line 19: Installation directories should probably not end in `base/inc/header.h`, but just `base/header.h`.

test/test_atomic.cu
```diff
@@ -0,0 +1,122 @@
...
+
+#include <iostream>
+#include <cassert>
+#include <AdePT/Atomic.h>
+#include <VecGeom/backend/cuda/Interface.h>
+
+// Example data structure containing several atomics
+struct SomeStruct {
+  adept::Atomic_t<int> var_int{nullptr};
+  adept::Atomic_t<float> var_float{nullptr};
```
amadio: It is quite unintuitive to have `adept::Atomic_t<float>` actually be a pointer type.

test/test_atomic.cu
```diff
@@ -0,0 +1,122 @@
...
+
+  void EmplaceDataAt(void *addr)
+  {
+    char *add_next = (char *)addr;
```
amadio: Why `char*` and not `void*` (or `(u)intptr_t` for that matter)?

agheata:
Added more reviewers, to allow more people comment on the style. I think a minimal reviewing check list is:
* Meaningful description of the PR. Closing #issue if any.
* Meaningful description in the commits. Commits not making sense standalone should be squashed in larger blobs. All commits should compile.
* Coding conventions being followed. Clang formatting should be enforced via an action.
* Comment on the file content, class content, minimal doxygen style for data members and methods in headers
* Comments along the algorithm, at least minimal.
* Presence of unit tests for all new meaningful features
I'm surely missing something, thanks for pointing it out

test/test_atomic.cu
agheata: This is a leftover to be fixed

test/test_atomic.cu
agheata: This is outdated, the method was removed

base/inc/AdePT/Utils.h
agheata: No reason, better no default

drbenmorgan:
Starting to take a look at this, so comments on code itself to come, but I'd recommend adding the following to the build:

1. `include(CTest)` in the top-level `CMakeLists.txt`, can go just after the the `include(CMakeSettings)` line
2. For each of the three test executables, add an `add_test` command for them, e.g. `add_test(NAME test_queue COMMAND test_queue)`

That'll set us up for CI integration and easy testing while developing.

drbenmorgan:
@agheata, with the merge of #12, if you rebase/push, it should apply the formatting changes to your files automatically!

drbenmorgan (approved):
As discussed, LGTM and address noted items through Issues.

base/inc/AdePT/BlockData.h
```diff
@@ -0,0 +1,157 @@
...
+  Type &operator[](const int index) { return fData[index]; }
+
+  /** @brief Dispatch next free element, nullptr if none left */
+  VECCORE_ATT_HOST_DEVICE
+  VECCORE_FORCE_INLINE
+  Type *NextElement()
+  {
+    // Try to get a hole index if any
+    int index = -1;
+    if (fHoles->dequeue(index)) return &fData[index];
```
WitekPokorski, line 121: Bug here! You need to increase fNused before returning.

PR #12, opened 2020-10-01
Add clang-format file and workflow

### LINKED ISSUE #10: adding .clang_format
We should add a .clang_format style and enforce usage of clang_format as action. So we should decide first on the style and minimum clang-format version. We can use the style we use in VecGeom, with 2 spaces indentation, column limit of 120, replace tab with 2 spaces, .... We also have some scripts allowing to check/apply locally clang-format to all files touched in a branch.

Any opinions on the style/version?

drbenmorgan:
Initial clang-format import using VecGeom's style. Add workflow to check/apply changes on pushes and pull requests.

Closes #10

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,0 +1,92 @@
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+# This would be nice to have but seems to also (mis)align function parameters
+AlignConsecutiveDeclarations: false
+AlignEscapedNewlinesLeft: true
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: false
+# This option is "deprecated and is retained for backwards compatibility."
+# AlwaysBreakAfterDefinitionReturnType: None
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: false
+AlwaysBreakTemplateDeclarations: true
+BinPackArguments: true
+BinPackParameters: true
+BraceWrapping:
+  AfterClass:      false
+  AfterControlStatement: false
+  AfterEnum:       false
+  AfterFunction:   true
+  AfterNamespace:  false
+  AfterObjCDeclaration: false
+  AfterStruct:     false
+  AfterUnion:      false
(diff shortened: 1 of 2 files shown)


.clang-format
```diff
@@ -0,0 +1,94 @@
+# SPDX-FileCopyrightText: 2020 CERN
+# SPDX-License-Identifier: Apache-2.0
+---
+Language:        Cpp
+# BasedOnStyle:  LLVM
+AccessModifierOffset: -2
+AlignAfterOpenBracket: Align
+AlignConsecutiveAssignments: true
+# This would be nice to have but seems to also (mis)align function parameters
```
graeme-a-stewart, line 9: This comment leads me to wonder if an updated clang-format would work better (I'm not 100% sure what the problem aluded to is). Maybe @agheata remembers?

.clang-format
```diff
@@ -0,0 +1,94 @@
...
+AlignEscapedNewlinesLeft: true
+AlignOperands:   true
+AlignTrailingComments: true
+AllowAllParametersOfDeclarationOnNextLine: true
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Inline
+AllowShortIfStatementsOnASingleLine: true
+AllowShortLoopsOnASingleLine: false
+# This option is "deprecated and is retained for backwards compatibility."
```
graeme-a-stewart, line 20: Remove then...

.clang-format
```diff
@@ -0,0 +1,94 @@
...
+SpacesBeforeTrailingComments: 1
+SpacesInAngles:  false
+SpacesInContainerLiterals: true
+SpacesInCStyleCastParentheses: false
+SpacesInParentheses: false
+SpacesInSquareBrackets: false
+Standard:        Cpp11
+TabWidth:        2
+UseTab:          Never
+...
```
graeme-a-stewart, line 93: Does this elipsis serve a purpose?

graeme-a-stewart (changes requested):
Looks grand, just a few comments to quickly discuss before merge

.clang-format
drbenmorgan: This is the YAML document delimeter (that goes with the `---` at the top of the document).

drbenmorgan:
One thing I just noted is that the action won't consider CUDA files per this line in the entry point:

https://github.com/HSF/clang-format-action/blob/802a9bf553bed6e1deea1ae330667424526c4eec/entrypoint.sh#L8

Guess we should put a PR in there to either always consider `.cu` files or an option.

graeme-a-stewart (approved):

graeme-a-stewart:
> Guess we should put a PR in there to either always consider `.cu` files or an option.

Good point. We should take `.cuh` as well.

drbenmorgan:
Proposed in https://github.com/HSF/clang-format-action/pull/1

PR #17, opened 2020-10-02
Adding the Alpaka version of FisherPrice

shefmarkh:
Hello,

This PR includes the current version of the FisherPrice code reworked to interface with Alpaka.

At the moment I put authors of all the c++ headers as Dvaide/myself. I think some/all of the classes might have contributions from @drbenmorgan - should I therefore add you as an author to some/all of them too? (also I don't know if the versions I started from have any other authors I'm not aware of to add).

This is a WIP for now, but starting it so as to coordinate with code from @drbenmorgan. In particular I added doxygen (as requested in the contribution requirements) to the c++ classes I picked up from FisherPrice. These classes are the same as the ones in the CUDA/c++ branch, so presumably we don't want people independently spending time to add doxygen. This also brings a general question about the fact we are going to end up with same code once for alpaka and once for the version that ifdefs between cuda and c++. I'm not sure the best way to handle that going forward, but if FisherPrice ends up with a lot of code in the future it could become painful to maintain (I suppose its not so hard right now with little code).

The requirements also specify we should provide Unit Tests for any new code in a PR - again there would be the same issue of presumably not wanting to design this twice independently for the cuda/c++ and alpaka versions of code. 

I suspect some of the things we have currently used (e.g. particleStack.h) already have more advanced equivalents in this repository (from @agheata), so we could setup issues to migrate where relevant.

Cheers,

Mark

### DIFF
--- a/test/FisherPrice_Alpaka/CMakeLists.txt
+++ b/test/FisherPrice_Alpaka/CMakeLists.txt
@@ -1,5 +1,5 @@
-// SPDX-FileCopyrightText: 2020 CERN
-// SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2020 CERN
+# SPDX-License-Identifier: Apache-2.0
 
 #We must specify that we want to use alpaka
 find_package(alpaka)
--- a/test/FisherPrice_Alpaka/CMakeLists.txt
+++ b/test/FisherPrice_Alpaka/CMakeLists.txt
@@ -1,7 +1,11 @@
-# This makes sure cmake knows where to find alpaka
-set(CMAKE_PREFIX_PATH "/install")
-# Then we must specify that we want to use alpaka
-find_package(alpaka REQUIRED)
-alpaka_add_executable(FisherPriceAlpaka alpaka_fisher_price.cu)
-target_link_libraries(FisherPriceAlpaka PUBLIC alpaka::alpaka)
+#We must specify that we want to use alpaka
+find_package(alpaka)
+
+if(NOT alpaka_FOUND)
+  message(STATUS "No alpaka install found, disabling build of FisherPrice_Alpaka")
+  return()
+else()  
+  alpaka_add_executable(FisherPriceAlpaka alpaka_fisher_price.cu)
+  target_link_libraries(FisherPriceAlpaka PUBLIC alpaka::alpaka)
+endif()
 
--- a/test/FisherPrice_Alpaka/CMakeLists.txt
+++ b/test/FisherPrice_Alpaka/CMakeLists.txt
@@ -1,3 +1,6 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
 #We must specify that we want to use alpaka
 find_package(alpaka)
 
--- a/test/FisherPrice_Alpaka/CMakeLists.txt
+++ b/test/FisherPrice_Alpaka/CMakeLists.txt
@@ -1,0 +1,7 @@
+# This makes sure cmake knows where to find alpaka
+set(CMAKE_PREFIX_PATH "/install")
+# Then we must specify that we want to use alpaka
+find_package(alpaka REQUIRED)
+alpaka_add_executable(FisherPriceAlpaka alpaka_fisher_price.cu)
+target_link_libraries(FisherPriceAlpaka PUBLIC alpaka::alpaka)
+
--- a/test/FisherPrice_Alpaka/alpaka_fisher_price.cu
+++ b/test/FisherPrice_Alpaka/alpaka_fisher_price.cu
@@ -3,9 +3,97 @@
 
 //This is the main include file to give us all of the alpaka classes, functions etc.
 #include <alpaka/alpaka.hpp>
+
+#include <iostream>
+
 #include "particleProcessor.h"
+
+// This processes one event on the device. A list of particles is passed through
+// For each particle we call process(*partList, iTh) 
+// process will then work the same was as in the C++ version
+struct processEvent {
+  template<typename Acc> ALPAKA_FN_ACC void operator() (Acc const& acc, part *partList, int* steps) const{
+
+    using namespace alpaka;
+    // iTh is the thread number we use this throughout 
+    uint32_t iTh = idx::getIdx<Grid, Threads>(acc)[0];
+    //dummy sensitive, not used yet 
+    sensitive SD(400.,500.);  
+
+    //Create an alpaka random generator using a seed of 1984
+    auto generator = rand::generator::createDefault(acc,1984,iTh); 
+    
+    // Here we split threads that start immediatly, from threads that will have to wait until the shower is under way 
+    particleProcessor particleProcessor;
+    int ns=particleProcessor.processParticle(acc,partList, iTh, generator, SD); 
+    steps[iTh] = ns;
+  }
+};
+
+
 
 int main()
 {
+
+  using namespace alpaka; //alpaka functionality all lives in this namespace
+
+  using Dim = dim::DimInt<1>;
+  using Idx = uint32_t;
+
+  //Define the alpaka accelerator to be Nvidia GPU
+  using Acc = acc::AccGpuCudaRt<Dim,Idx>;
+
+  //Get the first device available of type GPU (i.e should be our sole GPU)/device
+  auto const device = pltf::getDevByIdx<Acc>(0u);
+  // Create a device for host for memory allocation, using the first CPU available
+  auto devHost = pltf::getDevByIdx<dev::DevCpu>(0u);
+
+  //Allocate memory on both the host and device for part objects and numbers of steps
+  uint32_t NPART = 200;
+  vec::Vec<Dim, Idx> bufferExtent{NPART};
+  auto d_event = mem::buf::alloc<part,Idx>(device,bufferExtent);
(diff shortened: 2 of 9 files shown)


test/FisherPrice_Alpaka/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# This makes sure cmake knows where to find alpaka
+set(CMAKE_PREFIX_PATH "/install")
```
drbenmorgan, line 2: This should be removed and instead the README updated with instructions on pointing to the Alpaca install.

test/FisherPrice_Alpaka/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
...
+# Then we must specify that we want to use alpaka
+find_package(alpaka REQUIRED)
```
drbenmorgan, line 4:
```suggestion
find_package(alpaka)

if(NOT alpaka_FOUND)
  message(STATUS "No alpaka install found, disabling build of FisherPrice_Alpaka")
  return()
endif()
```

As alpaka won't be a hard dependency to begin with, the above is the simplest way to enable/disable as needed. We can add a proper option to enforce later.

drbenmorgan (changes requested):
Just a couple of changes requested!

drbenmorgan:
> These classes are the same as the ones in the CUDA/c++ branch, so presumably we don't want people independently spending time to add doxygen. This also brings a general question about the fact we are going to end up with same code once for alpaka and once for the version that ifdefs between cuda and c++. I'm not sure the best way to handle that going forward, but if FisherPrice ends up with a lot of code in the future it could become painful to maintain (I suppose its not so hard right now with little code).

That's fine for now, @WitekPokorski's branch on the CUDA fisher price really supersedes the initial work as it can use the proper "track block" data structure.

drbenmorgan:
BTW, don't worry about the clang-format check for now, I'll take a look at why it's refusing to push the changes back.

test/FisherPrice_Alpaka/alpaka_fisher_price.cu
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+//This is the main include file to give us all of the alpaka classes, functions etc.
+#include <alpaka/alpaka.hpp>
```
agheata, line 5: We should ideally converge to a single executable that, depending on a configuration option, will be compiled against alpaka or cuda. Includes like this would the need to be handled by a higher-level one that will handle the ifdefs

test/FisherPrice_Alpaka/particleProcessor.h
```diff
@@ -0,0 +1,145 @@
...
+#include "part.h"
+#include "particleStack.h"
+#include "sensitive.h"
+#include <alpaka/alpaka.hpp>
+
+class particleProcessor {
+
+  public:
+    /** @brief A minimal constructor */
+    ALPAKA_FN_ACC particleProcessor (){}
```
agheata, line 23: Besides "abstracting" the apaka-dependent headers, we may want also to abstract macros like ALPAKA_FN_ACC to avoid depending on specific back-ends

agheata:
As a general remark, I would be in favor of trying to abstract out the Alpaka dependencies (macros, allocations, ...) using home-made macros that would allow having single source, at least for the algorithmic part. Not to be done now, but to think about how to do it in future.

shefmarkh:
> As a general remark, I would be in favor of trying to abstract out the Alpaka dependencies (macros, allocations, ...) using home-made macros that would allow having single source, at least for the algorithmic part. Not to be done now, but to think about how to do it in future.

I agree that we don't want the same algorithmic code maintained in parallel multiple times if it could be avoided (this would get even worse once we consider looking at other technologies such as OneAPI).

Cheers,

Mark

shefmarkh:
> BTW, don't worry about the clang-format check for now, I'll take a look at why it's refusing to push the changes back.

is it a permissions issue on my fork perhaps?

drbenmorgan:
O.k., definitely something up with the formatting action, on PR #20 I submitted an initial commit that had bad format, which was fixed. Then another commit with bad formatting, and that's fixed, but there's then no status attached to the fix commit. Not the same as your error, but somethings definitely not right. To be investigated further

test/FisherPrice_Alpaka/CMakeLists.txt
shefmarkh: Is there a reason not to use the following setup? If alpaka is installed in the standard default area it will build it, otherwise it prints a message (if either the Alpaka is not in the default area or if the find_package command fails) instructing the user to check the README for the instructions about alpaka. Everything else will build ok in either scenario.

Cheers,

Mark

# This makes sure cmake knows where to find alpaka, if it was installed here
if (EXISTS "/install/include/alpaka/")
  set(CMAKE_PREFIX_PATH "/install")
else()
  message(STATUS "Could not find /install/include/alpaka, disabling build of FisherPrice_Alpaka. If you would like to build this please check the README for details of how to install Alpaka.")
endif()
# Then we must specify that we want to use alpaka
find_package(alpaka)

if(NOT alpaka_FOUND)
  message(STATUS "No alpaka install found, disabling build of FisherPrice_Alpaka. If you would like to build this please check the README for details of how to install Alpaka.")
  return()
else()  
  alpaka_add_executable(FisherPriceAlpaka alpaka_fisher_price.cu)
  target_link_libraries(FisherPriceAlpaka PUBLIC alpaka::alpaka)
endif()

test/FisherPrice_Alpaka/CMakeLists.txt
shefmarkh: I proposed a modified version of this in the previous comment.

Cheers,

Mark

shefmarkh:
> BTW, don't worry about the clang-format check for now, I'll take a look at why it's refusing to push the changes back.

I'm a bit confused about this....I actually have clang-format installed in my copy of Visual Studio Code. So in principle the code should be correctly formatted, but clearly that has not happened (reading the documentation again I can't seem to find what I have missed). Anyway if it would reformat via GitHub that would do the same job anyway (once it works).

Cheers,

Mark

test/FisherPrice_Alpaka/CMakeLists.txt
drbenmorgan: I'd rather we didn't introduce a hard coded path that's specific to one machine, even if only temporary. It's easy enough to set `CMAKE_PREFIX_PATH` local via command line, or as an environment variable.

drbenmorgan:
> I'm a bit confused about this....I actually have clang-format installed in my copy of Visual Studio Code. So in principle the code should be correctly formatted, but clearly that has not happened (reading the documentation again I can't seem to find what I have missed). Anyway if it would reformat via GitHub that would do the same job anyway (once it works).

O.k., that is weird! For this PR, what we can try is for you to squash the commits into one on your local clone's branch, then force push that to your fork. Assuming things are formatted correctly, it should succeed or at least work in one go. Let's leave that until we're ready for merge though.

test/FisherPrice_Alpaka/particleProcessor.h
drbenmorgan: @agheta, this is good point and something we can/should address through Issue #24?

drbenmorgan:
Thanks @shefmarkh, the latest commits have triggered the auto clang-format, although why I don't know given you'd already formatted. The one remaining fix is to add the copyright in:

test/FisherPrice_Alpaka/CMakeLists.txt

which can be copy-pasted from that at the top of the other CMake scripts. Then I think we can merge.

shefmarkh:
Changing back to WIP - just spotted I forgot to fill in the code in the main function to actually run the simulation.

Cheers,

Mark

shefmarkh:
Hello @drbenmorgan 

Each time I push I get told:

it push 
warning: push.default is unset; its implicit value is changing in
Git 2.0 from 'matching' to 'simple'. To squelch this message
and maintain the current behavior after the default changes, use:

  git config --global push.default matching

To squelch this message and adopt the new behavior now, use:

  git config --global push.default simple

See 'git help config' and search for 'push.default' for further information.
(the 'simple' mode was introduced in Git 1.7.11. Use the similar mode
'current' instead of 'simple' if you sometimes use older versions of Git)

To git@github.com:shefmarkh/AdePT.git
 ! [rejected]        mhodgkin_addSimpleSimAlpaka -> mhodgkin_addSimpleSimAlpaka (fetch first)
error: failed to push some refs to 'git@github.com:shefmarkh/AdePT.git'
hint: Updates were rejected because the remote contains work that you do
hint: not have locally. This is usually caused by another repository pushing
hint: to the same ref. You may want to first merge the remote changes (e.g.,
hint: 'git pull') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

I think the clang-format here is making some subtle change, such that I have to merge it back into my branch locally before I can push? Sometimes this has caused many conflicts (though this time it did not).

Cheers,

Mark

drbenmorgan:
Hi @shefmarkh, for some reason, when you committed last the action actually ran and applied fixes. I think that if you `git pull --rebase` before re-pushing it should apply the changes and fix things (modulo any conflicts).

shefmarkh:
This is ready to be merged, unless someone has any further suggestions for improvements.

Cheers,

Mark

drbenmorgan (approved):

PR #19, opened 2020-10-02
adding cuda version of fisher_price example

WitekPokorski:
This pull request adds the CUDA version of the fisher_price example. The CUDA kernel consists of the 'process' (energy loss and pair production) run for each particle/step. For each iteration the 'main' schedules as many CUDA threads as particles in flight. New particle which are created during the pair-production process are pushed into the common BlockData container. Particles which die (E=0) are released from the container and the slots are used by new ones.

### DIFF
--- a/test/cufisher_price.cu
+++ b/test/cufisher_price.cu
@@ -62,7 +62,7 @@ __global__ void process(adept::BlockData<MyTrack> *block, Scoring *scor, curandS
 
     // here I need to create a new particle
     auto secondary_track = block->NextElement();
-    assert (secondary_track != nullptr && "No slot available for secondary track" );
+    assert(secondary_track != nullptr && "No slot available for secondary track");
     secondary_track->energy = eloss;
 
     // increase the counter of secondaries
--- a/test/cufisher_price.cu
+++ b/test/cufisher_price.cu
@@ -2,12 +2,132 @@
 // SPDX-License-Identifier: Apache-2.0
 
 #include <curand.h>
+#include <curand_kernel.h>
+#include <iostream>
+
+#include <AdePT/BlockData.h>
+
+struct MyTrack {
+  int index{0};
+  int pdg{0};
+  double energy{0};
+  double pos[3]{0};
+  double dir[3]{0};
+  bool flag1;
+  bool flag2;
+};
+
+struct Scoring {
+  adept::Atomic_t<int> secondaries;
+  adept::Atomic_t<float> totalEnergyLoss;
+
+  VECCORE_ATT_HOST_DEVICE
+  Scoring() {}
+
+  VECCORE_ATT_HOST_DEVICE
+  static Scoring *MakeInstanceAt(void *addr)
+  {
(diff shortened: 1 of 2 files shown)


test/cufisher_price.cu
```diff
@@ -2,12 +2,132 @@
...
+
+  // call the 'process'
+  if (r < 0.5f) {
+    // energy loss
+    float eloss = 0.2f * (*block)[particle_index].energy;
+    scor->totalEnergyLoss.fetch_add(eloss < 0.001f ? (*block)[particle_index].energy : eloss);
+    (*block)[particle_index].energy = (eloss < 0.001f ? 0.0f : ((*block)[particle_index].energy - eloss));
+
+    // if particle dies (E=0) release the slot
+    if ((*block)[particle_index].energy == 0) block->ReleaseElement(particle_index);
```
drbenmorgan, line 57:
```suggestion
    if ((*block)[particle_index].energy < 0.001f) block->ReleaseElement(particle_index);
```

I think this guarantees the release, given the cutoff above without worrying about exact floating point comparison.

test/cufisher_price.cu
```diff
@@ -2,12 +2,132 @@
...
+  scor->totalEnergyLoss = 0;
+
+  // Allocate a block of tracks with capacity larger than the total number of spawned threads
+  // Note that if we want to allocate several consecutive block in a buffer, we have to use
+  // Block_t::SizeOfAlignAware rather than SizeOfInstance to get the space needed per block
+  using Block_t    = adept::BlockData<MyTrack>;
+  size_t blocksize = Block_t::SizeOfInstance(capacity);
+  char *buffer2    = nullptr;
+  cudaMallocManaged(&buffer2, blocksize);
+  auto block = adept::BlockData<MyTrack>::MakeInstanceAt(capacity, buffer2);
```
drbenmorgan, line 108:
```suggestion
  auto block = Block_t::MakeInstanceAt(capacity, buffer2);
```

drbenmorgan (changes requested):

test/cufisher_price.cu
```diff
@@ -2,12 +2,132 @@
...
+    // if particle dies (E=0) release the slot
+    if ((*block)[particle_index].energy == 0) block->ReleaseElement(particle_index);
+  } else {
+    // pair production
+    float eloss = 0.5f * (*block)[particle_index].energy;
+    (*block)[particle_index].energy -= eloss;
+
+    // here I need to create a new particle
+    auto secondary_track = block->NextElement();
+    if (!secondary_track) return;
```
agheata, line 65:
```suggestion
    assert (secondary_track != nullptr && "No slot available for secondary track" );
```
This is a bad situation and the program should crash, otherwise the energy loss will not add up to the primary energy

drbenmorgan (approved):
Fine for me! @agheata, if you have no further items, go ahead and merge!

agheata (approved):

PR #20, opened 2020-10-05
Test only: check clang-format action

drbenmorgan:
This is an experimental PR to see what happens to the clang-format action. PR #17 seems to not get the fixes applied and it's not clear why, when #19 worked.


drbenmorgan:
Closing this as superseded by #29

PR #23, opened 2020-10-06
Integrate CopCore build with AdePT

drbenmorgan:
This adds the minimum set of changes needed to add `CopCore` as a subproject of AdePT. It's a simple `add_subdirectory` and `CopCore` supplies the needed `CopCore::CopCore` imported target. That should make use transparent if/when it's split out into its own repo.

A trivial compile/link is added to check behaviour, but can be removed once we have actual AdePT code using it.

I've also updated the contribution guide to recommend/require that changes under `base/inc/CopCore` are made in dedicated commits without changes from anywhere else in the repo. There's no enforcement of that yet through actions.


test/test_copcore_link.cpp
```diff
@@ -0,0 +1,6 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+#include <CopCore/CopCore.h>
+
+int main() {}
```
graeme-a-stewart, line 6: Tiny point here - needs a newline

graeme-a-stewart (changes requested):

graeme-a-stewart (approved):

agheata (commented):
Looks good

PR #29, opened 2020-10-08
Remove format CI action and document git clang-format

drbenmorgan:
Use of the GitHub Action for automatic clang-formatting was leading to unexpected or confusing behaviour likely caused by it pushing back to an active PR.

Remove action from workflow to avoid confusion. Update contributing guide with instruction on use of the `git-clang-format` script supplied by installs of `clang-format`.


agheata:
So how do we enforce clang formatting if we have no github action for it?

agheata:
> Edit: could also get the action to output a patch to apply and comment with "run this command to apply the formatting changes".

:+1:

agheata (approved):

PR #31, opened 2020-10-21
Quieten CMake warnings when not finding Alpaka

drbenmorgan:
This addresses the question in one of our catchups about whether Alpaka is required or not. It isn't, but CMake will moan if it's not found. This just quietens that warning, but the behaviour is otherwise unchanged

### DIFF
--- a/test/FisherPrice_Alpaka/CMakeLists.txt
+++ b/test/FisherPrice_Alpaka/CMakeLists.txt
@@ -2,12 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 #We must specify that we want to use alpaka
-find_package(alpaka)
+find_package(alpaka QUIET)
 
 if(NOT alpaka_FOUND)
   message(STATUS "No alpaka install found, disabling build of FisherPrice_Alpaka")
   return()
-else()  
+else()
   alpaka_add_executable(FisherPriceAlpaka alpaka_fisher_price.cu)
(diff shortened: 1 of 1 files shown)


test/FisherPrice_Alpaka/CMakeLists.txt
```diff
@@ -2,12 +2,12 @@
 # SPDX-License-Identifier: Apache-2.0
 
 #We must specify that we want to use alpaka
-find_package(alpaka)
+find_package(alpaka QUIET)
 
 if(NOT alpaka_FOUND)
   message(STATUS "No alpaka install found, disabling build of FisherPrice_Alpaka")
```
agheata, line 8: I would profit and replace the message with one more generic, given that likely we're going to build more than FisherPrice_Alpaka. Maybe: "Alpaka installation not found: support disabled."? But this can be done as well later.

drbenmorgan: I think that can be left to later, the message is clear for this context. Any later general support for Alpaka would result in the `find_package` call being moved anyway.

shefmarkh:
Looks fine to me, less verbose is generally nice for log files. Even with this change there is a message to inform users that the alpaka based code won't build.

Cheers,

Mark

WitekPokorski:
Looks good to me.

WitekPokorski (approved):

PR #32, opened 2020-10-22
Cuda fisher

WitekPokorski:
This is to start creating sub-directories for the different parts of the code and populating them with some embryonic code. Examples are now split from the tests, and there are directories (to be populated) for physics and tracking.


examples/FisherPrice_cu/cufisher_price_v2.cu
```diff
@@ -0,0 +1,231 @@
...
+    // check if the particle is still alive (E>0)
+    if ((*block)[particle_index].energy == 0) return;
+
+    // pair production
+    float eloss = 0.5f * (*block)[particle_index].energy;
+    (*block)[particle_index].energy -= eloss;
+
+    // here I need to create a new particle
+    auto secondary_track = block->NextElement();
+    assert(secondary_track != nullptr && "No slot available for secondary track");
```
drbenmorgan, line 109: Should we have the creation attempt+assertion prior to the calculations? That way, if we replace the assertion with a later "couldn't process this" like error/return, we don't change any physical state of the particle.

examples/FisherPrice_cu/cufisher_price_v2.cu
```diff
@@ -0,0 +1,231 @@
...
+  track2->energy = 30.0f;
+
+  //
+
+  const int num_streams = 8;
+  cudaStream_t streams[num_streams];
+
+  cudaStreamCreate(&streams[0]);
+  cudaStreamCreate(&streams[1]);
+  cudaStreamCreate(&streams[2]);
```
drbenmorgan, line 187: Should these be commented out, or the version using streams be in a "v3" example?

examples/FisherPrice_cu/cufisher_price_v2.cu
```diff
@@ -0,0 +1,231 @@
...
+
+//
+
+int main()
+{
+
+  curandState_t *state;
+  cudaMalloc((void **)&state, sizeof(curandState_t));
+  init<<<1, 1>>>(state);
+  cudaDeviceSynchronize();
```
drbenmorgan, line 132: I think this is fine for now, but we should think about a general way to handle random numbers (e.g. one state per thread as per https://docs.nvidia.com/cuda/curand/device-api-overview.html#performance-notes).

drbenmorgan (approved):

drbenmorgan:
This is all fine for me, so unless @agheata has additional comments, I think it's good to merge!

examples/FisherPrice_cu/cufisher_price_v2.cu
WitekPokorski: Yes indeed. I have changed the order to be more consistent.

examples/FisherPrice_cu/cufisher_price_v2.cu
WitekPokorski: Yes, a left over from a different 'experiment'. Removed for now.

examples/FisherPrice_cu/cufisher_price_v2.cu
WitekPokorski: Sure. This is very basic. One of the components we should implement in our 'base tools' is some proper RND generator interface/infrastructure/kernel.

PR #36, opened 2020-11-06
Update README and fix build on CentOS

hahnjo:


graeme-a-stewart (approved):
Wow, so timely - I just bumped into exactly the same problem on titan-x!

PR #37, opened 2020-11-09
Fix execution of cufisher_price

hahnjo:


base/inc/AdePT/Atomic.h
```diff
@@ -83,7 +83,13 @@ struct AtomicBase_t {
 
   /** @brief Atomically loads and returns the current value of the atomic variable. */
   VECCORE_ATT_DEVICE
-  Type load() const { return fData; }
+  Type load() const
+  {
+    // There is no atomic load on CUDA. Issue a memory fence to get the required
+    // semantics and make sure that the value is really loaded from memory.
+    __threadfence();
```
agheata, line 90: Indeed, I overlooked this, such protection is needed

examples/FisherPrice_cu/cufisher_price.cu
```diff
@@ -33,32 +33,35 @@ struct Scoring {
 };
 
 // kernel function that does energy loss or pair production
-__global__ void process(adept::BlockData<MyTrack> *block, Scoring *scor, curandState_t *states)
+__global__ void process(adept::BlockData<MyTrack> *block, Scoring *scor, curandState_t *states, int max_index)
 {
   int particle_index = blockIdx.x * blockDim.x + threadIdx.x;
 
   // check if you are not outside the used block
-  if (particle_index > block->GetNused() + block->GetNholes()) return;
```
agheata, line 41: Indeed this was not correct, in the old approach should read: particle index >= ...

agheata (approved):

hahnjo:
When revisiting the code, I noticed that `cufisher_price` is still faulty:
1. As @agheata commented, the condition should be `particle_index >= max_index` (greater equals instead of greater than). Not sure why the current code works, do we initialized all tracks with `energy = 0` initially? Of course could be just luck...
2. My fix regarding the race condition is only partly correct: It's still possible that a thread processes a particle that is brought to live in one of the holes. I think this can only be solved by having at least one queue, cf. the other examples.

I think we should remove this example to avoid future confusion (took me around 30 minutes to recover the above).

agheata:
> I think we should remove this example to avoid future confusion (took me around 30 minutes to recover the above).
👍 this is superseded by the others, no point in maintaining it, especially if it has bugs

PR #40, opened 2020-11-17
MParray

agheata:
Introduced a fast concurrent multi-producer container for fast selection of track indices in kernel functions. Adapted the test_launcher example to use this instead of the previous MPMC queue.


WitekPokorski (commented):
Don't we want to have the equivalent of 'dequeue' method?

WitekPokorski:
So, if I understand well, you can remove elements from that array, right? You can only clear the entire arrray?

WitekPokorski (approved):

PR #45, opened 2020-11-27
adding example1 demonstrating workflow with pipelining kernels for processes

WitekPokorski:
This is a new example where some basic infrastructure is implemented for handling a set of processes inheriting from the same base class. A class process_list manages all available processes. The files (even if only example ones, to be replaced in the future) have been put in the appropriate directories to exercise the directory structure. MParrays are used for queues for each of the process. The main program implements some basic workflow checking first for the interactions length (dummy), selecting the appropriate queues for the particles and then calling the pipeline of processes. 
This example is an attempt of some basic 'architecture' of stepping sequence with a split between CPU and GPU calls.

### DIFF
--- a/examples/Example1/example1.cu
+++ b/examples/Example1/example1.cu
@@ -1,0 +1,188 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+#include <curand.h>
+#include <curand_kernel.h>
+#include <iostream>
+#include <iomanip>
+#include <stdio.h>
+
+#include <AdePT/BlockData.h>
+
+#include "process.h"
+#include "process_list.h"
+#include "pair_production.h"
+#include "energy_loss.h"
+
+#include "track.h"
+
+#include <AdePT/MParray.h>
+
+// some simple scoring
+struct Scoring {
+  adept::Atomic_t<int> secondaries;
+  adept::Atomic_t<float> totalEnergyLoss;
+
+  VECCORE_ATT_HOST_DEVICE
+  Scoring() {}
+
+  VECCORE_ATT_HOST_DEVICE
+  static Scoring *MakeInstanceAt(void *addr)
+  {
+    Scoring *obj = new (addr) Scoring();
+    return obj;
+  }
+};
+
+// kernel select processes based on interaction lenght and put particles in the appropriate queues
+__global__ void DefinePhysicalStepLength(adept::BlockData<track> *block, process_list** proclist, adept::MParray **queues, curandState_t *states)
+{
+  int n = block->GetNused() + block->GetNholes();
+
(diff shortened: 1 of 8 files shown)


tracking/inc/track.h
```diff
@@ -0,0 +1,20 @@
...
+    double dir[3]{1};
+    int mother_index{0};
+    TrackStatus status{alive}; 
+    int current_process{0};
+    float interaction_length{FLT_MAX};
+    float energy_loss{0}; // primitive version of scoring 
+    int number_of_secondaries{0}; // primitive version of scoring
+  };
+
+#endif
```
agheata, line 20: Some "no newline at end of file" indicate that clang-format may not be applied.

agheata (changes requested):
Some licence headers seem to be missing and clang-format may be needed

examples/Example1/example1.cu
```diff
@@ -0,0 +1,188 @@
...
+  int n = block->GetNused() + block->GetNholes();
+
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < n; i += blockDim.x * gridDim.x) {
+    
+    // skip particles that are already dead
+    if ((*block)[i].status == dead) continue;
+
+    (*proclist)->GetPhysicsInteractionLength(i, block, states); // return value (if step limited by physics or geometry) not used for the moment
+    // now, I know which process wins, so I add the particle to the appropriate queue
+    queues[(*block)[i].current_process]->push_back(i);
```
hahnjo, line 49: right now, I think we always have a process in this simple example. Just curious, how would this extend to transportation?

tracking/inc/track.h
hahnjo: Yes, some other whitespace also doesn't look correctly formatted. For `.cu` files, you have to check if your `clang-format` is recent enough to recognize this extensions. My patch for that is a few years old already, but not sure if it already propagated to all distributions...

hahnjo (commented):
LGTM, one question about future extensions inline

examples/Example1/example1.cu
WitekPokorski: transportation is just yet another process (it gives the distance to the next boundary as 'interaction lenght'. and it's a 'along step process' that always takes place (you can have several along step process). The distance to the boundary is used to check if the step is limited by physics or geometry .

tracking/inc/track.h
agheata: Looks like the system does not check it any more. @drbenmorgan looks clang formatting  was removed as action, right?

examples/Example1/example1.cu
agheata: This is of course how it is done in Geant4, in our case we are not bound to implement as a process having the same interface, but reduction to get the minimum proposed step (geometry vs. a given physics process) has to be done at some point for every step.

agheata (approved):

PR #47, opened 2020-11-30
[copcore] Porting physics constants and system of units from GeantV.

agheata:
No need to reinvent the wheel for units and physics constants.

### DIFF
--- a/base/inc/CopCore/include/CopCore/SystemOfUnits.h
+++ b/base/inc/CopCore/include/CopCore/SystemOfUnits.h
@@ -1,0 +1,309 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @brief   System of units.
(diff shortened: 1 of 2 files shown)


agheata:
I'm aware that we'll need also the 'float' version for these constants

base/inc/CopCore/include/CopCore/SystemOfUnits.h
```diff
@@ -0,0 +1,309 @@
...
+ * \li  luminous intensity      (candela)
+ * \li  radian                  (radian)
+ * \li  steradian               (steradian)
+ *
+ * It's a non exhaustive list of derived and pratical units (i.e. mostly the SI
+ * units).
+ *
+ * The value of \f$ \pi \f$ (kPi) is defined here as it is needed for radian to degree
+ * conversion. The other physical constants are defined in the header file :
+ * Geant/PhysicalConstants.h
```
hahnjo, line 30: `CopCore/PhysicalConstants.h`

PR #49, opened 2020-12-02
LoopNavigator Class and use rays from BlockData<T> containers

antoniopetre:
I added two new files: ```LoopNavigator.h``` and ```kernels.h```.

I refactored the navigation methods as a separate class ```LoopNavigator```. These methods were defined before in ```Raytracer.cpp```.

The rays are used now from ```BlockData<T>``` containers. ```kernels.h``` contains the function ```generateRays``` which is responsible to allocate slots for the ```BlockData```.

### DIFF
--- a/examples/Raytracer_Benchmark/RaytraceBenchmark.cpp
+++ b/examples/Raytracer_Benchmark/RaytraceBenchmark.cpp
@@ -16,6 +16,9 @@
 #include "ArgParser.h"
 #include "examples/Raytracer_Benchmark/Raytracer.h"
 #include <CopCore/Global.h>
+#include <AdePT/BlockData.h>
+#include "kernels.h"
+#include "examples/Raytracer_Benchmark/LoopNavigator.h"
 
 #ifdef VECGEOM_GDML
 #include <VecGeom/gdml/Frontend.h>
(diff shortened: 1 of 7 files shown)


hahnjo:
It's a bit abusing the `BlockData` (calling `NextElement` to exhaust the full capacity and then calling the array operator with unrelated `index`es), but I'll let @agheata review because I don't know what the plans going forward are. The `LoopNavigator` looks good to me, as far as I understand it's just moving the code for now, right?

examples/Raytracer_Benchmark/RaytraceBenchmark.cpp
```diff
@@ -126,6 +129,30 @@ int main(int argc, char *argv[])
 
 int RaytraceBenchmarkCPU(cxx::RaytracerData_t &rtdata)
```
agheata, line 130: This will need to become something like:
template <copcore::BackendType backend> int RaytraceBenchmarkCPU(RaytracerData_t<backend> &rtdata)
Needed to unify the launcher sequence, bit not for this PR

antoniopetre:
> ?

examples/Raytracer_Benchmark/RaytraceBenchmark.cpp
agheata: We will need to check if RaytracerData_t will have to be templated by the backend or not

antoniopetre:
> It's a bit abusing the `BlockData` (calling `NextElement` to exhaust the full capacity and then calling the array operator with unrelated `index`es), but I'll let @agheata review because I don't know what the plans going forward are. The `LoopNavigator` looks good to me, as far as I understand it's just moving the code for now, right?

Yes, I just moved the code from ```Raytracer.cpp```.

examples/Raytracer_Benchmark/RaytraceBenchmark.cu
```diff
@@ -59,7 +68,8 @@ __global__ void RenderLine(RaytracerData_t rtdata, int py, unsigned char *line)
   if (px >= rtdata.fSize_px) return;
 
   Ray_t ray;
-  adept::Color_t pixel_color = Raytracer::RaytraceOne(rtdata, ray, px, py);
+  adept::BlockData<Ray_t> *rays;
+  adept::Color_t pixel_color = Raytracer::RaytraceOne(rtdata, rays, px, py, 0);
```
hahnjo, line 72: Can we just drop the other render kernels that are not used by default? (`RenderLine`, `RenderInterlaced`, `RenderTile`)

agheata (approved):

PR #50, opened 2020-12-03
Cleanup navigation code

hahnjo:


agheata (approved):
Renaming some method names will mismatch with VNavigator API, but we're not currently using that anyway.

hahnjo:
> Renaming some method names will mismatch with VNavigator API, but we're not currently using that anyway.

I'd say that's actually an advantage because it's different than the existing API (that continues to have problems).

PR #51, opened 2020-12-03
Removal of extra dead code and using COPCORE_CUDA_CHECK for RT example.

agheata:
A bit more cleanup in the ray tracer example


hahnjo:
Can we also remove `RenderTiledImage` and `RenderTile`?

antoniopetre (approved):

agheata:
Cannot remove RenderTiledImage and RenderTile since this is used with the flag use_tiles=1

hahnjo:
> Cannot remove RenderTiledImage and RenderTile since this is used with the flag use_tiles=1

Yes I know, but does it make sense to keep this flag?

PR #52, opened 2020-12-03
WIP: Migrate CopCore/AdePT to Allen-style target device choice

drbenmorgan:
As we've been discussing, it's useful to try out the host/device portability system from LHCb's Allen project. There were some issues with the initial attempts, but this is now understood, and thus this PR has a go at a using "Allen style" in CopCore/AdePT.

It's quite cut down from @graeme-a-stewart's first import so that it only imports the basic preprocessor macros and CPU backend (though this is not used yet). It adds a new macro `DEVICE_COMPILATION_TRAJECTORY` that's equivalent to VecCore's `VECCORE_CUDA_DEVICE_COMPILATION`, and is defined if and only if the compiler is compiling for the device. 
It's never defined for CPU, and has only been set up for CUDA (there is a HIP equivalent, but later...). It also adds `COPCORE_FORCE_INLINE` to replace the VecCore equivalent.

The major changes in the CopCore/Adept code are relatively straightforward:

- Add a CMake option `TARGET_DEVICE` to CopCore's CMakeLists.txt to select the backend.
  - This is CUDA by default since Adept is (but see below).
- Include the new `CopCore/backend/BackendCommon.h` in `CopCore` headers
- Change `VECCORE_` macros to Allen's `__host__` etc and `COPCORE_` for the inlines

Have to admit I'm not *totally* convinced, but that's likely due to the remaining macros for inlining etc...

*Most* examples and tests then run, but there are a couple of exceptions:

- The RayTracing example. However, this didn't compile for me when starting this branch, so something else is wrong with that (to be checked).
- The `test/test_launcher.cpp` test.
  - The issue appears to be that the Allen style effectively hardcodes CPU/CUDA. In CUDA mode, you can't then use CopCore from a C++ translation unit (at least in the above naive translation). This is logical for Allen's use case as it's explicitly chosen at configure time, and uniform over compile time. We're not doing that in Adept though, at least as the tests/examples are implemented.

Admittedly I'm at the limit of my knowledge on the above, but I think comes down to how to separate C++/CUDA APIs and whether we are doing so in a *library* or an *application*. At least for the initial benchmark of a simple Geant4 app offloading a set of tracks in a calo to GPU, it _feels_ like we'll eventually need a C++ only API, 

```c++
// Adept.h
SomeReturnType offloadTracks(TrackCollection col);

// ---------
// Adept.cu
__global__ offloadTracks_CUDA(...) {
  ...
}

SomeReturnType offloadTracks(TrackCollection col) {
   ... prep data, etc
   offloadTracks_CUDA<<<...>>>(...);
   ... extract data
   return SomeReturnType(...);
}
```

... of course this brings us right back to the problem of compiling/deploying for heterogenous resources...

### DIFF
--- a/base/inc/AdePT/Atomic.h
+++ b/base/inc/AdePT/Atomic.h
@@ -10,10 +10,10 @@
 #ifndef ADEPT_ATOMIC_H_
 #define ADEPT_ATOMIC_H_
 
-#include <VecCore/VecCore>
+#include <CopCore/CopCore.h>
 #include <cassert>
 
-#ifndef VECCORE_CUDA_DEVICE_COMPILATION
+#ifndef DEVICE_COMPILATION_TRAJECTORY
 #include <atomic>
 #endif
 
@@ -24,7 +24,7 @@ namespace adept {
 template <typename Type>
 struct AtomicBase_t {
 
-#ifndef VECCORE_CUDA_DEVICE_COMPILATION
+#ifndef DEVICE_COMPILATION_TRAJECTORY
   /// Standard library atomic behaviour.
   using AtomicType_t = std::atomic<Type>;
 #else
@@ -35,19 +35,19 @@ struct AtomicBase_t {
   AtomicType_t fData{0}; ///< Atomic data
 
   /** @brief Constructor taking an address */
-  VECCORE_ATT_HOST_DEVICE
+  __host__ __device__
   AtomicBase_t() : fData{0} {}
 
   /** @brief Copy constructor */
-  VECCORE_ATT_HOST_DEVICE
+  __host__ __device__
   AtomicBase_t(AtomicBase_t const &other) { store(other.load()); }
 
   /** @brief Assignment */
-  VECCORE_ATT_HOST_DEVICE
+  __host__ __device__
   AtomicBase_t &operator=(AtomicBase_t const &other) { store(other.load()); }
 
   /** @brief Emplace the data at a given address */
-  VECCORE_ATT_HOST_DEVICE
+  __host__ __device__
   static AtomicBase_t *MakeInstanceAt(void *addr)
   {
     assert(addr != nullptr && "cannot allocate at nullptr address");
@@ -57,7 +57,7 @@ struct AtomicBase_t {
   }
 
   /// Implementation-dependent
-#ifndef VECCORE_CUDA_DEVICE_COMPILATION
+#ifndef DEVICE_COMPILATION_TRAJECTORY
 
   /** @brief Atomically replaces the current value with desired. */
   void store(Type desired) { fData.store(desired); }
@@ -78,11 +78,11 @@ struct AtomicBase_t {
 #else
 
   /** @brief Atomically replaces the current value with desired. */
-  VECCORE_ATT_DEVICE
+  __device__
   void store(Type desired) { atomicExch(&fData, desired); }
 
   /** @brief Atomically loads and returns the current value of the atomic variable. */
-  VECCORE_ATT_DEVICE
+  __device__
   Type load() const
   {
     // There is no atomic load on CUDA. Issue a memory fence to get the required
@@ -92,12 +92,12 @@ struct AtomicBase_t {
   }
 
   /** @brief Atomically replaces the underlying value with desired. */
-  VECCORE_ATT_DEVICE
+  __device__
   Type exchange(Type desired) { return atomicExch(&fData, desired); }
 
   /** @brief Atomically compares the stored value with the expected one, and if equal stores desired.
    * If not loads the old value into expected. Returns true if swap was successful. */
-  VECCORE_ATT_DEVICE
+  __device__
   bool compare_exchange_strong(Type &expected, Type desired)
   {
     Type old    = atomicCAS(&fData, expected, desired);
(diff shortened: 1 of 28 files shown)


graeme-a-stewart:
Thanks a lot @drbenmorgan!

Out of the box this isn't working for me, I immediately get a compilation error:

```
[4/29] Building CXX object test/CMakeFiles/test_launcher.dir/test_launcher.cpp.o
FAILED: test/CMakeFiles/test_launcher.dir/test_launcher.cpp.o
/usr/bin/c++ -DTARGET_DEVICE_CUDA -DVECCORE_ENABLE_CUDA -I/home/graemes/code/AdePT/base/inc -I/home/graemes/code/AdePT/base/inc/CopCore/include -isystem /opt/nvidia/cuda-11.0/include -fabi-version=0 -faligned-new -msse2 -O2 -g -DNDEBUG -std=c++14 -MD -MT test/CMakeFiles/test_launcher.dir/test_launcher.cpp.o -MF test/CMakeFiles/test_launcher.dir/test_launcher.cpp.o.d -o test/CMakeFiles/test_launcher.dir/test_launcher.cpp.o -c /home/graemes/code/AdePT/test/test_launcher.cpp
In file included from /home/graemes/code/AdePT/base/inc/CopCore/include/CopCore/CopCore.h:18,
                 from /home/graemes/code/AdePT/base/inc/AdePT/BlockData.h:13,
                 from /home/graemes/code/AdePT/test/test_launcher.cpp:11:
/home/graemes/code/AdePT/base/inc/CopCore/include/CopCore/Launcher.h: In function 'void copcore::kernel_launcher_impl::kernel_dispatch(int, Function, const Args ...)':
/home/graemes/code/AdePT/base/inc/CopCore/include/CopCore/Launcher.h:25:18: error: 'blockIdx' was not declared in this scope
   for (auto id = blockIdx.x * blockDim.x + threadIdx.x; id < data_size; id += blockDim.x * gridDim.x) {
                  ^~~~~~~~
...
```

I will dig into it.

drbenmorgan:
Ah, that's one of the known errors! The issue is that `test_launcher.cpp` is trying to use a `CopCore` that's targeting CUDA (`-DTARGET_DEVICE_CUDA`). At present, only one device can be targeted in any one build. I'm not sure how we want to handle this:

- Disable tests for devices other than being targeted.
- Build CopCore once for each device and link explicitly to the required one

graeme-a-stewart:
OK, now I read your initial notes more carefully - so do you also see a compilation failure on `test/test_launcher.cpp`? Initially I thought it was just a runtime error...

drbenmorgan:
Yes, I also get that error at compile time.

graeme-a-stewart:
Thanks @drbenmorgan - ok, I can confirm that skipping over that then the basic fisher_price examples run. Of course, these are dyed-in-the-wool CUDA examples (explicit use of curand) so not helpful for attempting prove CPU/CUDA device switching.

agheata:
Well, I'll try on my side to understand how to use a launcher in this approach (or if it still makes sense) and what are the constraints on how the code must be instrumented. I'd really like to see an implementation of the ray-tracer example using both styles. Meanwhile, it would be great if we manage to push this in keeping still the possibility to use the previous examples - even if it's ugly for a while. Not sure this is possible due to the differences in using macro annotations. I'll start looking into this.

drbenmorgan:
O.k., I've tried the approach we discussed on Friday of having two versions of CopCore compiled concurrently, one for CPU, `libCopCore_CPU`, and one for GPU, `libCopCore_CUDA`. That is easy enough in CMake, it's just two different `add_library` calls. What fails, and I think near terminally, for this approach is that examples like `test_launcher` would still require linking to _both_ libraries since it's a mix of C++ and CUDA. At least it is if you care about dispatching to one or the other backend at runtime. 

That's terminal for the Allen approach as it stands (and as AdePT/CopCore are currently implemented) because you'll always end up with symbol clashes from the redefinition of things like the cuda API calls/dim3 variables.  Only way around that is inline namespaces as acknowledged here:

https://github.com/apt-sim/AdePT/blob/6e5dd37b3584e26ef40de97cdfff211b34d3aa19/test/sim_kernels.h#L23-L24

and that means Allen's overrides cannot be used directly in the sense of "write everything CUDA style and the macros handle everything on CPU/HIP". They *could* be hidden in inline namespace(s) but not sure we gain much by that, or what traps/pitfalls there might be (e.g. host vs device memory management).

Personally at this point I think we need simplicity and clarity over portability. If that means _some_ double coding to compare CPU/GPU then I think within the scope of AdePT that would be more efficient, but further discussion and ideas are very welcome!

agheata:
I'm not very fan of the mixed CPU/GPU approach either, because the support provided by CUDA for doing this is very limited - this is what gave bread on the table to portability frameworks. I tried to combine the two worlds with the Launcher approach, but eventually the code in more complex cases becomes a mess: what runs on host/device, which inline namespace is applied where, or what has to be templated by the backend. I think what we should stick to is to have a unique definition of the algorithmic library. The fact that we need to duplicate some code for backend-specific launching or memory management is OK. So we can have backend-specific compilation targets and lower our expectations that examples such as test_launcher would work. On the other hand, this should not prevent us in using utilities that are templated by the backend, when the provided functionality can be mapped (e.g. memory allocators).

On the practical side, I'd still try to modify the test_launcher to work separately for the 2 backends (not in the same example) - to get an idea of how much of the code can be put in common. Would you push your mods for creating 2 separate libraries/executables for the examples that can be migrated?

drbenmorgan:
Thanks @agheata, the _minimal_ dual build is in the latest commit. It's not sophisticated, but should give the general picture.

WitekPokorski:
I must agree here with the statement 'we need simplicity and clarity over portability'. Our goal is to prove that we can generate and EM shower on GPU and that this gives speedup with respect to a similar shower simulated on CPU (with Geant4 being the baseline). We can't forget about portability and we should certainly avoid doing things that would never work outside CUDA, but shouldn't also spend too much energy of providing portability of something we don't have yet. I would say that our priority should be now to have a proper stepping-like pipeline of kernels with transportation, real physics and geometry (and magnetic field).

agheata:
I checked out this branch and rebased on master, fixing locally some minor conflicts. Indeed, if I disable the test_launcher the raytracing and example2 it compiles and runs for COPCORE_ENABLE_CUDA=ON. Switching this flag to OFF produces a missing <CopCore/CopCore.h>because the tests are linking against CopCore::CopCore_CUDA but this target is not created. I guess we want to support 2 versions for some (most) of the tests and examples, for CPU and CUDA. What do you think about refactoring the tests produce separate targets for the 2 cases, like test_track_block_cpp and test_track_block_cuda? I would be interested to see a clean separation between cpu/cuda/common files/methods.

PR #55, opened 2020-12-09
Add example2 with basic geometry navigation

hahnjo:
Move and reuse `ArgParser.h` and `LoopNavigator.h` from `examples/Raytracer_Benchmark/`. Also keep one `curand_state` per `track` which allows for identical results between multiple runs (apart from small variations due to floating point arithmetic in the energy deposition).


examples/Example2/example2.cu
```diff
@@ -0,0 +1,224 @@
...
+
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < n; i += blockDim.x * gridDim.x) {
+    // skip particles that are already dead
+    track &mytrack = (*block)[i];
+    if (mytrack.status == dead) continue;
+
+    // return value (if step limited by physics or geometry) not used for the moment
+    // now, I know which process wins, so I add the particle to the appropriate queue
+    float physics_step = proclist->GetPhysicsInteractionLength(i, block);
+    float step         = LoopNavigator::ComputeStepAndPropagatedState(mytrack.pos, mytrack.dir, physics_step,
```
agheata, line 62: Indeed, the geometry step has to be computed after having the physics step limitation. In a future version we have to try to split it out as separate kernel, i.e. cache the physics proposed step in the track and exit this, then pass the relay to the geometry kernel. We can do it later though.

examples/Example2/example2.cu
```diff
@@ -0,0 +1,224 @@
...
+}
+
+// kernel to call Along Step function for particles in the queues
+__global__ void CallAlongStepProcesses(adept::BlockData<track> *block, process_list *proclist, adept::MParray **queues,
+                                       Scoring *scor)
+{
+  int particle_index;
+
+  // loop over all processes
+  for (int process_id = 0; process_id < proclist->list_size; process_id++) {
```
agheata, line 84: Separating the kernels per process would have to be tried, but again, in a future version

agheata (approved):
Looks good to go if all examples are running after this. I added some notes suggesting splitting out the process and geometry kernels in a next version.

examples/Example1/CMakeLists.txt
```diff
@@ -9,4 +9,4 @@ target_include_directories(example1 PUBLIC
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/tracking/inc>
   $<INSTALL_INTERFACE:base>
 )
-target_link_libraries(example1 PRIVATE CUDA::curand VecCore::VecCore CopCore::CopCore)
+target_link_libraries(example1 PRIVATE CUDA::curand VecCore::VecCore VecGeom::vecgeom CopCore::CopCore)
```
WitekPokorski, line 12: Is it intentional, or just by mistake?

agheata: You refer to linking with vecgeom? The geometry is defined there...

WitekPokorski: Yes, but this is Example1 cmake file.

hahnjo: This adds the VecGeom include path which is now needed by the (shared) `track.h`. The linking would not be needed, but I didn't try to find a more correct solution.

PR #59, opened 2020-12-11
Use Allen/prototype macros for host/device markup

drbenmorgan:
Testing and discussion showed that using the full Allen approach would not work immediately for AdePT due to symbol clashes from the requirement for a mix of CPU/GPU code. However, its macro system is still useful to simplify markup of functions and indicate CUDA use and the device-compile pass.

Implement a simple header in CopCore based on Allen and Guilherme Amadio's prototype system. Define macros to indicate:

- Use of a CUDA compiler (i.e. compiling code as CUDA)
- Compilation is on the device pass

Use Allen's style of defining CUDA's `__host__` etc keywords as
empty defines when compiling for a host/non-cuda compiler.

Migrate CopCore, AdePT, example, and tests code to use the new macros.

There are some warnings in the `examples`, but I need to update my VecGeom install with the NavStateIndex to get this fully working. The core behaviour is in the new `Macros.h` header, the rest is simple replacement of the VecCore macros with the new ones.

This replaces #52, which I'll close.

### DIFF
--- a/base/inc/AdePT/Atomic.h
+++ b/base/inc/AdePT/Atomic.h
@@ -10,10 +10,10 @@
 #ifndef ADEPT_ATOMIC_H_
 #define ADEPT_ATOMIC_H_
 
-#include <VecCore/VecCore>
+#include <CopCore/Global.h>
 #include <cassert>
 
-#ifndef VECCORE_CUDA_DEVICE_COMPILATION
+#ifndef COPCORE_DEVICE_COMPILATION
 #include <atomic>
 #endif
 
@@ -24,7 +24,7 @@ namespace adept {
 template <typename Type>
 struct AtomicBase_t {
 
-#ifndef VECCORE_CUDA_DEVICE_COMPILATION
+#ifndef COPCORE_DEVICE_COMPILATION
   /// Standard library atomic behaviour.
   using AtomicType_t = std::atomic<Type>;
 #else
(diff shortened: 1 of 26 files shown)


agheata (changes requested):
All examples are running with this, including test_launcher. This is because the PR does not create separate CPU/GPU targets for the headers, but mainly changes the definition of attribute macros. The PR can be merged as soon as we make the warnings go away.

agheata:
Fixing the redefinition warning is important, because the header inclusion order can trigger using the wrong macros. The warning fix is non-trivial, because VecGeom includes cuda/Interface.h in all compilation trajectories when VECGEOM_ENABLE_CUDA is ON. This includes driver_types.h from CUDA, which has the definition of the overridden macros. This hits whenever we need to include VecGeom entities (even indirectly) having e.g. CopyToGPU functionality (such as Vector.h)

agheata:
If we include CopCore/Global.h before the VecGeom headers, the warning goes away, while the example compiles and runs correctly, so maybe this is the fix...:
```
diff --git a/examples/Example2/example2.h b/examples/Example2/example2.h
index a1cd3a7..6ef0dbc 100644
--- a/examples/Example2/example2.h
+++ b/examples/Example2/example2.h
@@ -4,6 +4,7 @@
 #ifndef EXAMPLE2_H
 #define EXAMPLE2_H
 
+#include <CopCore/Global.h>
 #include <VecGeom/base/Config.h>
 #include <VecGeom/volumes/PlacedVolume.h>
```

agheata:
We'll merge like this and apply the warning fix later

agheata (approved):

PR #60, opened 2020-12-17
Fix for symbol redefinition warning.

agheata:
This warning seems to have an easy fix w/o impacting the execution of tests and examples

### DIFF
--- a/examples/Example2/example2.h
+++ b/examples/Example2/example2.h
@@ -4,6 +4,7 @@
 #ifndef EXAMPLE2_H
 #define EXAMPLE2_H
 
+#include <CopCore/Global.h>
 #include <VecGeom/base/Config.h>
 #include <VecGeom/volumes/PlacedVolume.h>
 


hahnjo:
As far as I understand, the problem is that `crt/host_defines.h` correctly guards the macro definition by `!defined(__host__)` whereas `CopCore/Macros.h` does not. What I don't understand yet is why we don't continue to have problems with `__forceinline__`, the following minimal code throws a warning for me:
```c++
#define __forceinline__ inline __attribute__((always_inline))
#include <cuda_runtime.h>
```

(note that `__inline__` instead of `inline` is clean because this is the exact same value that `crt/host_defines.h` and at least GCC doesn't warn about the redefinition)

hahnjo:
> What I don't understand yet is why we don't continue to have problems with `__forceinline__`

The reason is that CMake passes the include path for CUDA via `-isystem` which suppresses all warnings. In conclusion, I think we should make `CopCore/Macros.h` more robust in conjunction with the `define`s coming from CUDA headers. @drbenmorgan do you know if the original Allen headers have some more magic for this?

hahnjo:
https://gitlab.cern.ch/lhcb/Allen/-/blob/master/backend/include/CPUBackend.h doesn't have any magic, but then again they probably don't include `driver_types.h` -> `crt/host_defines.h`.

I think we have three options (assuming `VecGeom/backend/cuda/Interface.h` needs access to CUDA types even in C++ mode, I didn't check):
1. Always include "our" definitions from `CopCore/Macros.h` first, possibly via `CopCore/Global.h` (this PR). This might need changes to many files over time, if for example a header file includes another CUDA header at some point.
2. Make `CopCore/Macros.h` more robust (https://github.com/apt-sim/AdePT/pull/61) and avoid redefinitions.
3. First `undef` keywords in `CopCore/Macros.h` and always ensure our header provides the true definition. This might lead to problems when we're indeed compiling CUDA code and the real definitions are important.

agheata:
> 2. Make `CopCore/Macros.h` more robust (#61) and avoid redefinitions.

I think this approach is quite safe, we only need to make sure that the cuda header defining these symbols is always included before our Macros.h in nvcc compilation units, but this does not need explicit inclusion of headers, right? If this is the case I would close this and approve #61

hahnjo:
> > 1. Make `CopCore/Macros.h` more robust (#61) and avoid redefinitions.
> 
> I think this approach is quite safe, we only need to make sure that the cuda header defining these symbols is always included before our Macros.h in nvcc compilation units, but this does not need explicit inclusion of headers, right? If this is the case I would close this and approve #61

I *think* that when compiling with `nvcc`, the compiler injects the keywords very early in the process and that's also why the definitions in the CUDA headers are guarded with `#ifndef`s, to not redefine them.

agheata:
Superseeded by #61

PR #61, opened 2021-01-06
Avoid redefinition of CUDA keywords

hahnjo:
The header `VecGeom/backend/cuda/Interface.h` includes some files from
the CUDA toolkit that also define these keywords. Avoid warnings by
first checking if the keywords are already defined.


agheata (approved):
This looks like the safe way of doing it, see discussions on #60

PR #62, opened 2021-01-08
Implementation of a sparse vector.

agheata:
SparseVector is a variable-size vector, adopting pre-allocated space and targeting both host/device use. It supports in-place construction while distributing concurrently elements. The vector allows releasing elements and checking if a given index is in use. It provides getters for the number of shared, used, and total elements, as well as ratios representing sparsity, shared, and selected fractions. It supports selection operations using arbitrary predicate functions and executes concurrently as a kernel. SparseVector can be compacted concurrently into another SparseVector.

Commented tests demonstrating main functionality: test_sparsevector (GPU) and test_sparsevector_cpu (CPU).

### DIFF
--- a/base/inc/AdePT/SparseVector.h
+++ b/base/inc/AdePT/SparseVector.h
@@ -1,0 +1,731 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @file SparseVector.h
+ * @brief Multi-producer variable-size vector adopting pre-allocatrd memory
+ * @details The vector is templated on the stored type and size.
+ * - Allocation: pre-allocate sizeof(SparseVector) and call the static allocator:
+ *   SparseVector<Type,N>::MakeInstance(void *addr)
+ * - Use the base class SparseVectorInterface<Type> as main interface. It provides
+ *   iterators and random access,
+ * - Distributing concurrently vector elements, withing the available size:
+ *     Type *elem = vect->next_free(parameters);
+ *   The compiler will match the parameters with those of a constructor of Type
+ * - Releasing elements at a given index:
+ *     vect->release(index);
+ * - Getting number of shared and used elements:
+ *     SparseVectorInterface::size() / size_used();
+ * - Sparsity is defined as used/shared, shared fraction is nshared/capacity
+ * - The SparseVectorInterface provides static methods for predicate-based selection of elements:
+ *     select(SparseVectorInterface<Type> *vect, Predicate select_func, Container *ind, unsigned *nsel)
+ *       vect - Vector to select from
+ *       select_func - predicate function having the format:
+ *         __device__ bool func(int, const Vector_t*) // index in the vector and const vector pointer
+ *       ind - containter holding the indices of the selected elements
+ *       nsel - number of selected elements
+ *     select_and_move(SparseVectorInterface<Type> *source, Predicate select_func,
+ *                     SparseVectorInterface<Type> *dest, unsigned nsel)
+ *       source - source vector
+ *       select_func - selection predicate function
+ *       dest - destination vector
+ *       nsel - number of selected elements
+ *       The selected elements are removed from source and added at the end of dest.
+ *     relese_selected(SparseVectorInterface<Type> *source, Container *ind, unsigned *nsel)
+ *       Releases nsel elements stored in the ind vector
+ *     select_used(SparseVectorInterface<Type> *source, Container *ind, unsigned *nsel)
+ *       Copies the indices of all used elements in ind.
+ *     compact(SparseVectorInterface<Type> *source, SparseVectorInterface<Type> *source, unsigned *nsel)
+ *       Moves all used elements from source to the end of dest, returning the number of copied elements nsel
+ *
+ * @author Andrei Gheata (andrei.gheata@cern.ch)
+ */
+
+#ifndef ADEPT_SPARSEVECTOR_H_
+#define ADEPT_SPARSEVECTOR_H_
+
+#include <CopCore/CopCore.h>
+#include <AdePT/Atomic.h>
+
+namespace adept {
+
+template <typename Type>
+class SparseVectorInterface;
+
+template <typename Type, unsigned N>
+class SparseVector;
+
+namespace sa_detail {
+
+#ifdef COPCORE_CUDA_COMPILER
+
+__device__ inline int lane_id(void)
+{
+  return threadIdx.x & 31;
+}
+
+template <typename Type>
+__device__ inline bool is_used(int index, const SparseVectorInterface<Type> *svector)
+{
+  return svector->is_used(index);
+}
+
+__device__ void print_mask(unsigned int mask)
+{
+  for (int lane = 0; lane < 32; ++lane) {
+    if ((mask & (1 << lane)))
+      printf("1");
+    else
+      printf("0");
+  }
+  printf("\n");
+}
+
+template <typename Type, unsigned N>
+__global__ void construct_vector(void *addr, int numSMs)
+{
+  auto vect = new (addr) SparseVector<Type, N>(N);
+  vect->setNumSMs(numSMs);
+}
+
+template <typename Type>
+__global__ void release_selected_kernel(SparseVectorInterface<Type> *svect, const unsigned int *selection,
+                                        unsigned *nselected)
+{
+  // Release the selected entries
+  for (auto tid = blockIdx.x * blockDim.x + threadIdx.x; tid < *nselected; tid += blockDim.x * gridDim.x)
+    svect->release(selection[tid]);
+}
+
+template <typename Type>
+__global__ void release_selected_kernel_launcher(SparseVectorInterface<Type> *svect, const unsigned int *selection,
+                                                 unsigned *nselected)
+{
+  // Launcher for releasing the selected entries
+  constexpr unsigned int warpsPerSM = 32; // target occupancy
+  constexpr unsigned int block_size = 256;
+  unsigned int grid_size =
+      min(warpsPerSM * svect->getNumSMs() * 32 / block_size, (*nselected + block_size - 1) / block_size);
+  // printf("running release_selected_kernel<<<%d, %d>>>\n", grid_size, block_size);
+  sa_detail::release_selected_kernel<Type><<<grid_size, block_size>>>(svect, selection, nselected);
+}
+
+template <typename Type>
+__global__ void clear_kernel(SparseVectorInterface<Type> *svect)
+{
+  // Clear the vector
+  svect->clear();
+}
+
+/// Make a selection of elements according to the user predicate and write it to the output.
+/// Assume that IndexContainer::operator[] is implemented and use it to copy selected element indices
+template <typename Type, typename Predicate, typename IndexContainer>
+__global__ void select_kernel(const SparseVectorInterface<Type> *svect, Predicate pred_func, IndexContainer *output,
+                              unsigned *nselected)
+
+{
+  // Non-order-preserving stream compacting algorithm
+  // Based on algorithm described in https://www.jstage.jst.go.jp/article/ijnc/7/2/7_208/_pdf/-char/ja
+  // nselected content must be zeroed before calling the kernel
+  constexpr unsigned int warp_mask = 0xFFFFFFFF;
+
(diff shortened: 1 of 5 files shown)


antoniopetre (approved):

base/inc/AdePT/SparseArray.h
```diff
@@ -0,0 +1,417 @@
...
+    static_assert(std::is_copy_constructible<Type>::value,
+                  "SparseArray::compact: Error: stored type is not copy constructible");
+    // select elements still in use
+    select_used();
+    assert(fNselected < (dest->size_max() - dest->size_shared()) &&
+           "SparseArray::compact: Error: not enough free space in destination array");
+    // copy to destination array
+    const int num_items = size_shared();
+    int num_blocks      = (num_items + 127) >> 7;
+    sa_detail::copy_kernel<Array_t, Array2_t><<<num_blocks, 127>>>(this, dest);
```
hahnjo, line 375: I think you want to launch with `num_threads = 128`, with `127 = 128 - 1` to round up. Can you also change the computation of `num_blocks` to the more idiomatic `(num_items + num_threads - 1) / num_threads` (here and in other places)? The compiler will do the optimization to `>> 7` for you, but that way it's easier to change `num_threads`.

base/inc/AdePT/SparseArray.h
```diff
@@ -0,0 +1,417 @@
...
+  }
+#endif
+
+  /** @brief Fills selection array fSelected with indices of remaining used elements. */
+#ifdef COPCORE_CUDA_COMPILER
+  void release_selected()
+  {
+    const int num_items = size_shared();
+    int num_blocks      = (num_items + 127) >> 7;
+    sa_detail::release_selected_kernel<Array_t><<<num_blocks, 127>>>(this);
```
hahnjo, line 402: Same here, should also be 128 threads

hahnjo (commented):
Some comments on the code inline, but I've also got a few design questions:

1. `SparseArray` is templated on the number of elements `N` which requires every called function to be templated. I think LLVM's `SmallVector` had a clever solution for this using (plain) inheritance, but I'll have to look this up in the code.
2. Do we need `fSelected` in every array? From my understanding, its nature is rather temporary to communicate selection from one kernel to another, but it shouldn't be needed for the destination of `compact`.
3. The code currently passes `this` to the device kernel which requires the object to live in managed memory. This is easier to program and work with, but might have performance implications. So again the question is whether this restriction is built into the design or if it could be lifted later on?

agheata:
Thanks Jonas, all good comments, working on them.

agheata:
> Some comments on the code inline, but I've also got a few design questions:
> 
> 1. `SparseArray` is templated on the number of elements `N` which requires every called function to be templated. I think LLVM's `SmallVector` had a clever solution for this using (plain) inheritance, but I'll have to look this up in the code.
I agree that this is a big limitation for the interfaces. We can use the existing VariableSizeObjectInterface to get around that - I don't know about the SmallVector.

> 2. Do we need `fSelected` in every array? From my understanding, its nature is rather temporary to communicate selection from one kernel to another, but it shouldn't be needed for the destination of `compact`.
It was convenient for testing the implementation, if it's not there it forces the user to provide pre-allocated arrays. But that's fine if we can have a memory manager providing those. For compacting, the copy to the destination array can be direct indeed. I'll remove it and adapt the test.

> 3. The code currently passes `this` to the device kernel which requires the object to live in managed memory. This is easier to program and work with, but might have performance implications. So again the question is whether this restriction is built into the design or if it could be lifted later on?
Indeed, there are performance implications due to having to synchronize in order to update the host-resident class data. No, this is not by design, I mainly focused on having an optimized selection mechanism, that is all. The design is intended to have the array data on device, but be able to launch from the host kernels doing array operations that do not require array state to be available on the host. This is not achieved in the current implementation. We should probably use static methods to initiate the launches. The problem is that the launch configuration requires at minimum the number of elements to iterate on, which is state. My idea was to initiate a <<<1,1>>> kernel that would see the device state of the array and do the actual launch (dynamic parallelism), but not yet there. Another way is to initiate a grid-size loop in the kernels, but this may imply large launches that will penalize a multi-stream approach.

We need to iterate on this, I think the functionality behind this is essential for the prototype, so we need to do it right.

base/inc/AdePT/SparseArray.h
```diff
@@ -0,0 +1,417 @@
...
+   *  @returns Number of selected elements
+   */
+#ifdef COPCORE_CUDA_COMPILER
+  template <typename PREDICATE>
+  void select(PREDICATE pred_func)
+  {
+    int num_items = size_shared();
+    int num_warps = (num_items + 1023) >> 10; // a warp processes groups of 1024 elements
+    fNselected    = 0;
+    sa_detail::select_kernel<Array_t, PREDICATE><<<num_warps, 32>>>(this, pred_func, fSelected, &fNselected);
```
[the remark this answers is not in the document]
agheata: I'll have to check if the kernel still works and if an extra early return is needed, but you are right, we should target at least 128

base/inc/AdePT/SparseArray.h
```diff
@@ -0,0 +1,417 @@
...
+template <typename Array_t>
+__global__ void release_selected_kernel(Array_t *sarray)
+{
+  // Release the selected entries
+  const int nselected = sarray->size_selected();
+
+  int tid = blockIdx.x * blockDim.x + threadIdx.x;
+  if (tid >= nselected) return;
+  const unsigned int *selection = sarray->get_selection_array();
+  sarray->release(selection[tid]);
```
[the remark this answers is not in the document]
agheata: Converted to a grid-size loop, allowing to pass an optional launch configuration as argument.

agheata:
> Some comments on the code inline, but I've also got a few design questions:
> 
> 1. `SparseArray` is templated on the number of elements `N` which requires every called function to be templated. I think LLVM's `SmallVector` had a clever solution for this using (plain) inheritance, but I'll have to look this up in the code.
> 2. Do we need `fSelected` in every array? From my understanding, its nature is rather temporary to communicate selection from one kernel to another, but it shouldn't be needed for the destination of `compact`.
> 3. The code currently passes `this` to the device kernel which requires the object to live in managed memory. This is easier to program and work with, but might have performance implications. So again the question is whether this restriction is built into the design or if it could be lifted later on?

All these issues are fixed in the new class SparseVector

base/inc/AdePT/SparseVector.h
```diff
@@ -0,0 +1,746 @@
...
+  sa_detail::select_kernel<Type, Predicate, IndexContainer>
+      <<<num_blocks, num_threads>>>(svect, pred_func, output, nselected);
+}
+
+/// Make a selection of elements according to the user predicate and write it to the output.
+/// Copy directly elements at the end of the output vector.
+/// Similar to select_kernel, but having extra actions applied to the vectors
+template <typename Type, typename Predicate>
+__global__ void select_and_move_kernel(SparseVectorInterface<Type> *svect, Predicate pred_func,
+                                       SparseVectorInterface<Type> *output, unsigned *nselected, int dest_offset)
```
hahnjo, line 214: This is very similar to `select_kernel`, any chance we can deduplicate the code?

base/inc/AdePT/SparseVector.h
```diff
@@ -0,0 +1,746 @@
...
+
+  /** @brief Fills selection vector fSelected with element indices passing the user predicate
+   *  @param pred Predicate function taking as arguments the SparseVector pointer and the element index
+   *  @returns Number of selected elements. Must be reset to 0 when calling the function.
+   */
+#ifdef COPCORE_CUDA_COMPILER
+  template <typename Predicate, typename Container>
+  static void select_and_move(Vector_t *svect, Predicate pred_func, Container *output, unsigned *num_selected)
+  {
+    sa_detail::select_and_move_kernel_launcher<Type><<<1, 1>>>((Vector_t *)svect, pred_func, output, num_selected);
```
hahnjo, line 605: The type of `output` should be `Vector_t` as well, at least that's what `select_and_move_kernel_launcher` expects. Also I don't think you need to cast `svect`

base/inc/AdePT/SparseVector.h
```diff
@@ -0,0 +1,746 @@
...
+  AtomicInt_t fNused;         ///< Number of elements currently in use
+  unsigned fCapacity{0};      ///< Capacity of the vector
+  unsigned fMaskCapacity{0};  ///< Capacity of the mask vector
+  int fNumSMs{0};             ///< number of streaming multi-processors
+
+  /// Set first element address
+  __host__ __device__ __forceinline__ void setFirstEl(void *addr) { fBegin = addr; }
+
+  /// Set first mask address
+  __host__ __device__ __forceinline__ void setFirstElMask(void *addr) { fMaskBegin = addr; }
```
hahnjo, line 345: Not needed, the derived class has direct access to the pointers

base/inc/AdePT/SparseVector.h
```diff
@@ -0,0 +1,746 @@
...
+template <typename Type, unsigned N>
+struct SparseVectorStorage {
+  alignas(Type) char fData[N * sizeof(Type)];
+  unsigned int fMasks[N / 32];
+};
+
+/// Aligned storage may be needed also in case of SparseVector of size 0, if allocating in a buffer
+template <typename Type>
+struct alignas(Type) SparseVectorStorage<Type, 0> {
+};
```
hahnjo, line 697: Can we fold this into `SparseVector` directly? LLVM only does this to save a pointer while making sure that multiple inheritance puts the data array first in the layout of the derived class. I don't think the specialization `SparseVectorStorage<Type, 0>` works for us, `SparseVector` attempts to access `fData` and `fMasks` regardless of the size.

hahnjo (commented):
Looks very good, some minor comments inline. The question in `select_and_move_kernel` is not strongly needed for the initial version, although I'd prefer not to duplicate the code...

base/inc/AdePT/SparseVector.h
agheata: I tried to split the code, unfortunately, one cannot rely on return values since decisions are taken at the thread level, to know what to copy and when. If no immediate solution, we can postpone the refactoring for later.

agheata: One possibility is to use an extra template parameter, tried that and got some compilation error so decided not to spend extra time on it.

base/inc/AdePT/SparseVector.h
agheata: Left-overs from the previous implementation...

base/inc/AdePT/SparseVector.h
agheata: leftovers - removed

base/inc/AdePT/SparseVector.h
hahnjo: Yes, not required in this PR. I might have a look when time allows, but that can happen as a follow-up.

base/inc/AdePT/SparseVector.h
agheata: Sure, I'll only add an extra static_assert against 0 capacity (that we won't have to support).

agheata:
I squashed the commit resolving the remaining conversations. I think this is good to merge once it passes all tests.

base/inc/AdePT/SparseVector.h
hahnjo: No, I meant to completely remove `SparseVectorStorage` and put the members into `SparseVector`, or is there a reason this needs to be a separate class? Also `Type fData[N]` should be easier than `char` + `alignas`.

agheata: OK, I see. I'm getting some errors doing that, but I'll try to fix them.

agheata: Got rid of SparseVectorStorage but I want to avoid Type fData[N] which calls the default ctor.

hahnjo (approved):
LGTM :+1: 

> I squashed the commit resolving the remaining conversations. I think this is good to merge once it passes all tests.

Yes, please squash the two commits so that we have the implementation and the test in the same commit

agheata:
> LGTM 👍
> 
> > I squashed the commit resolving the remaining conversations. I think this is good to merge once it passes all tests.
> 
> Yes, please squash the two commits so that we have the implementation and the test in the same commit

OK, done

PR #63, opened 2021-01-14
Initial developments for propagation in (constant) magnetic field

jonapost:
Propagate tracks in uniform magnetic field using helix.

Changed Example2 to propagate tracks in field along Bz axis.  Method to propagate tracks with arbitrary B field also prepared (will integrate into example2 in next step.)

Multiple changes in track: added number of steps, mother index, methods for charge & mass.  Changed pdg & current process to type 'char'.


hahnjo:
I'm confused, should I look at `field_cu` or the changes to `Example2`? Maybe make this an `Example3`?

jonapost:
@hahnjo: the changes to Example2 are the main development.  ( I will update them slightly to use a general magnetic field - but it's along exactly the same lines as the current Bz-only field.)  

field_step.cu is basically a simple test.  I will need to find a better place for it.

I can copy these into Example3 for extending to non-Bz field -- but if you can comment on the current developments it will be helpful.

physics/processes/inc/pair_production.h
```diff
@@ -25,26 +25,37 @@ __device__ void pair_production::GenerateInteraction(int particle_index, adept::
...
 
-  float eloss = 0.5f * mytrack->energy;
+  float esecond = 0.5f * mytrack->energy;
 
   // pair production
-  mytrack->energy -= eloss;
+  mytrack->energy -= esecond;
   mytrack->energy_loss           = 0;
-  mytrack->number_of_secondaries = 1;
+  mytrack->number_of_secondaries++;
```
hahnjo, line 33: This might be badly named, but `number_of_secondaries = 1` is actually correct I think: this sets the number of secondaries produced in this process. This is accumulated in the global scoring which is now overcounting (ie a second secondary produced by the same primary track now counts as two additional particles, then three and so on)

physics/processes/inc/pair_production.h
```diff
@@ -25,26 +25,37 @@ __device__ void pair_production::GenerateInteraction(int particle_index, adept::
...
-  secondary_track->energy_loss           = 0;
-  secondary_track->number_of_secondaries = 0;
-  // Inherit current position and direction.
-  secondary_track->pos           = mytrack->pos;
-  secondary_track->dir           = mytrack->dir;
-  secondary_track->current_state = mytrack->current_state;
-  // Initialize a new PRNG state.
-  secondary_track->rng_state = mytrack->rng_state;
-  secondary_track->rng_state.Skip(1 << 15);
+  if( secondary_track != nullptr){
```
hahnjo, line 37: This may need a better solution, but for now I think this should really crash (hence the `assert`) because if there's no slot available, we should not silently skip the secondary

tracking/inc/track.h
```diff
@@ -15,28 +17,153 @@ enum TrackStatus { alive, dead };
...
+
+  uint16_t    num_step{0}; // number of Steps -- for killing after MaX
+   
+  char        current_process{0}; // -1 for Navigation / -2 for Field Propagation
+  char        pdg{0};             // Large enough for e/g. Then uint16_t for p/n
+
+  __device__ __host__ double uniform() { return rng_state.Rndm(); }
+
+  __device__ __host__ int    charge() const
+  { int chrg= (pdg== -11)-(pdg==11); return chrg; }
```
hahnjo, line 46: 😄

I think this deserves a comment. Or you can write it with `if`s / ternaries for now and let the compiler take care of optimizing it.

hahnjo (commented):
Two high-level comments:
1. Could you please run `clang-format` on the code?
2. A large part of the PR is debugging code to dump various bits of data. That is certainly helpful during initial bringup (my code certainly looked similar when developing `example2`), but I think we should discuss how much of it we want to merge (and maintain going-forward).

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
   }
 };
 
+#include "IterationStats.h"
+#include "initTracks.h"
```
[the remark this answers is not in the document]
jonapost: Thanks - indeed that was an oversight.  Part of the next step to enable it to track many electron tracks ... I will delete it for now, and include it in Example3.

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
...
+#include "ConstBzFieldStepper.h"
+
+
 constexpr double kPush = 1.e-8;
 
+// Determine the step along curved trajectory for charged particles in a field.
+//  ( Same name as as navigator method. )
+__device__  //  __host__
+double
+FieldPropagator__ComputeStepAndPropagatedState( track   & aTrack,
```
WitekPokorski, line 87: I think we should start modularizing more the current 'main' and move such functionality to separate files/modules. Shall we create 'magneticfield' directory and have all the stuff related to it there?

examples/Example2/example2.cu
```diff
@@ -52,26 +204,62 @@ __global__ void DefinePhysicalStepLength(adept::BlockData<track> *block, process
...
     // return value (if step limited by physics or geometry) not used for the moment
     // now, I know which process wins, so I add the particle to the appropriate queue
     float physics_step = proclist->GetPhysicsInteractionLength(i, block);
-    float step         = LoopNavigator::ComputeStepAndPropagatedState(mytrack.pos, mytrack.dir, physics_step,
-                                                              mytrack.current_state, mytrack.next_state);
-    mytrack.pos += (step + kPush) * mytrack.dir;
-    if (mytrack.next_state.IsOnBoundary()) {
+    float step= 0.0;
+
+    if( ! BfieldOn ) {
```
WitekPokorski, line 222: So this becomes now an embryonic transportation process. I guess, we should also modularize it and put into a separate class.

tracking/inc/trackBlock.h
```diff
@@ -0,0 +1,53 @@
...
+#define TRACKBLOCKH
+
+#include "track.h"
+
+using trackBlock_t  = adept::BlockData<track>;
+
+// static std::atomic<unsigned int> atomicTrackId= 0;
+
+__host__
+void printTracks( trackBlock_t* trackBlock,       //     adept::BlockData<track>*
```
WitekPokorski, line 14: So, the only purpose of the file is to have the print method? I would rather call it printTracks, or similar.

examples/Example2/example2.cu
jonapost: I am working on this.  My next step is to move this into a "fieldPropagatorConstBz" class, a similar "fieldPropagatorUniformB" and later one that uses Runge-Kutta.

examples/Example2/IterationStats.h
```diff
@@ -0,0 +1,145 @@
...
+      // __device__ IterationStats_dev* devicePtrHolder;
+      IterationStats_dev** devicePtrHolder;
+      cudaMallocManaged(&devicePtrHolder, sizeof(IterationStats_dev*));  // Unified memory
+      
+      MakeInstanceAt_glob<<<1,1>>>( buffer_iterStat, devicePtrHolder );
+      cudaDeviceSynchronize();
+      
+      // devPtr= ..
+      devPtr= *devicePtrHolder;
+      // cudaMemcpyFromSymbol( &devPtr, devicePtrHolder, sizeof(IterationStats_dev*) ); // If ptr was in device memory
```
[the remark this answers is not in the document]
jonapost: Great suggestion. Thanks!
Adopting.

physics/processes/inc/pair_production.h
jonapost: My problem was that the assert does not exist in non-debug runs.  I got a crash due to this - when I reduced the size of memory allocated to a small value (64K).  

I agree that the simulation must cope with this in a better way.  At least signal that it has failed (exiting in all types of runs, not just when compiled with debugging.)  

We may need a mechanism to 'throttle' the generation of new tracks.  I can imagine several ways.  One would be a kernel that reduces the population of tracks by moving only the lowest-energy tracks, which will typically range out.  This may also happen naturally nearly all the time, so long as these tracks are processed along with the others.  ( Maintaining a healthy number of live tracks is important.) 

But if there is no protection at all when the we are running in "Release" mode, I foresee problems.  Maybe a tally of the number of killed tracks and their total energy, plus an error report if this exceed a non-trivial value would be an ok first step?

tracking/inc/track.h
jonapost: I thought that it was adequate for the time being.  With only electrons and positrons tracked, it does not need to be complicated.

OK?

examples/Example2/example2.cu
jonapost: I agree - I am separating it already, so that I can use it in the test (like field_cu) and in multiple examples. 

I agree with the idea of a 'magneticfield' directory - and in fact have created it already.  It will be in one of my next commits - either in this MR or the next.

physics/processes/inc/pair_production.h
hahnjo: I would argue that having the code crash is better than silently loosing objects (and their energy).

Maybe we can just exchange the `assert` with something that aborts even in Release mode?

tracking/inc/track.h
hahnjo: Sure, it's ok for now, I just found the clever conversions from boolean values from comparisons to integer values of a charge remarkably clever :+1: 

The compiler probably emits the same code for `(pdg == -11) ? 1 : ((pdg == 11) ? -1 : 0)`, up to you which route to go (maybe constants for pdg numbers of electrons and positrons help, but maybe that's just me coming from a CS background)

agheata (changes requested):
There are several magnetic field files that would be maybe better placed in a `/magfield` folder than in this example. I've made mostly comments than change requests.

examples/Example2/example2.cu
```diff
@@ -52,26 +204,62 @@ __global__ void DefinePhysicalStepLength(adept::BlockData<track> *block, process
...
+
+    if ( mytrack.next_state.IsOnBoundary()) { // ( step == physics_step ) {  // JA-TEMP 
       // For now, just count that we hit something.
       scor->hits++;
+      mytrack.SwapStates();      
     } else {
       // Enqueue track index for later processing.
-      queues[mytrack.current_process]->push_back(i);
+      // assert(mytrack.current_process >= 0);
+      if(mytrack.current_process >= 0)  // JA-TOFIX --- take this out !!!
```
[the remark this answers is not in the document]
jonapost: That's a reasonable suggestion.  I will also set a small(er) maximum number of iterations. 
In both cases, though, there will need to be a new state of the particle such as 'transit_in_field" to show that it's integration is ongoing.   

It's like it has reached a 'virtual' boundary in the volume - it will continue in the same volume.  ( The physics processes will not have to be resampled, but if there are it is not ok. )

I will not add these to the current MR - they will come in a future refinement.

physics/processes/inc/pair_production.h
jonapost: I agree with you that this is the better solution.

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
...
+   else
+   {
+      bool fullChord= false;
+
+      //  Locate the intersection of the curved trajectory and the boundaries of the current
+      //    volume (including daughters). 
+      //  Most electron tracks are short, limited by physics interactions -- the expected
+      //    average value of iterations is small.
+      //    ( Measuring iterations to confirm the maximum. )
+      constexpr int maxChordIters= 250;
```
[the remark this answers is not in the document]
jonapost: That was the original intent.  
Jonas suggested to limit the length integrated, but instead we could change this maximum integration parameter to a small value, and output tracks which are partially integrated (as though they have met an internal virtual boundary.)

jonapost:
I have included a number of improvements from your suggestions into this update.
( No clang format yet, though, as I have some developments that still carry the current formatting.)

hahnjo:
> ( No clang format yet, though, as I have some developments that still carry the current formatting.)

If you're worried about local development branches, it usually works fine to run `clang-format` on all of them and then do a diff with respect to the (also formatted) base branch. I really think we should format all change before going in to keep a coherent code style.

hahnjo:
Things will become more complicated if we move the integration of the field propagator to an `example3`, which is a good idea IMO.

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
...
+
+         fullChord= ( move == chordLen );         
+         if( fullChord ) 
+         {
+            position=  endPosition;
+            direction= endDirection;
+         } else {
+            // Accept the intersection point on the surface (bias - TOFIX !)
+            position=  position + move * chordDir;
+         // Primitive approximation of end direction ...
```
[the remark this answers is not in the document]
jonapost: I don't understand the reference to "lateral displacement".  That usually refers to the effects of multiple scattering, which is not related to neutral particle simulation.

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
...
+         
+         // fieldPropagatorConstBz( aTrack, BzValue, endPosition, endDirection ); -- Doesn't work
+         helixBz.DoStep( position,    direction, charge, momentumMag, safeMove,
+                         endPosition, endDirection);
+      
+         vecgeom::Vector3D<double>  chordVec= endPosition - aTrack.pos;
+         double  chordLen= chordVec.Length();
+         vecgeom::Vector3D<double>  chordDir= (1.0/chordLen) * chordVec;
+         
+         double move= LoopNavigator::ComputeStepAndPropagatedState(position,
```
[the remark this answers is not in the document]
jonapost: Agreed.  

A future refinement.

examples/Example2/example2.cu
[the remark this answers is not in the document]
hahnjo: My comment was a bit misleading: Originally I wondered if this condition is necessary (based on the comment "take this out"). I now understand that, with a maximum number of iterations, this situation can arise and needs to be dealt with. I certainly don't have enough knowledge to recommend one variant over the other - it's not even clear which performance impact this will have, given that control flow will diverge anyway.

For now, it may be best to just go with the easiest solution. From an implementation perspective, that's likely an upper bound on the number of iterations as already coded. I think we can just leave this for the first version.

examples/Example2/example2.cu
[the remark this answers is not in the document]
hahnjo: See my other comment; this was meant as a question rather than a suggestion. For now, I'd start simple with a constant bound.

tracking/inc/track.h
jonapost: I agree that the code is made more readable with "constexpr pdgElec= 11".  I am adopting this.

examples/Example2/example2.cu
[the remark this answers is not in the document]
agheata: I meant the displacement of the computed intersection point against the real crossing.

physics/processes/inc/pair_production.h
jonapost: As it is a member of the track, it must mean the secondaries produced by the track.  In this sense I can see uses.
It is not clear to me what use a flag that a secondary was produced in the current step would have.

examples/Example2/example2.cu
```diff
@@ -41,8 +47,154 @@ struct Scoring {
...
+                                                                   aTrack.current_state,
+                                                                   aTrack.next_state);
+
+         fullChord= ( move == chordLen );         
+         if( fullChord ) 
+         {
+            position=  endPosition;
+            direction= endDirection;
+         } else {
+            // Accept the intersection point on the surface (bias - TOFIX !)
```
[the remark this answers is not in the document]
jonapost: Thanks.
Done in revision to fieldPropagatorConstBz.h

physics/processes/inc/pair_production.h
hahnjo: It's used to count how many secondaries have been produced globally. Pair production in the current step adds exactly one particle, not more versions of all particles ever produced. (As I said, maybe badly named and other schemes potentially make more sense, but that's what the current code does)

examples/Example2/example2.cu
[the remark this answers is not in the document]
jonapost: I have reduced the max value to 10 as a more realistic value (should likely be lower.)

Resolving for now.

examples/Example2/CMakeLists.txt
```diff
@@ -8,7 +8,9 @@ target_include_directories(example2 PUBLIC
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/physics/processes/inc>
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/tracking/inc>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/magneticfield/inc>
```
hahnjo, line 11: Is this still needed?

examples/Example4/CMakeLists.txt
```diff
@@ -0,0 +1,16 @@
+# SPDX-FileCopyrightText: 2020 CERN
+# SPDX-License-Identifier: Apache-2.0
+
+# Example 2 of particle processing with GPU, including geometry
+# and reproducible results using one RANLUX++ state per track.
```
hahnjo, line 5: Example 4 of particle processing with GPU, including geometry and magnetic field propagation.

examples/field_cu/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+set(ADEPT_COMMON_SRCS
+  field_step.cu
+)
+
+# add field_step.cu
+foreach(SRC_FILE ${ADEPT_COMMON_SRCS})
+  get_filename_component(SRC_FILENAME ${SRC_FILE} NAME_WE)
+
+  set(SRC_FILE_CPP ${CMAKE_CURRENT_SOURCE_DIR}/${SRC_FILE})
+  set(SRC_FILE_CU ${CMAKE_CURRENT_BINARY_DIR}/cuda_src/${SRC_FILENAME}.cu)
```
hahnjo, line 13: This shouldn't be needed, the file already has the extensions `.cu` so `nvcc` is happy

examples/field_cu/CMakeLists.txt
```diff
@@ -0,0 +1,54 @@
...
+
+endforeach()
+
+list(APPEND ADEPT_CUDA_SRCS
+#	field_step.cu
+)
+
+# library used for fieldStepExe.cpp
+#add_library(field_step_culib ${ADEPT_CUDA_SRCS})
+#target_include_directories(field_step_culib PUBLIC
```
hahnjo, line 30: Please remove

examples/field_cu/field_step.cu
```diff
@@ -0,0 +1,221 @@
...
+
+#if 0
+#include "IterationStats.h"
+
+// Statistics for propagation chords
+IterationStats      *chordIterStatsPBz      = nullptr;
+__device__
+IterationStats_impl *chordIterStatsPBz_impl = nullptr;
+// Needed for fieldPropagatorConstBz.h -- for now
+#endif
```
hahnjo, line 32: Is this still needed?

physics/processes/inc/pair_production.h
hahnjo: Can you please revert this change for now? As I said, it gives wrong counts with the current code.

tracking/inc/track.h
```diff
@@ -15,28 +17,153 @@ enum TrackStatus { alive, dead };
...
+   
+  char        current_process{0}; // -1 for Navigation / -2 for Field Propagation
+  char        pdg{0};             // Large enough for e/g. Then uint16_t for p/n
 
-  __device__ void SwapStates()
+  __device__ __host__ double uniform() { return rng_state.Rndm(); }
+
+  __device__ __host__ int    charge() const  // charge for e- / e+ / gamma only
+  { constexpr char pdgElec= 11;
+    int chrg= (pdg== -pdgElec ? 1 : 0)+(pdg==  pdgElec ? -1 : 0 ); return chrg; }
```
hahnjo, line 47: Can you please run `clang-format` on this PR, but at least on the modified files?

tracking/inc/track.h
```diff
@@ -15,28 +17,153 @@ enum TrackStatus { alive, dead };
...
+      
+   // auto nextTouchVolume = trk_dev.next_state.ValueAt(nextLevel);   
+}
+
+#include <cstdio>
+#include <iomanip>
+
+// inline
+__host__
+void track::print( int extId , bool verbose ) const
```
hahnjo, line 106: I'm still concerned about this debugging code in a global header file: Kernel launches using the angle brackets is not valid C++, so host code trying to include this file will probably not compile.

examples/field_cu/field_step.cu
jonapost: Yes - as field_step is for testing the field I will be using it in future.

examples/field_cu/CMakeLists.txt
jonapost: Agree.

examples/field_cu/CMakeLists.txt
jonapost: Thanks.

examples/Example4/CMakeLists.txt
jonapost: Thanks.

examples/Example2/CMakeLists.txt
jonapost: Sharp eyes! Well spotted.

tracking/inc/track.h
jonapost: I have a difficulty with naming this 'number_of_secondaries' if it is really something else.

Can we either a) adapt the scoring code to the real use case (seems relatively easy - it's one line?) or b) rename thje one for scoring and keep the name 'number_of_secondaries' for its correct meaning ?
For example, the RNG should skip by number_of_secondaries * skip_value ( = 1 << 15 now) , not a constant 1<<15 which would make sequences for subsequent tracks overlap with each other.

If it's supposed to flag whether the track has created more than zero secondaries, the current 'naive' scoring can simply check whether the 'true' number_of_secondaries is above zero

jonapost: It seems that the current use was more 'produced_seconaries_in_current_step' -- but as far as I can tell it was never cleared.
So a track that had ever produced a secondary in a previous step, was thought to have produced a secondary in the current step ...

tracking/inc/track.h
jonapost: I understand your concerns and share them.

I am open to suggestions how to organise it better, while keeping the functionality.  Can we resolve this in a future PR, though?

physics/processes/inc/pair_production.h
WitekPokorski: It's like Jonas says, the number of secondaries created by the given process. The same is for energy loss. I agree, that this should go into some 'Geant4-like particle change' struct in the long term. We need to figure out how we want the process to communicate back what has changed to the track.

hahnjo: Yes, we should change it, but we shouldn't break the current examples in this PR.

> If it's supposed to flag whether the track has created more than zero secondaries, the current 'naive' scoring can simply check whether the 'true' number_of_secondaries is above zero

Unfortunately no, because that would assume a process creates secondaries in every step after it has created at least one...

tracking/inc/track.h
hahnjo: Yes, ok. As long as it builds right now, we can iterate on this down the road.

tracking/inc/track.h
jonapost: I have revised the 'scoring' in example2.cu to try to restore the previous result - simply 'squashing' the value of number_of_secondaries: if it is > 0 it is counted as 1.
Does that restore the counting that you expected (for now?)

hahnjo: see my reply above:

> Unfortunately no, because that would assume a process creates secondaries in every step after it has created at least one...

tracking/inc/track.h
jonapost: We agreed to solve this in the future - thanks!

examples/Example2/example2.cu
jonapost: Done - see above.

physics/processes/inc/pair_production.h
jonapost: A change in pair_production was made in master that crashes if there's no slot.
Resolved.

tracking/inc/track.h
jonapost: I revised the code in example2 to fetch the previous number of secondaries, before an interaction, and the current after it.  It will add the number of secondaries (only if it is positive - to avoid adding zero. )  
Changed energy_loss.h (no longer zeroes number_of_secondaries) and pair_production.h (uses ++).
This now reproduces the old output.

hahnjo:
CI says (https://lcgapp-services.cern.ch/spi-jenkins/job/AdePT-CI/24/console):
```
[ 31%] Building CUDA object examples/Example1/CMakeFiles/example1.dir/example1.cu.o
ptxas fatal   : Unresolved extern function '_ZN7vecgeom20globaldevicegeomdata11GetNavIndexEv'
```

jonapost:
@hahnjo - thanks for the heads up.  I looked but could not find the issue in Jenkins.

examples/Example2/example2.cu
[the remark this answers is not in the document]
jonapost: Ensuring that tracking works correctly when field propagation reaches a maximum number of iterations remains an issue.  I will tackle it in a future revision.

tracking/inc/trackBlock.h
jonapost: I addressed this.   I renamed the file 'trackBlock.h' to 'printTracks.h'

physics/processes/inc/pair_production.h
jonapost: This is now be fixed - the output of example2 is reproduced as before.

examples/Example2/IterationStats.h
```diff
@@ -0,0 +1,145 @@
...
+      cudaFree(devicePtrHolder);
+   }
+   
+   ~IterationStats()  {
+      cudaFree(devPtr);
+   }
+
+   __host__   /*unsigned*/ int GetMax(){ /*unsigned*/ int maxNow=0;   // Was GetMaxFromDevice()
+     cudaMemcpyFromSymbol( &maxNow,
+                           &(devPtr->maxDone), sizeof( /*unsigned*/ int) ); return maxNow; }
```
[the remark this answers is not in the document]
jonapost: What I can say is that it worked.  It still exists, and for me it was clearer, as the method name explains what it does - I don't need to look at the final argument. 

But I have replaced all its instances with 'cudaMemcpy'.

agheata (approved):

PR #64, opened 2021-01-19
Abort simulation if secondary cannot be produced

hahnjo:
The `assert()` would only be present for Debug builds, leaving Release builds with an "illegal memory access" due to accessing nullptr. Use the existing `COPCORE_EXCEPTION` and extend it to print a message on the device. This still isn't perfect because all threads will print, but better than developers wondering what happened.


hahnjo:
@agheata @jonapost ping, does this make sense?

agheata (approved):
Sure it makes sense

PR #76, opened 2021-02-03
Add example for basic interaction with G4HepEm

hahnjo:
Process selection as well as continuous and discrete effects on the kinetic energy of one single electron.


hahnjo:
One thing we should probably discuss early (hence this PR) is how to "pin" the version of G4HepEm. Implementation of MSC will likely require some changes to the interfaces, so during R&D it might be easiest to add G4HepEm as a submodule and build a specified commit along with AdePT. Other ideas?

examples/Example7/example7.cu
```diff
@@ -0,0 +1,233 @@
...
+  CopyG4HepEmDataToGPU(data);
+
+  // Create G4HepEmData with the device pointers.
+  G4HepEmData dataOnDevice;
+  dataOnDevice.fTheMatCutData   = data->fTheMatCutData_gpu;
+  dataOnDevice.fTheMaterialData = data->fTheMaterialData_gpu;
+  dataOnDevice.fTheElementData  = data->fTheElementData_gpu;
+  dataOnDevice.fTheElectronData = data->fTheElectronData_gpu;
+  dataOnDevice.fThePositronData = data->fThePositronData_gpu;
+  // fTheSBTableData missing for now.
```
mnovak42, line 104: It's available now on the device in `G4HepEm`. However,  this data is used only in the SB bremsstrahlung  interaction model, as we discussed, so I just wanted to let you know.

hahnjo: Great, didn't see yesterday's commit! Yes, I'll probably start with final state sampling of ionization and keep you posted on how that works out...

mnovak42:
I don't know if someone has already started to work on writing/reading the `G4HepEmData` into/from file(s), but this example could be greatly simplified and made much more generic by 
 - writing the `G4HepEmData` from an already initialised Geant4-G4HepEm application (e.g. [the G4HepEm TestEm3 example application](https://github.com/mnovak42/g4hepem/tree/master/apps/examples/TestEm3) or any of the [G4HepEm tests](https://github.com/mnovak42/g4hepem/tree/master/apps/tests))
- reading those `G4HepEmData`  from the file(s) in this above type of tests and specify the G4HepEm material-cut index. 

This would save all those duplicated `Geant4` and `G4HepEm` codes related to the initialisation, make the whole example more generic and the corresponding codes (write/read `G4HepEmData`) need to be developed anyway (for `AdePT`).  

However, I understand that this example is (at the moment) more about to check the feasibility of reusing the `G4HepEm::HowFar` implementation on the device for the step limit computations.

agheata (approved):
I think this is a nice demonstrator that we can use G4HepEm types directly in our kernels and use the physics tables, which is essential. I noticed that the interfaces are now forcing using G4HepEMxxxTrack, while we may want to only pass the relevant state to the sampling methods. In addition, we may not want to initialize a full empty track on the stack of every GPU thread. Is it difficult to change the interfaces in this spirit?

agheata:
Having G4HepEm as a temporary sub-module for keeping things in sync sounds like a better alternative to having to go manually to a specific commit number, so 👍

hahnjo:
> I noticed that the interfaces are now forcing using G4HepEMxxxTrack, while we may want to only pass the relevant state to the sampling methods. In addition, we may not want to initialize a full empty track on the stack of every GPU thread. Is it difficult to change the interfaces in this spirit?

Yes, that's one of the not-so-nice parts of the current approach. However, it may be less severe than it initially sounds because the compiler is able to inline everything and propagate only the needed values. So in the end, I guess no object will actually end up in memory. But yes, we should certainly look into this once we have a fully working demonstrator.

mnovak42:
If you want to be free form the `G4HepEmTrack` (that higher level `G4HepEm` functions like `HowFar` relies on) then probably better to write your own `HowFar` function on your side and use only the data interpolation methods from `G4HepEm`. Those interpolation method has only primitive data types as arguments.  It should be very easy based on the `G4HepEm` implementation and even right now the related code is already duplicated on the `G4HepEm` side.

examples/Example7/example7.cu
```diff
@@ -0,0 +1,233 @@
...
+  for (int i = 0; i < 20; i++) {
+    printf("-----------------------------------------\n");
+    // Sample the `number-of-interaction-left`.
+    for (int ip = 0; ip < 3; ++ip) {
+      if (theTrack->GetNumIALeft(ip) <= 0.) {
+        theTrack->SetNumIALeft(-std::log(r.Rndm()), ip);
+      }
+    }
+
+    electronManager.HowFar(&g4HepEmData, &g4HepEmPars, theElTrack);
```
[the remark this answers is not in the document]
mnovak42: Just a minor correction, the material-cut index is passed in the `G4HepEmTrack`. `G4HepEmParameters` stores some global configuration options.

hahnjo:
With the latest code from https://github.com/hahnjo/AdePT/tree/g4hepem (not sure why it doesn't show up in the PR):
```
 === G4HepEm global init ... 
 === G4HepEm init for particle index = 0 ...
     ---  InitElectronData ... 
     ---  BuildELossTables ...
     ---  BuildLambdaTables ... 
     ---  BuildElementSelectorTables ... 
     ---  BuildSBBremTables ... 
fNumG4MatCuts = 2, fNumMatCutData = 2
Launching particle transport on GPU
-----------------------------------------
Starting with 100000.000000MeV
-----------------------------------------
sampled process: 1, particle travels 0.918880mm
energy after continuous process: 99999.669322MeV
energy after discrete process:   87934.992027MeV
-----------------------------------------
[...]
-----------------------------------------
sampled process: 1, particle travels 1.483651mm
energy after continuous process: 123.264522MeV
energy after discrete process:   0.000000MeV
-----------------------------------------
sampled process: -1, particle travels 0.000005mm
energy after continuous process: 0.000000MeV
```

and the electron is stopped after 80 steps. Credits to Mihaly and his library, I just copied code around and wrote some glue code...

hahnjo:
> someone has already started to work on writing/reading the G4HepEmData into/from file(s)

Ben started looking into this and I agree we need this for full applications. This example is more a playground how the integration could work.

> If you want to be free form the G4HepEmTrack (that higher level G4HepEm functions like HowFar relies on) then probably better to write your own HowFar function on your side and use only the data interpolation methods from G4HepEm. Those interpolation methods have only primitive data types as arguments. It should be very easy based on the G4HepEm implementation and even right now the related code is already duplicated on the G4HepEm side.

Of course it's possible to duplicate large parts from G4HepEm. For a proof-of-concept I did just that for the energy transfer sampling of the discrete processes, but I really dislike copying hundreds of lines. Writing some glue code and interfaces to our data structures is fine, but IMHO all the physics computations should stay in G4HepEm as much as possible. John proposed that we abstract the random number generation into a global function and "inject" a suitable implementation for the GPU. That could work as long as we get around passing `G4HepEmTLData`, but we'll see...

mnovak42:
Ok then we might try to think about the following:
 - **abstract random  number generator** such that G4HepEm can be built either with the native Geant4 random engine or with the one that is used by adept (but these details are hidden from the G4HepEm interaction implementations behind a common interface or simple typedef etc.)
 - I can try to **split into 3 parts** each of *the G4HepEm interaction implementations*:
   1.  input: depends on the actual track structure and other possible input argument objects    
   2. the core: the main part of the interaction generation, i.e. sampling according to the given probability densities; this should have only primitive types as input/output 
   3. output: depends again on the actual track structure and this contains the post-interaction update of the primary particle track and generation of the secondary track(s) using the output from the previous (3.) step (provided in pure primitive types)

Points 1. and 2. would be small portion of codes specific for the environment, i.e. **separate and different implementations on the `G4HepEm` and `AdePT` sides**  (but the difference would be only due to the different track representations). The main and (usually) big part (3.), that contains the model related complications, would have only **one implementation on the G4HepEm side** that could be **reused** in this way **on the GPU** (with building the appropriate random engine). 
 
These would minimise the code duplications on the `AdePT` side and would probably cause not a big change on my (`G4HepEm` side).

hahnjo:
> the core: the main part of the interaction generation, i.e. sampling according to the given probability densities; this should have only primitive types as input/output

I had a look at this and this leads to very long argument lists, like:
```c++
void HowFar(struct G4HepEmData* hepEmData, struct G4HepEmParameters* hepEmPars, double theEkin, double theLEkin,
            int theIMC, bool isElectron, const double* numIALeft, double* MFP, double* rangeOut,
            double* pStepLengthOut, double* gStepLengthOut, int* indxWinnerProcessOut);
bool PerformContinuous(struct G4HepEmData* hepEmData, struct G4HepEmParameters* hepEmPars, double theEkin,
                       double theLEkin, int theIMC, bool isElectron, double theRange, double pStepLength,
                       double* numInterALeft, const double* preStepMFP, double* elossOut);
bool CheckDelta(struct G4HepEmData* hepEmData, double theEkin, double theLEkin, int theIMC, bool isElectron,
                int iDProc, double mfp, double rand);
```
which isn't really the easiest interface to work with. Moreover it seems that this has a negative performance impact on `TestEm3`... Let me know if you want me to publish the changes somewhere, but I'm not really convinced by the early results.

agheata:
> > the core: the main part of the interaction generation, i.e. sampling according to the given probability densities; this should have only primitive types as input/output
> 
> I had a look at this and this leads to very long argument lists, like:
> 
> ```c++
> void HowFar(struct G4HepEmData* hepEmData, struct G4HepEmParameters* hepEmPars, double theEkin, double theLEkin,
>...
> ```
> 
> which isn't really the easiest interface to work with. Moreover it seems that this has a negative performance impact on `TestEm3`... Let me know if you want me to publish the changes somewhere, but I'm not really convinced by the early results.

Maybe we should try something intermediate as exchange format, that can be a native G4EMHep type, but not a track. We could have a simple data structure to exchange with specific functionality. For example:
```
struct HowFarExchange { 
            double theEkin, double theLEkin,
            int theIMC, bool isElectron, const double numIALeft, double MFP, double rangeOut,
            double pStepLengthOut, double gStepLengthOut, int indxWinnerProcessOut
}
```
Then we can create such an object on the stack and pass it to HowFar. This simplifies the interface, keeps the data in the cache (even it need some copying from the AdePT track) and is independent of the track model. This could be exposed as an interface by G4EmHep

hahnjo:
> Maybe we should try something intermediate as exchange format, that can be a native G4EMHep type, but not a track. We could have a simple data structure to exchange with specific functionality. For example:
> 
> ```
> struct HowFarExchange { 
>             double theEkin, double theLEkin,
>             int theIMC, bool isElectron, const double numIALeft, double MFP, double rangeOut,
>             double pStepLengthOut, double gStepLengthOut, int indxWinnerProcessOut
> }
> ```
> 
> Then we can create such an object on the stack and pass it to HowFar. This simplifies the interface, keeps the data in the cache (even it need some copying from the AdePT track) and is independent of the track model. This could be exposed as an interface by G4EmHep

Then there's no difference to using the existing data structures. Neither of this will really end up on the stack anyway.

mnovak42:
> > the core: the main part of the interaction generation, i.e. sampling according to the given probability densities; this should have only primitive types as input/output
> 
> I had a look at this and this leads to very long argument lists, like:
> 
> ```c++
> void HowFar(struct G4HepEmData* hepEmData, struct G4HepEmParameters* hepEmPars, double theEkin, double theLEkin,
>             int theIMC, bool isElectron, const double* numIALeft, double* MFP, double* rangeOut,
>             double* pStepLengthOut, double* gStepLengthOut, int* indxWinnerProcessOut);
> bool PerformContinuous(struct G4HepEmData* hepEmData, struct G4HepEmParameters* hepEmPars, double theEkin,
>                        double theLEkin, int theIMC, bool isElectron, double theRange, double pStepLength,
>                        double* numInterALeft, const double* preStepMFP, double* elossOut);
> bool CheckDelta(struct G4HepEmData* hepEmData, double theEkin, double theLEkin, int theIMC, bool isElectron,
>                 int iDProc, double mfp, double rand);
> ```
> 
> which isn't really the easiest interface to work with. Moreover it seems that this has a negative performance impact on `TestEm3`... Let me know if you want me to publish the changes somewhere, but I'm not really convinced by the early results.

I meant not the determination of the physics step length i.e. not the HowFar but the performing the commutation of the interactions i.e. Perform: BremSB, BremRB, Ioni and all those that will come.

agheata:
> Then there's no difference to using the existing data structures. Neither of this will really end up on the stack anyway.

Yes, because this does not contain info about position/direction/... that is not needed and no need to take space on the stack nor to initialize. But maybe we should chat about all this rather than commenting in the PR...

mnovak42:
> > Then there's no difference to using the existing data structures. Neither of this will really end up on the stack anyway.
> 
> Yes, because this does not contain info about position/direction/... that is not needed and no need to take space on the stack nor to initialize. But maybe we should chat about all this rather than commenting in the PR...

And those are exactly what will be needed (and even more) when MSC will be included.

hahnjo:
> [...] the electron is stopped after 80 steps.

For people following around, it now takes 108 steps with a recent fix in G4HepEm. I'm attaching the output I'm getting: [output_die.txt](https://github.com/apt-sim/AdePT/files/5942643/output_die.txt)

PR #82, opened 2021-02-11
Adapted the system of units to Geant4, using PDG 2019 values.

agheata:
Adapted to changes done in 17.07.20 in CLHEP

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -49,7 +49,7 @@ find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
 
 # Find VecGeom geometry headers library
-set(VecGeom_VERSION 1.1.7)
+set(VecGeom_VERSION 1.1.11)
 find_package(VecGeom ${VecGeom_VERSION} REQUIRED)
 message(STATUS "Using VecGeom version ${VecGeom_VERSION}")
 # make sure we import VecGeom architecture flags - is this needed?
--- a/base/inc/CopCore/include/CopCore/PhysicalConstants.h
+++ b/base/inc/CopCore/include/CopCore/PhysicalConstants.h
@@ -23,6 +23,7 @@
  * \li  steradian               (steradian)
  *
  * You can add your own constants.
+ * // CLHEP 17.07.20   use PDG 2019 values
  *
  */
 
@@ -37,7 +38,7 @@ namespace units {
 //
 //
 //
-static constexpr double kAvogadro = 6.02214179e+23 / mole;
+static constexpr double kAvogadro = 6.02214076e+23 / mole;
 
 //
 // kCLight       ->   c   = 2.9979246e+10 [cm/s]
@@ -47,11 +48,11 @@ static constexpr double kCLight       = 2.99792458e+8 * m / s;
 static constexpr double kCLightSquare = kCLight * kCLight;
 
 //
-// kHPlanck          -> h        = 4.1356673e-24 [GeV*s]
-// kHBarPlanck       -> h/(2Pi)  = 6.582119e-25  [GeV*s]
-// kHBarPlanckCLight -> hc/(2Pi) = 1.9732696e-14 [GeV*cm]
+// kHPlanck          -> h        = 4.13566e-12   [MeV * ns]
+// kHBarPlanck       -> h/(2Pi)  = 6.58212e-13   [MeV * ns]
+// kHBarPlanckCLight -> hc/(2Pi) = 197.32705e-12 [MeV * mm]
 //
-static constexpr double kHPlanck                = 6.62606896e-34 * joule * s;
+static constexpr double kHPlanck                = 6.62607015e-34 * joule * s;
 static constexpr double kHBarPlanck             = kHPlanck / kTwoPi;
 static constexpr double kHBarPlanckCLight       = kHBarPlanck * kCLight;
 static constexpr double kHBarPlanckCLightSquare = kHBarPlanckCLight * kHBarPlanckCLight;
@@ -96,7 +97,7 @@ static constexpr double kBohrRadius                = kRedElectronComptonWLenght 
 //
 //
 //
-static constexpr double kBoltzmann = 8.617343e-11 * MeV / kelvin;
+static constexpr double kBoltzmann = 8.617333e-11 * MeV / kelvin;
 
 //
 // STP -> Standard Temperature and Pressure
--- a/base/inc/CopCore/include/CopCore/SystemOfUnits.h
+++ b/base/inc/CopCore/include/CopCore/SystemOfUnits.h
@@ -50,15 +50,15 @@ static constexpr double kPiSquare = kPi * kPi;
 //
 // Length [L]
 //
-static constexpr double centimeter  = 1.;
-static constexpr double centimeter2 = centimeter * centimeter;
-static constexpr double centimeter3 = centimeter * centimeter * centimeter;
-
-static constexpr double millimeter  = 0.1 * centimeter;
+static constexpr double millimeter  = 1.;
 static constexpr double millimeter2 = millimeter * millimeter;
 static constexpr double millimeter3 = millimeter * millimeter * millimeter;
 
-static constexpr double meter  = 100. * centimeter;
+static constexpr double centimeter  = 10. * millimeter;
+static constexpr double centimeter2 = centimeter * centimeter;
+static constexpr double centimeter3 = centimeter * centimeter * centimeter;
+
+static constexpr double meter  = 1000. * millimeter;
 static constexpr double meter2 = meter * meter;
 static constexpr double meter3 = meter * meter * meter;
 
@@ -125,10 +125,10 @@ static constexpr double deg  = degree;
 //
 // Time [T]
 //
-static constexpr double second      = 1.;
+static constexpr double nanosecond  = 1.;
+static constexpr double second      = 1.e+9 * nanosecond;
 static constexpr double millisecond = 1.e-3 * second;
 static constexpr double microsecond = 1.e-6 * second;
-static constexpr double nanosecond  = 1.e-9 * second;
 static constexpr double picosecond  = 1.e-12 * second;
 
 static constexpr double hertz     = 1. / second;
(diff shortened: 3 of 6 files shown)


hahnjo:
Without the discussed changes in VecGeom, we'll now run into trouble for the magnetic field, right?

agheata:
> Without the discussed changes in VecGeom, we'll now run into trouble for the magnetic field, right?

Correct, I'm doing the necessary change in VecGeom, let's wait until that one is merged first.

agheata:
OK, the change in VecGeom is now merged in the master. This can be merged as well, but to avoid compilation errors before adopting the VecGeom master, I commented out in the examples the lines that need to be activated before GDML loading for the units to be taken in mm.
```  
  // The vecgeom unit has to be changed before loading the geometry
  // vecgeom::GeoManager::SetMillimeterUnit(1.); // activate this line after migrating to VecGeom master
```

hahnjo:
> OK, the change in VecGeom is now merged in the master. This can be merged as well, but to avoid compilation errors before adopting the VecGeom master, I commented out in the examples the lines that need to be activated before GDML loading for the units to be taken in mm.

I don't think it's a good idea to get wrong results from the examples by default. So there is no other way than using the new function?

agheata:
> > OK, the change in VecGeom is now merged in the master. This can be merged as well, but to avoid compilation errors before adopting the VecGeom master, I commented out in the examples the lines that need to be activated before GDML loading for the units to be taken in mm.
> 
> I don't think it's a good idea to get wrong results from the examples by default. So there is no other way than using the new function?

No, because VecGeom API does not provide units. When you ask for a length you get what was computed based on the numbers stored internally in VecGeom. The user has to decide if these represent cm/mm/something else. What I added is just a way to force VecGeom store a given value as internal representation for millimeter. But this does not make the examples fail by default, it just forces defining the VecGeom length unit, and then getting it back naturally w/o having to do runtime conversions.

hahnjo:
Without the call, it will give different results when changing the units as done in this PR

agheata:
> Without the call, it will give different results when changing the units as done in this PR

Yes, because VecGeom is units aware now (and has an internal unit cm which was before just implicit) and you would be misusing it if you assumed that the default unit was mm.

agheata:
Can you think of a better alternative?

hahnjo:
I think we should do either of:
1. Use the new API you added to VecGeom (after `cm` is fixed). Requires either `master` or asking for a new release.
2. Merge this PR and transform the units before calling VecGeom. Should be relatively straight-forward in the `LoopNavigator`.
3. Wait until we can do 1. and until then, transform the units before calling G4HepEm. That's what I currently have for the upcoming `example6`.

agheata:
> I think we should do either of:
> 
> 1. Use the new API you added to VecGeom (after `cm` is fixed). Requires either `master` or asking for a new release.
> 2. Merge this PR and transform the units before calling VecGeom. Should be relatively straight-forward in the `LoopNavigator`.
> 3. Wait until we can do 1. and until then, transform the units before calling G4HepEm. That's what I currently have for the upcoming `example6`.

I think 1. can be done fast and requires fewer steps. I bound now setting the internal VecGeom unit to the vgdml::Frontend::Load method because it really depends on the method we use to create the geometry. Default is still mm=0.1, so we need to pass mm=1 as last parameter to the Load method for AdePT. We'll have a tag in VecGeom and change the minimum requirement in our cmake to match this (with this PR), then we can merge this safely. It might be that we're going to get errors in the CI if we don't update also there the VecGeom we build against. But we have to wait for the VecGeom tag first.

agheata:
Indeed:
CMake Error at CMakeLists.txt:53 (find_package):
  Could not find a configuration file for package "VecGeom" that is
  compatible with requested version "1.1.10".

  The following configuration files were considered but not accepted:

    /cvmfs/sft-nightlies.cern.ch/lcg/views/devAdePT/Fri/x86_64-centos7-gcc8-opt/lib64/cmake/VecGeom/VecGeomConfig.cmake, version: 1.1.9

@peremato can you boost the version of vecgeom in the nightlies to 1.1.10?

hahnjo:
Note that 1.1.10 incorrectly treats `cm` as `mm` (see https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/785/diffs#note_4176659), so we need to fix that and wait for 1.1.11

agheata:
> Note that 1.1.10 incorrectly treats `cm` as `mm` (see https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/785/diffs#note_4176659), so we need to fix that and wait for 1.1.11

@hahnjo I saw your comment in VecGeom late, but the change was intentional. The convention is to interpret missing units in GDML as "mm" and the old behavior was inconsistent.

hahnjo:
> > Note that 1.1.10 incorrectly treats `cm` as `mm` (see https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/785/diffs#note_4176659), so we need to fix that and wait for 1.1.11
> 
> @hahnjo I saw your comment in VecGeom late, but the change was intentional. The convention is to interpret missing units in GDML as "mm" and the old behavior was inconsistent.

Then we need to add `cm` to the list of handled units. Just treating it as `mm` is definitely wrong and will be disastrous with `trackML.gdml`.

peremato:
bot build

agheata:
OK, nice spot @hahnjo, we have already 1.1.11 fixing this, and @peremato is making a VecGeom build using this

peremato:
bot build

agheata:
OK, I think this is ready to merge now

hahnjo:
Hm, I'm getting different results from `example2` after this PR. For one, that's because we have to scale up the random values in the processes by a factor of 10 as well as the position:
```diff
diff --git a/examples/Example2/example2.cu b/examples/Example2/example2.cu
index 7b07003..aefbd58 100644
--- a/examples/Example2/example2.cu
+++ b/examples/Example2/example2.cu
@@ -191,7 +191,7 @@ void example2(const vecgeom::cxx::VPlacedVolume *world)
   track2->energy      = 30.0f;
   track2->energy_loss = 0.0f;
   //  track2->index = 2; // this is not use for the moment, but it should be a unique track index
-  track2->pos = {0, 0.05, 0};
+  track2->pos = {0, 0.5, 0};
   track2->dir = {1, 0, 0};
   init_track<<<1, 1>>>(track2, gpu_world);
   COPCORE_CUDA_CHECK(cudaDeviceSynchronize());
diff --git a/physics/processes/inc/energy_loss.h b/physics/processes/inc/energy_loss.h
index 2915fce..223e6a5 100644
--- a/physics/processes/inc/energy_loss.h
+++ b/physics/processes/inc/energy_loss.h
@@ -17,7 +17,7 @@ __device__ float energy_loss::GetPhysicsInteractionLength(int particle_index, ad
 {
   track *mytrack = &((*block)[particle_index]);
   // here I will need to calculate the IL based on the particle energy, material, etc
-  float current_length = mytrack->uniform() * 0.2f;
+  float current_length = mytrack->uniform() * 2.0f;
   return current_length;
 }

diff --git a/physics/processes/inc/pair_production.h b/physics/processes/inc/pair_production.h
index 0e52f78..b5a3a38 100644
--- a/physics/processes/inc/pair_production.h
+++ b/physics/processes/inc/pair_production.h
@@ -17,7 +17,7 @@ __device__ float pair_production::GetPhysicsInteractionLength(int particle_index
 {
   track *mytrack = &((*block)[particle_index]);
   // here I will need to calculate the IL based on the particle energy, material, etc
-  float current_length = mytrack->uniform() * 0.25f;
+  float current_length = mytrack->uniform() * 2.5f;
   return current_length;
 }

```

But even then, I get a different number of secondaries and number of hits...

(`example4` doesn't reproduce either, but I think that's a different reason...)

agheata:
Can you track down why it diverges maybe by limiting the number of tracks?

hahnjo:
We can avoid the difference in `example2` with
```diff
diff --git a/examples/Example2/example2.cu b/examples/Example2/example2.cu
index 7b07003..050748d 100644
--- a/examples/Example2/example2.cu
+++ b/examples/Example2/example2.cu
@@ -39,7 +39,7 @@ struct Scoring {
   }
 };

-constexpr double kPush = 1.e-8;
+constexpr double kPush = 1.e-7;

 // kernel select processes based on interaction lenght and put particles in the appropriate queues
 __global__ void DefinePhysicalStepLength(adept::BlockData<track> *block, process_list *proclist,
```
(which is odd enough that this makes a difference in the number of hits - it really shouldn't!)

agheata:
> (which is odd enough that this makes a difference in the number of hits - it really shouldn't!)

If this causes extra or fewer calls to the rng, it is not. Can you count the calls to rng in an easy way in the 2 cases?

hahnjo:
> > (which is odd enough that this makes a difference in the number of hits - it really shouldn't!)
> 
> If this causes extra or fewer calls to the rng, it is not. Can you count the calls to rng in an easy way in the 2 cases?

This is the amount that is pushed into a new volume, if this results in different call graph then that is still bad. But I didn't check, I'm currently looking into `example4` (which still doesn't reproduce even with #85). Probably because of different energies in the magnetic field. I'm trying to hunt down all constants that didn't have a unit before, and I think we should really understand the difference or we risk breaking some physics computation in the magnetic field...

agheata:
> > > (which is odd enough that this makes a difference in the number of hits - it really shouldn't!)
> > 
> > 
> > If this causes extra or fewer calls to the rng, it is not. Can you count the calls to rng in an easy way in the 2 cases?
> 
> This is the amount that is pushed into a new volume, if this results in different call graph then that is still bad. But I didn't check, I'm currently looking into `example4` (which still doesn't reproduce even with #85). Probably because of different energies in the magnetic field. I'm trying to hunt down all constants that didn't have a unit before, and I think we should really understand the difference or we risk breaking some physics computation in the magnetic field...

The amount of a push actually can influence the call graph, in one case the track may be passing through a corner in the other it can actually enter the new volume, triggering computation of physics. So this is not bad if changing the tolerance to be unit-aware fixes the divergence. I agree though that we have to understand divergencies if any, before merging,

agheata:
Note that this PR slightly changes also the electron charge, so the trajectories in field are a bit different, which may be enough to get a divergence.

agheata:
This is now rebased on the master. Example4 shows now no divergence on my machine before and after applying this. @jonapost can you review this?

WitekPokorski (approved):
looks ok to me

hahnjo:
FWIW I've posted the proposed changes I already mentioned above in #96 to restore the output of the existing examples.

PR #84, opened 2021-02-12
Add realistic demonstrator with physics processes

hahnjo:
Example 6 uses one container for all tracks, includes geometry, magnetic field with constant `Bz`, and physics processes for e-/e+. Photons immediately pair-produce if allowed or deposit their energy.


hahnjo:
Here's the output I get: [example6.txt](https://github.com/apt-sim/AdePT/files/5987301/example6.txt) This should be reproducible up to rounding errors in the energy deposit.

For merging, please don't squash: the first two commits are unrelated fixes that should not be melded with `example6`.

agheata (approved):

agheata:
bot build

agheata:
Why is the build pending, anybody knows?

PR #85, opened 2021-02-12
Initialize PDG for secondary tracks

hahnjo:
This makes `example4` reproducible because the particle's charge decides if it is subjected to the magnetic field.


agheata (approved):

agheata:
that is fine, but does this reproduce when dealing with the new units?

PR #86, opened 2021-02-17
example6: Split and parallelize relocation to next volume

hahnjo:
When a particle leaves the current volume or enters a daughter, we need to relocate to the next volume. This could involve walking the hierarchy which is expensive, at least in the current `LoopNavigator`. The problem is that if only one particle needs relocation, all others have to wait.
As a first step, move the relocation step out of navigation, which is called from the magnetic field, to a common place when a particle is detected to need it. This already improves the total runtime from around 1.4s to 1.3s on my system.

---

As there are usually fewer particles to be relocated than there are threads on the GPU, we can use one level of parallelism to check daughter volumes in parallel. This improves the total runtime from around 1.3s to 1.2s on my system.

---

*Note*: I think it would be good to merge this as two commits (rather than squashing) so that we could go back later and check incremental improvements for other setups etc.

### DIFF
--- a/examples/Example6/example6.cu
+++ b/examples/Example6/example6.cu
@@ -182,8 +182,8 @@ __device__ struct G4HepEmElectronManager electronManager;
 // Compute the physics and geomtry step limit, transport the particle while
 // applying the continuous effects and possibly select a discrete process that
 // could generate secondaries.
-__global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *discreteQueue, Scoring *scoring,
-                            int maxIndex)
+__global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *relocateQueue,
+                            adept::MParray *discreteQueue, Scoring *scoring, int maxIndex)
 {
   fieldPropagatorConstBz fieldPropagatorBz(BzFieldValue);
 
@@ -280,7 +280,7 @@ __global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *
       // For now, just count that we hit something.
       scoring->hits++;
 
-      LoopNavigator::RelocateToNextVolume(currentTrack.pos, currentTrack.dir, currentTrack.next_state);
+      relocateQueue->push_back(i);
 
       // Move to the next boundary.
       currentTrack.SwapStates();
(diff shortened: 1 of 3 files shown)


examples/Example6/example6.cu
```diff
@@ -304,6 +306,89 @@ __global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *
...
+      bool godeeper;
+      do {
+        godeeper = false;
+
+        vecgeom::Vector3D<vecgeom::Precision> localPoint(localCoordinates[0], localCoordinates[1], localCoordinates[2]);
+
+        const auto &daughters = currentVolume->GetDaughters();
+        auto daughtersSize    = daughters.size();
+        vecgeom::Vector3D<vecgeom::Precision> transformedPoint;
+        vecgeom::VPlacedVolume const *nextVolume = nullptr;
```
agheata, line 357:
```suggestion
        vecgeom::VPlacedVolume const *nextVolume = nullptr;
        // Each thread checks relocation in different daughters
```

examples/Example6/example6.cu
```diff
@@ -304,6 +306,89 @@ __global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *
...
+          const auto *daughter = daughters[d];
+          if (daughter->Contains(localPoint, transformedPoint)) {
+            nextVolume = daughter;
+            break;
+          }
+        }
+
+        auto *previousVolume = currentVolume;
+        __syncthreads();
+        if (nextVolume) {
```
agheata, line 368:
```suggestion
        if (nextVolume) {
          // The thread that relocated successfully updates the coordinates in the new local system
```

examples/Example6/example6.cu
```diff
@@ -304,6 +306,89 @@ __global__ void PerformStep(adept::BlockData<track> *allTracks, adept::MParray *
...
+          localCoordinates[0] = transformedPoint.x();
+          localCoordinates[1] = transformedPoint.y();
+          localCoordinates[2] = transformedPoint.z();
+
+          currentVolume = nextVolume;
+          state.Push(currentVolume);
```
agheata, line 374: Note that this leads to undefined behavior in case two of the daughters are overlapping, and the propagated point is in the overlapping region. In the sequential version, always the same daughter wins. To have the same here maybe the `currentVolume` and `localCoordinates` need to be atomic.

agheata (changes requested):
Using parallelism for relocation is a good idea, we should be careful though for undefined behavior in case of overlaps. I felt like needing some comments in the code, I suggested some myself but maybe more are needed (especially for describing what the new top-level navigation methods in LoopNavigator are doing)

examples/Example6/example6.cu
agheata: ... but even this will not eliminate undefined behavior

hahnjo: My assumption was that
1. a volume strictly contains all daughters, and
2. all daughters of a given volume are non-intersecting.

So you're saying that the second point is not correct? If so, what is the motivation for using the first daughter that contains a point? After all, using an acceleration structure might change ordering. Or do we just need to pick one (and always the same)?

agheata: In fact, this happens only in the case of geometry definition errors, but each large LHC geometry has many of those. Picking any of the candidates in such a case is equivalent (both bad), what matters is that the order is reproducible, otherwise the entire simulation becomes non-reproducible. Even the first point assumption can be broken, but the simulation still has to 'survive' (with the associated errors of course)

hahnjo: I've implemented a scheme that works at warp level without atomics (and without shared variables). You can launch the kernel with more than one warp (and each of them will relocate one track) or incomplete warps, but `relocateThreads  = 32` still results in the best performance.

agheata (approved):
Much better!

PR #87, opened 2021-02-17
Example7 for export and import of geometry/physics data between Geant4 and AdePT

drbenmorgan:
To decouple AdePT test programs from Geant4 as much as possible in the testing phase, and to avoid potentially expensive recomputation of physics data at application initialization, there is a use case to

- Persist geometry and physics data from a standard Geant4 application to file
- Read this file in a pure AdePT based application, reconsitituting the geometry and physics data for use on device

GDML/vecgeom is used for the geometry part, and [G4HepEM](https://github.com/mnovak42/g4hepem) for physics data.

This is a WIP to provide these "example7" applications for the Geant4 (data export) and AdePT (data import) parts. It's a pure placeholder at present to enable ongoing discussion/requirements.


drbenmorgan:
The latest update is a very noddy implementation of a _mock_ Geant4 setup just to get the Geant4+export side working and see what we have. This can be updated to the full TestEm3 later on, but for now I wanted to keep things simple for the next step which is adding G4HepEM on the Geant4 side and then exploring how to export the data.

NB: one thing I've added is a second `G4Region` with different production cuts so we can explore how that affects things.

Just pinging @mnovak42 for info (I can't seem to add you as reviewer, but you should be!) and input!

drbenmorgan:
... and the basic mocked geometry/regions/cuts are exported to a gdml file, so we can already see what relevant info is there already.

mnovak42:
I'm a bit confused here since I don't see the reason why this development is done here. I mean, for me it would be more obvious to develop Read/Write functionalities on objects where they are defined.

In more details
---------------

`AdePT` relies on `G4HepEm` regarding the physics simulation:
 1.  all physics simulation, material related, etc. data since `AdePT` doesn't have  these concepts.  
 2. all the run time physics simulation functionalities that depend on these above data structures 

Additionally, `AdePT` requires a geometry and some association of the volumes to a `G4HepEm`
material-cut index.  


**All these above can be extracted from a `Geant4` application (connected with `G4HepEm`:**
 1. the `G4HepEm` data are initialised, then they can be (should be in the future) written into files:
    - after setting a singe process for `e-/e+` and `gamma` particles in the physics list of the `Geant4` application
    -  or simple copy the physics list of  `G4HepEm: /apps/examples/TestEm3/src/PhysListHepEm` 
 2. the association, between the `G4HepEm` data and geometry volume (i.e. the `G4HepEm` material-cut index of the given volume) can be done easily: 
    - the `G4HepEm` material-cut index can be added to each volume as an extra, user information
    - in order to have this index, one can get the `G4MaterialCutsCouple` object of the volume, obtain its index and get the corresponding `G4HepEm` material-cut index
    - so one line in a loop over the volumes 
 3. then the geometry can be written into a `.gdml` file (with one line of code) using standard Geant4 functionalities (see the `Geant4/geant4.10.07/examples/extended/persistency` examples in Geant4). 


**So according to my understanding, these developments would be better placed in the `G4HepEm` side, since:** 
  1. the data formats are defined there (and will be changed as part of the R&D).  So each of the data structures, defined in `G4HepEmData`, can have their  `Read/Write` (or whatever) function for writing into and reading from file. Then the top level `G4HepEmData` `Read/Write` method can be the responsible for: 
       - calling all these individual write-functions to write the corresponding data into a file 
       - reading the individual data structures by calling their read-function (that will construct the corresponding data structure and fill)

     Just as right now the `copy to device` and `fee on device` functions works.  I have not implemented this because there were long discussions on the data format (I think whatever format is suitable now and its easy to change the format later if the functionality is there). 

     `NOTE`: **this is the only way of guarantee that the data Write/Read always up-to-date regarding the data formats.** 

  2. the geometry related information can also be written from the same Geant4 application into a `.gdml` file (after the extra  information regarding the volume - G4HepEm-MatCat-index is set for each volume (but this is much simpler).

  3. then on the `AdePT` side (i.e. when using these data), one can easily:
      - create the top level `G4HepEm` data structure and call its top level `Read` method: this will make sure constructing and loading all data 
      - load the geometry from the `.gdml` file (including the extra information on the `G4HepEm` material-cut index)
      - then ready to perform the simulation      


Let me know if I misunderstand something.

drbenmorgan:
Hi @mnovak42, I wasn't aware you're thinking about persistency for `G4HepEm` data, hence doing it here! If it makes more sense to do this in `G4HepEm` that's fine and I can move things there - unless @agheata and @hahnjo have other thoughts.

mnovak42:
I just do not see other way to ensure consistency. For example, I have just added the gamma interaction to `G4HepEm` that comes with a new set of data structures and data. These would be immediately available to Write/Read, copy to device, etc. and use following my suggestions, while it would require a kind of "follow up" development on the AdePT side if the Write/Read functionality is decoupled form their definitions. The same happens when I change one or more of the data structures (that will happen frequently).

Also, here you had to repeat again the detector construction code and fake `Geant4` initialisation  in order to be able to start to work. All those are available on the `G4HepEm` side by default since `G4HepEm` lives only together with a given `Geant4` application. 

But of course, this and all above are just my opinion. You need to decide what is more suitable for `AdePT`.

drbenmorgan:
I agree on the consistency part! How about we move this "example" over as a new "test" in `G4HepEm`? That starts things off, and can use the testing process to develop and iterate the interfaces you mention.

agheata:
> I agree on the consistency part! How about we move this "example" over as a new "test" in `G4HepEm`? That starts things off, and can use the testing process to develop and iterate the interfaces you mention.

I think it is a good idea to have the feature in G4HepEm, at the end we're producing all the data there. To test this we can have an example in G4HepEm, but the feature should work with any setup. The example for TestEm3 should stay IMO on the AdePT side,  invoking this feature, because we will have to maintain (and maybe change often) the GPU equivalent.

mnovak42:
It's up to you @drbenmorgan. It can also stay here while only the `G4HepEm` data structure related `Read`/`Write` functionality is developed in the corresponding `G4HepEm` side (together with the related individual unit tests).   

@agheata, this has nothing to do with the `TestEm3` like `AdePT` example. 

@hahnjo provides (we will make it complete soon) a `TestEm3` like `AdePT` example (I think in `example6`) where the corresponding simplified sampling calorimeter detector is build from code, the corresponding `Geant4` and `G4HepEm` initialisations are also done in order to have the `G4HepEm` data structures in memory. Then the simulation can be started. 

This example helps us progressing with the simulation development and cross verification while the functionality, targeted by this WIP PR (writing/reading geometry and material, physics, etc. data), is under development. When this functionality will be available, the detector construction, the `Geant4` and `G4HepEm` initialisation parts can be dropped even from `example6` (geometry can then be loaded form `gdml` file with the auxiliary material-cut index information as well as all `G4HepEm` data) and everything will work just as fine as before.

hahnjo:
A bit late to the party, but I fully agree on the consistency part. It will be much easier to mirror changes to data structures (including things like padding for performance optimizations) when everything is in one repository.

For `TestEm3`, the current prototype (only with e-/e+) is in #94 (based on `example8` in #91, which in turn is derived from `example6`). I don't think it's useful to read "fixed" physics data and GDML files for that because the verification is heavily based on tweaking settings and checking values across different simulation runs. In that case, it's easier to have the Geant4 part in the same executable instead of first dumping a new set of physics tables.

drbenmorgan:
Thanks all! Let's do the following:

- I'll move this to G4HepEM for test/development of the persistency interfaces/format
- I'll update _this_ PR as a WIP to _use_ those interfaces from G4HepEM when ready

drbenmorgan:
This is now updated with the basics of export/import, with:

- `example7-geant4` creates a mock Geant4 geometry that should be identical to the default `TestEm3` setup
  - It persists the geometry and G4HepEmData to GDML and JSON files respectively (by default, `example7.{gdml,g4hepem}`)
- `example7-adept` reads in the two files and for now
  - Creates a VecGeom world volume from the input GDML file
  - Create a G4HepEMData instance from the input G4HepEM file
  
The build's updated and both should run as tests to check basic functionality. Next is to try and use the VecGeom/G4HepEmData as in TestEm3, but I need to understand how that links VecGeom volumes and material-cut couples.

drbenmorgan:
The latest commit implements:

- export of Geant4 Material Cut Couple index for a given G4LogicalVolume as an `<auxiliary>` tag for that volume
- export of the associated G4HepEmData JSON file name as an auxiliary tag under the GDML `<userinfo>` section
- import/read of the G4HepEmData file from the GDML
- import/reconnection of the G4HepEmData material cut couple index to the VecGeom Logical Volume Id

The end result at the moment is an array of ints as in TestEm3 which maps VecGeom LogicalVolume ids to the G4HepEm material cut couple index, i.e.

```
int mccIndex = MCCMap[lv->id()];
```

The build _will_ fail at the moment due to the requirement for the new VecGeom interfaces, but that's fine for now until we get a new release/deploy of VecGeom (and we may need to update other AdePT examples that read GDML to use the new interface).

drbenmorgan:
O.k., this is just about ready - it'll require VecGeom 1.1.14 or above to test, but the example is disabled for earlier versions so shouldn't break things. `example7-adept` is now a near copy of TestEm3, and values for global scoring are identical for 1 particle. Whilst I think it's in a sufficient state to demo setup from the persisted GDML/Data there are a few things to consider:

- The scoring-per-volume works, but the results aren't as in TestEm3 because the volume hierarchies are different
  -  TestEm3 has direct placement of each absorber, whilst example7 places one Pb/Ar absorber in each layer, the layer being placed 50 times. Scoring by PlacedVolume id therefore doesn't work, would need to get mother volume id/copy number.
- We should think about whether we should be looking up Volume-CutCouple indices by PlacedVolume id or LogicalVolume id. In the end the result is the same, so this is purely for whether we want/need consistency with how Geant4 treats this lookup.
- More for `G4HepEm`, but we should think about how best to copy/use this on the device. The copy to device and ownership semantics (see TestEm3.cu) are _awkward_. It's fine for this trivial case, but we should address things for the long term.

Anyway, ready for review!

examples/Example7/example7-adept.cc
```diff
@@ -0,0 +1,193 @@
...
+      return 5;
+    }
+
+    // Value is of auxiliary is Geant4 mcc index, so use G4HepEm to get index conversion
+    int geant4MCCIdx = std::stoi(mccAux.GetValue());
+    MCCIndex[pv->id()] = inData->fTheMatCutData->fG4MCIndexToHepEmMCIndex[geant4MCCIdx];
+  }
+
+  // Place particles between the world boundary and the calorimeter.
+  double startX = (-0.5 * fWorldSizeX) + 1.0*copcore::units::mm;
```
agheata, line 150: I was wondering if we could avoid sharing a common geometry header with constants by injecting auxiliary key-value pairs for the constants we need in the gdml file, which would become ideally the single source for the setup configuration in AdePT. We could have:
```
<auxiliary auxtype="Generator" auxvalue="gun">
  <auxiliary auxtyype="xgun" auxunit="mm" auxvalue="startX"/>  // well, what startX value is
  <auxiliary auxtyype="ygun" auxunit="mm" auxvalue="startY"/>
  ...
</auxiliary>
```
For other generators, the content could be e.g. the file name with hepmc3 particles, .... We can actually pass also other generic constants (for which the meaning is known apriori) via this mechanism if we need. This would allow having the AdePT app generic, dependent only on the gdml content. And if you want to run it with some different parameters (e.g. changing generator parameters), you could just modify the gdml. Since gdml also allows having parts written in different files, we could also keep the configuration part in a separate file included from the main gdml. I'm just trying to figure out a way to avoid writing a different app for each setup we want to test...

examples/Example7/example7-geometry.h
```diff
@@ -0,0 +1,31 @@
...
+constexpr int fNbOfAbsorbers        = 2;
+constexpr double GapThickness      = 2.3 * copcore::units::mm;
+constexpr double AbsorberThickness = 5.7 * copcore::units::mm;
+constexpr double fAbsorThickness[fNbOfAbsorbers+1] = {0.0, GapThickness, AbsorberThickness};
+
+constexpr double fLayerThickness = GapThickness + AbsorberThickness;
+constexpr double fCalorThickness = fNbOfLayers * fLayerThickness;
+
+constexpr double fWorldSizeX  = 1.2 * fCalorThickness;
+constexpr double fWorldSizeYZ = 1.2 * fCalorSizeYZ;
```
agheata, line 29: Maybe use `g` instead of `f` to prefix global variables. However, if we manage to avoid having to share this via a header file would not bind the app to the specific TestEm3 setup.

examples/Example7/example7-adept.cc
hahnjo: I don't think it makes sense to pass generator settings via the geometry description; we should really put our efforts in having the connection to Geant4 instead of artificially pumping the standalone demonstrators. If anything, we could move the geometry such that the generator is at `(0, 0, 0)`...

agheata (commented):
This looks really good! The only comment I have is that ideally, the AdePT application would be agnostic to the setup, reading everything linked to the gdml file prepared in the Geant4 counterpart application. I've made some comments in the code along this line. Great work!

hahnjo (changes requested):
> TestEm3 has direct placement of each absorber, whilst example7 places one Pb/Ar absorber in each layer, the layer being placed 50 times. Scoring by PlacedVolume id therefore doesn't work, would need to get mother volume id/copy number.

I don't think this will work: As you demonstrate, you can nest placements so you need copy numbers of the entire chain to figure out a volume ID. Besides, I seem to remember that the GDML format flattens the volume hierarchy and duplicates the PlacedVolumes?

examples/Example7/example7-adept.cc
agheata: Generator settings are usually input program parameters and they could stay this way. However, hardcoding numbers of layers, distances, and other geometry-specific parameters should not be needed, I mean I would avoid computing gun position based on such data. No, I don't think it is a good idea to move the geometry.

hahnjo: But they are *not* part of the geometry description.

agheata: auxiliary info does not necessarily need to be part of the geometry description, for example, region production cuts are user-defined properties that connect to the geometry. It could be done the other way around and have a setup file containing this info and linking to the gdml file. The point is that to have the application reusable, it should not contain in the compiled code recipes that are setup dependent. I realize however that this may be a long shot in what concerns for example scoring, which is very much setup-specific, so maybe we give it another thought after we merge this...

WitekPokorski: Anticipating the fact that we will have input particles from LHCb in a file (or passed from Geant4 through the fast simulation  hook) we should implement some simple 'primary particle' interface in AdePT which allows to inject particles from any external source. In other words, there should be a container of primary particles, that we populate the way we want and then AdePT does 'get next primary particle' from that container.

drbenmorgan: I've changed this so that the "startX" point is calculated from the world volume extent. Not perfect, but better than nothing for this example. @WitekPokorski's suggestion is the right long term solution overall.

drbenmorgan:
> > TestEm3 has direct placement of each absorber, whilst example7 places one Pb/Ar absorber in each layer, the layer being placed 50 times. Scoring by PlacedVolume id therefore doesn't work, would need to get mother volume id/copy number.
> 
> I don't think this will work: As you demonstrate, you can nest placements so you need copy numbers of the entire chain to figure out a volume ID. Besides, I seem to remember that the GDML format flattens the volume hierarchy and duplicates the PlacedVolumes?

At least in the GDML generated by the Geant4 side, the volume hierarchy is retained. A full scoring implementation would, I guess, need the same concepts as Geant4 so there's both step and geometry info available - the latter like Geant4's Touchable concept.

agheata:
> > TestEm3 has direct placement of each absorber, whilst example7 places one Pb/Ar absorber in each layer, the layer being placed 50 times. Scoring by PlacedVolume id therefore doesn't work, would need to get mother volume id/copy number.
> 
> I don't think this will work: As you demonstrate, you can nest placements so you need copy numbers of the entire chain to figure out a volume ID. Besides, I seem to remember that the GDML format flattens the volume hierarchy and duplicates the PlacedVolumes?

In general scoring cannot be done indeed per PlacedVolume because of replication, so the data structure is typically handled by the user, who has to be aware of the detector segmentation mapping to geometry, and can identify run-time the hit element by functionality like _MotherVolumeId(int level)_, which can query the volume id a number of levels up in the current path. Since we don't know yet how to handle user-defined data and scoring code in AdePT, I propose to leave this as is for the moment and merge.

examples/Example7/example7-adept.cc
```diff
@@ -0,0 +1,201 @@
...
+
+  // Connect up VecGeom _physical_ volume ids to G4HepEMData material cut couple indexes
+  // via Geant4 material cut couple index supplied as a volume auxiliary in the GDML
+  // Inputs: vecgeom::GeoManager, vgdml::Middleware, G4HepEmData
+  // Output: Array of ints. Array indices are VecGeom PV id numbers, value at that index is G4HepEmData MCC index
+  // Visting all PVs walks the tree, so we get multiple copies in the vector. Thus reduced to a
+  // set of pointers afterwards to uniq the list.
+  std::vector<vecgeom::VPlacedVolume*> tmpPlacedVolumes;
+  vecgeom::GeoManager::Instance().getAllPlacedVolumes(tmpPlacedVolumes);
+  std::set<vecgeom::VPlacedVolume*> allPlacedVolumes{tmpPlacedVolumes.begin(), tmpPlacedVolumes.end()};
```
hahnjo, line 127: `std::unordered_set` probably suffices, right? It doesn't make sense to sort pointers...

examples/Example7/CMakeLists.txt
```diff
@@ -0,0 +1,66 @@
+# SPDX-FileCopyrightText: 2020 CERN
```
hahnjo, line 1: 2021 (also in many other files)

examples/Example7/example7-geant4.cc
```diff
@@ -0,0 +1,77 @@
...
+  // - Should create geometry, regions and cuts
+  G4PVPlacement* world = geant4_mock();
+
+  // Construct the G4HepEmRunManager, which will fill the data structures
+  // on calls to Initialize
+  auto* runMgr    = new G4HepEmRunManager(true);
+  auto* rngEngine = new G4HepEmCLHEPRandomEngine(G4Random::getTheEngine());
+  runMgr->Initialize(rngEngine, 0);
+  runMgr->Initialize(rngEngine, 1);
+  runMgr->Initialize(rngEngine, 2);
```
hahnjo, line 39: Is there a particular reason to use `G4HepEmRunManager`? I discussed this with Mihaly initially, and AdePT should not depend on the `G4HepEm` library because we don't need their TLData and such. I'd prefer to initialize only the needed data structures directly, as we're doing in all other examples.

examples/Example7/geant4_mock.h
```diff
@@ -0,0 +1,11 @@
+// SPDX-FileCopyrightText: 2021 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef EXAMPLE6_GEANT4_MOCK_H
+#define EXAMPLE6_GEANT4_MOCK_H
```
hahnjo, line 5: `EXAMPLE7`

hahnjo (changes requested):
> Since we don't know yet how to handle user-defined data and scoring code in AdePT, I propose to leave this as is for the moment and merge.

Didn't we say in the very beginning that there will be no user-defined scoring code on the GPU? In that case, I don't think it makes sense to create an example that shows how it doesn't work. What's wrong with `TestEm3`'s way of creating the geometry without replication? That "just works"

examples/Example7/example7-geant4.cc
drbenmorgan: No other reason than simplicity (and following what's in the G4HepEM tests/examples) - agree though it would be better to have direct initialization, but that's an addition for G4HepEm.

drbenmorgan:
> > Since we don't know yet how to handle user-defined data and scoring code in AdePT, I propose to leave this as is for the moment and merge.
> 
> Didn't we say in the very beginning that there will be no user-defined scoring code on the GPU? In that case, I don't think it makes sense to create an example that shows how it doesn't work. What's wrong with `TestEm3`'s way of creating the geometry without replication? That "just works"

It still makes sense to think about the effect different geometry constructions may have on how scoring would need to be implemented. Examples can also be there, in the context of R&D, to show what doesn't/hasn't worked (or can't at the moment), and to provide a starting point for later resolution.

examples/Example7/example7-geant4.cc
hahnjo: No, it's already there and we're using that in AdePT, see https://github.com/apt-sim/AdePT/blob/0165af581767edcfc43afdcc4429369270de72ae/examples/TestEm3/TestEm3.cu#L51-L55

hahnjo: I don't see the point of replacing 4 calls that you can call selectively by 1 call, but I'll let Mihaly decide

agheata:
There seem to be un-resolved conversations blocking this merge. @drbenmorgan can you mark the ones you resolved already?

drbenmorgan:
@agheata I've marked everything I resolved, but it still seems to mark "changes requested" - I can't find anything, so guess it means I can't resolve overall (makes sense!).

examples/Example7/example7-geant4.cc
hahnjo: The code still seems to be there...

agheata:
@hahnjo still needs to approve on his side, then we can merge

examples/Example7/example7-geant4.cc
agheata: Indeed, @drbenmorgan you are not using here yet the interface introduced in mnovak42/g4hepem#47, can you update the code to merge this?

drbenmorgan: Apologies, the change will need #121 merged to pick up the new interface from mnovak42/g4hepem#47

hahnjo (approved):
Looks good from my side

PR #88, opened 2021-02-18
Create and use minimal interface target for AdePT

drbenmorgan:
Add a CMake `INTERFACE` target for AdePT with main locations of headers and dependency on CopCore as usage requirements. This allows all of the examples and tests to use `target_link_libraries` to pick up all the headers (and pave the way to simplify the repo layout in a separate PR), and these have been updated accordingly.

The AdePT target is not totally complete in that it doesn't declare its dependence on VecGeom (through `LoopNavigator.h` in the base, and others in magneticfield/tracking/physics) yet. That's deliberate as how to correctly declare the CUDA vs CPU only usage requirements needs some thought, but the target as implemented here is a good first step.

### DIFF
--- a/examples/Example5/CMakeLists.txt
+++ b/examples/Example5/CMakeLists.txt
@@ -7,4 +7,4 @@ if(NOT TARGET G4HepEm::g4HepEm)
 endif()
 
 add_executable(example5 example5.cu)
-target_link_libraries(example5 PRIVATE ${Geant4_LIBRARIES} g4HepEmData g4HepEmInit g4HepEmRun CopCore::CopCore)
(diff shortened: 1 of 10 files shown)


examples/Example5/CMakeLists.txt
```diff
@@ -1,14 +1,10 @@
 # SPDX-FileCopyrightText: 2020 CERN
 # SPDX-License-Identifier: Apache-2.0
 
-if(NOT TARGET g4HepEm)
+if(NOT TARGET G4HepEm::g4HepEm)
```
hahnjo, line 4: Is this needed? I think we should use the target consistently, either with or without scope (see below and `example6`).

drbenmorgan: Good point on the scope, I've changed the remainder of G4HepEm consumers to use the scoped target names.

hahnjo (approved):
I like this :+1: (not being a CMake expert)

agheata (approved):
Nice!

PR #91, opened 2021-02-23
Add example with simplistic data structures

hahnjo:
A `SlotManager` hands out monotonic slots (integers) as an index to a plain array. This is up to 40% faster than `example6`. In the future, fields could be split off into separate arrays if memory coalescing proves important.

The disadvantage is that this scheme does not allow to reclaim memory after a track was killed (which is a good thing if you want to see where the tracks died after transportation finished). However, better overall performance makes it feasible to just simulate the primaries in multiple chunks and still be faster.

---

Fold discrete interactions into single kernel

Without holes, it is totally fine for a single kernel to transport particles and produce secondaries. This seems consistently faster than the previous scheme with multiple kernels.

---

(Please don't squash on merge, it might be interesting in the future to have this example with two kernels in the history)


hahnjo:
I need to debug the existing examples first, all of them show different results and `example6` even crashes after #82. I'll rebase this PR after I'm done with that...

agheata (approved):
Looks good! Indeed, memory cannot be reclaimed, but the mechanism to reclaim it in BlockData is anyway too heavy in atomics. To check also in a future version a schema based on SparseArray allowing to compact the data from time to time. Anyway, this looks quite simple and efficient.

PR #94, opened 2021-02-26
Add TestEm3 similar to Geant4

hahnjo:
The code is based on example9 and has all processes for electrons, positrons, and gammas. Scoring includes the total energy deposit, the number of secondaries and their steps as well as hits. The app also sums the energy deposit and charged track length per volume.


hahnjo:
Seems to work as a baseline version, my comment about the magnetic field earlier today seems to be false alarm (though tests with higher number of particles are still pending). Please take a look!

agheata (approved):
Looks good, modulo the cross-check for the larger sample in magnetic field. I suppose there is an equivalent example TestEm3 in G4HepEm, is this also available? I saw that you map cut couples by placed volume id, and it is maybe better to do it by logical volume (by attaching a struct to the logical volume) - no change needed here, just to keep in mind when generalizing the hook.

hahnjo:
> I suppose there is an equivalent example TestEm3 in G4HepEm, is this also available?

Err yes? https://github.com/mnovak42/g4hepem/tree/master/apps/examples/TestEm3

> I saw that you map cut couples by placed volume id, and it is maybe better to do it by logical volume (by attaching a struct to the logical volume)

AFAICT logical volumes don't have a unique ID plus volumes may be placed in different regions with different cuts, leading to different material cuts couples.

agheata:
> AFAICT logical volumes don't have a unique ID plus volumes may be placed in different regions with different cuts, leading to different material cuts couples.

True, but the same is valid for placed volumes if the mother is placed in different regions. In practice, I don't recall using often different cuts for the same volume, but indeed we will need to use regions and access them during transportation. Side topic though.

WitekPokorski (commented):
Excellent, thanks!

PR #96, opened 2021-03-08
Fixes to deal with change of units

hahnjo:
This restores the output of the examples, except for rounding errors. Most of those are in `example4` and `example6` because the global energy deposit is a `float`, which doesn't have enough precision for the range of energies we need. In addition, `example4` doesn't produce two secondaries at the end, probably also because of rounding changes for the threshold energy.


agheata (approved):
Looks fine.

PR #99, opened 2021-03-11
Add example9 with gamma interactions

hahnjo:
Particles are stored per type to enable integration of gamma interactions (conversion, Compton scattering, and simple photoelectric absorption) and the code is split into multiple source files.


agheata (approved):
Looks good and this example has nice new features: different particle types stored in different containers and handled in different streams, synchronization managed by events, processes in separate files.

PR #100, opened 2021-03-18
WIP: Using SparseVector in Example9

agheata:
This replaces the tracks storage from plain array + SlotManager to the SparseVector, aiming at a version that is able to reclaim tracks storage space and allow other scheduling scenarios in the future. The performance is basically the same, taking care of constructing in place secondaries without having to seed them. This version prints also the sparsity of the track containers per particle type after each iteration.

Still work in progress, the current scheduling approach does not allow compacting the SparseVector on the fly because the slots in flight get invalidated.


hahnjo:
Very interesting, though I would vote to have this as separate example - especially when gaining the ability to compact.

The performance indeed looks very similar, and going back to my stashed changes I see what I did wrong: This PR still uses the queues AFAICT, whereas I was playing with `example6` that still used `BlockData`. So my changes spent most of the time skipping over dead particles at the beginning of the `SparseVector`, understandably being slower. However, the disadvantage of the queues (as you noted this morning) is that compaction doesn't work. Now the important question is: Is there a scheme that is both fast and offers the ability to compact memory?

(Another point from preliminary testing is that it seems to change results: Before the last line was
```
iter  214 -- tracks in flight:     0 energy deposition:    85.9723 number of secondaries: 60122 number of hits: 66254
```
and now it is (I removed the sparsity output to make `diff` work):
```
iter  211 -- tracks in flight:     0 energy deposition:    85.9762 number of secondaries: 60010 number of hits: 65812
```
Note that all values are slightly different, including the number of iterations...)

agheata:
> Very interesting, though I would vote to have this as separate example - especially when gaining the ability to compact.
> 
I agree.

> Now the important question is: Is there a scheme that is both fast and offers the ability to compact memory?
This is what I'm looking at, I think there are a couple of options to try out.

> 
> Note that all values are slightly different, including the number of iterations...)
Hmm, I haven't noticed that in the latest version, for sure it was giving the same result exactly at some earlier point. Of course, it has to be identical, I'll check where the divergence comes from.

agheata:
This will be implemented in a separate example.

PR #101, opened 2021-03-24
[R&D] Qualify calls to Box shapes

hahnjo:
For TestEm3, there are only Box shapes. Experiment with qualifying the calls to avoid virtual functions, so the compiler can do a better job. Improves performance by 10% on my GPU!

/cc @hageboeck

### DIFF
--- a/base/inc/AdePT/LoopNavigator.h
+++ b/base/inc/AdePT/LoopNavigator.h
@@ -15,6 +15,19 @@
 #include <VecGeom/base/Vector3D.h>
 #include <VecGeom/navigation/NavStateIndex.h>
 
+#define CAST_TO_BOX
+#define QUALIFY_CALLS
+
+#ifdef CAST_TO_BOX
+#include <VecGeom/volumes/Box.h>
+using Box_t = const vecgeom::SimpleBox;
+#ifdef QUALIFY_CALLS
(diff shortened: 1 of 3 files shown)


agheata:
> For TestEm3, there are only Box shapes. Experiment with qualifying the calls to avoid virtual functions, so the compiler can do a better job. Improves performance by 10% on my GPU!
> 
> /cc @hageboeck

Imagine what the same would do for TrackML or alike... More difficult to test there since there are 3 types of shapes

hageboeck:
> > For TestEm3, there are only Box shapes. Experiment with qualifying the calls to avoid virtual functions, so the compiler can do a better job. Improves performance by 10% on my GPU!
> > /cc @hageboeck
> 
> Imagine what the same would do for TrackML or alike... More difficult to test there since there are 3 types of shapes

I guess it would do 🚀, but it's clearly not the solution that we hardcode the shapes in a more general geometry. One could hack example9 to do it once and get a number, but I won't have time for this the near future.

hahnjo:
> I guess it would do :rocket:, but it's clearly not the solution that we hardcode the shapes in a more general geometry. One could hack example9 to do it once and get a number, but I won't have time for this the near future.

No, we need a general solution. I'm working on a prototype still using all shapes from VecGeom.

PR #102, opened 2021-03-25
Update g4hepem submodule to current master

drbenmorgan:
Just bumps the submodule commit to the latest upstream G4HepEm. The changes to that from the previous commit are:

- [PR 40](https://github.com/mnovak42/g4hepem/pull/40) The G4HepEmProcess type has been changed to (Forced) discrete
- [PR 43](https://github.com/mnovak42/g4hepem/pull/43) Implement serialization of G4HepEmData data to JSON

Other PRs are associated with initialization of G4HepEMData structures needed to implement PR 43.

Integrating this means we'll be able to export G4HepEMData to file and read it in elsewhere without requiring linking to Geant4 libraries. Example 7 here will provide the testbed for that.


hahnjo (approved):
Works for me.

> PR 40 The G4HepEmProcess type has been changed to (Forced) discrete

Note that this is not important for AdePT, only when running inside Geant4 with a magnetic field (as we did for the validation).

agheata (approved):
Works for me also. I noted while enabling BUILD_TESTING for the G4HepEm submodule to test this, CMake complains that TestEm3 target is duplicated since there is one in G4HepEm.

hahnjo:
> Works for me also. I noted while enabling BUILD_TESTING for the G4HepEm submodule to test this, CMake complains that TestEm3 target is duplicated since there is one in G4HepEm.

That's totally expected and I don't think we need to build the tests of G4HepEm inside AdePT, do we?

agheata:
> > Works for me also. I noted while enabling BUILD_TESTING for the G4HepEm submodule to test this, CMake complains that TestEm3 target is duplicated since there is one in G4HepEm.
> 
> That's totally expected and I don't think we need to build the tests of G4HepEm inside AdePT, do we?

I just wanted to run Ben's test from the AdePT build, it was just convenient. No, we don't want G4HepEm tests in AdePT by default, but if you occasionally want to build them it would be handy to compile.

PR #105, opened 2021-03-31
Add reflection and refraction in RaytraceBenchmark

antoniopetre:
Add reflection and refraction model for transparent objects. Add exponential attenuation for rays which traverse transparent layers. Structure storing optical properties (transparency, refraction index, object color) for every logical volume.

### DIFF
--- a/examples/Raytracer_Benchmark/CMakeLists.txt
+++ b/examples/Raytracer_Benchmark/CMakeLists.txt
@@ -30,12 +30,22 @@ list(APPEND ADEPT_CUDA_SRCS
 add_library(raytracercu ${ADEPT_CUDA_SRCS})
 target_include_directories(raytracercu PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
 )
-target_link_libraries(raytracercu AdePT CopCore::CopCore VecCore::VecCore VecGeom::vecgeom VecGeom::vecgeomcuda_static)
+
+target_link_libraries(raytracercu VecCore::VecCore VecGeom::vecgeom VecGeom::vecgeomcuda_static CopCore::CopCore)
+target_compile_options(raytracercu PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:--extended-lambda;-fmad=false>")
 set_target_properties(raytracercu PROPERTIES CUDA_SEPARABLE_COMPILATION ON)
 
 #exec RaytraceBenchmark.cpp
 add_executable(RaytraceBenchmark Raytracer.cpp RaytraceBenchmark.cpp)
-target_link_libraries(RaytraceBenchmark PRIVATE raytracercu AdePT CopCore::CopCore VecGeom::vecgeom VecGeom::vecgeomcuda VecGeom::vgdml)
+target_include_directories(RaytraceBenchmark PUBLIC
+  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
+  $<INSTALL_INTERFACE:base>
+)
+target_link_libraries(RaytraceBenchmark VecCore::VecCore VecGeom::vecgeom VecGeom::vecgeomcuda VecGeom::vgdml raytracercu CopCore::CopCore)
+target_compile_options(RaytraceBenchmark PRIVATE "$<$<AND:$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>,$<COMPILE_LANGUAGE:CUDA>>:-fmad=false>")
 set_target_properties(RaytraceBenchmark PROPERTIES CUDA_SEPARABLE_COMPILATION ON CUDA_RESOLVE_DEVICE_SYMBOLS ON)
 
--- a/examples/Raytracer_Benchmark/RaytraceBenchmark.cu
+++ b/examples/Raytracer_Benchmark/RaytraceBenchmark.cu
@@ -24,6 +24,7 @@
 #include <cstdio>
 #include <vector>
 
+
 __global__ void RenderTile(RaytracerData_t rtdata, int offset_x, int offset_y,
                            int tile_size_x, int tile_size_y, unsigned char *tile_in, unsigned char *tile_out, int generation)
 {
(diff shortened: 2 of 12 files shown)


examples/Raytracer_Benchmark/RaytraceBenchmark.cu
```diff
@@ -137,7 +137,42 @@ void RenderTiledImage(adept::BlockData<Ray_t> *rays, cuda::RaytracerData_t *rtda
   COPCORE_CUDA_CHECK(cudaGetLastError());
 }
 
-void initiliazeCudaWorld(cuda::RaytracerData_t *rtdata) {
+__global__ void getAllLogicalVolumes(vecgeom::VPlacedVolume *currentVolume, vecgeom::Vector<vecgeom::LogicalVolume *> *container)
```
agheata, line 140: This looks quite inefficient as a way to extract the logical volumes in a vector. You don't need a (recursive) kernel doing it, you can try to use: CudaManager::LookupLogical(LogicalVolume const *const host_ptr) that returns the device pointer for each logical volume on the host. So if you need device pointers you can have just a loop on host on the vector of logical volumes stored by GeoManager (```GeoManager::GetAllLogicalVolumes()```) and get the device pointer using CudaManager.

examples/Raytracer_Benchmark/RaytraceBenchmark.cu
```diff
@@ -153,11 +188,56 @@ void initiliazeCudaWorld(cuda::RaytracerData_t *rtdata) {
...
+void print_vector_cuda(cuda::RaytracerData_t *rtdata, int no_generations) {
+  for (int i = 0; i < no_generations; ++i)
+    print<<<1,1>>>(rtdata, i); 
 }
 
-int executePipelineGPU(const vecgeom::cxx::VPlacedVolume *world, int argc, char *argv[])
+// Check if there are rays in containers
+__global__ void check_containers(cuda::RaytracerData_t *rtdata, int no_generations, bool *value)
+{
+  for (int i = 0; i < no_generations; ++i) {
```
agheata, line 220: You should probably have: `*value = false;` to start with, not having to care if the user of the function initialized it or not.

examples/Raytracer_Benchmark/RaytraceBenchmark.cu
```diff
@@ -153,11 +188,56 @@ void initiliazeCudaWorld(cuda::RaytracerData_t *rtdata) {
...
+    auto x = &rtdata->sparse_rays[i];
+    if (x->size_used() > 0) {
+      *value = 1;
```
agheata, line 223:
```suggestion
      *value = true;
```

examples/Raytracer_Benchmark/RaytraceBenchmark.hpp
```diff
@@ -79,77 +123,124 @@ int runSimulation(const vecgeom::cxx::VPlacedVolume *world, int argc, char *argv
...
+  COPCORE_CUDA_CHECK(cudaDeviceSynchronize());
+
   if (backend == copcore::BackendType::CUDA && use_tiles) {
-    RenderTiledImage(rays, (RaytracerData_t *)rtdata, output_buffer, block_size);
+    RenderTiledImage((RaytracerData_t *)rtdata, output_buffer, 0, block_size);
   } else {
     Launcher_t renderKernel(stream);
-    renderKernel.Run(renderkernelFunc, rays->GetNused(), {0, 0}, rays, *rtdata, input_buffer, output_buffer);
-    renderKernel.WaitStream();
+    while (check_used<backend>(rtdata, no_generations)) {
```
agheata, line 201: This loop would need to be explained in the README

examples/Raytracer_Benchmark/Raytracer.cpp
```diff
@@ -89,137 +89,219 @@ void InitializeModel(vecgeom::VPlacedVolume const *world, RaytracerData_t &rtdat
...
+  auto nextvol = (Ray_t::VPlacedVolumePtr_t)ray.fNextState.Top();
+
+  // Get material structure for last and next volumes
+  auto medium_prop_last = (MyMediumProp *)lastvol->GetLogicalVolume()->GetBasketManagerPtr();
+  auto medium_prop_next = (MyMediumProp *)nextvol->GetLogicalVolume()->GetBasketManagerPtr();
+
+  bool exit_transparent        = medium_prop_last->material == kRTtransparent;
+  bool transparent_transparent = medium_prop_next->material == kRTtransparent &&
+                                 medium_prop_last->material == kRTtransparent &&
+                                 (medium_prop_next->refr_index * medium_prop_last->refr_index) > 1.;
```
agheata, line 199: Maybe the condition should be more strict here, you don't want to do refraction between 2 media having the same index, even if it is >1

agheata (changes requested):
Looks good, it needs some modifications here and there plus some more detailed README.

examples/Raytracer_Benchmark/CMakeLists.txt
```diff
@@ -30,14 +30,21 @@ list(APPEND ADEPT_CUDA_SRCS
 add_library(raytracercu ${ADEPT_CUDA_SRCS})
 target_include_directories(raytracercu PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>
+  $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
+  $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
 )
-target_link_libraries(raytracercu AdePT CopCore::CopCore VecCore::VecCore VecGeom::vecgeom VecGeom::vecgeomcuda_static)
-target_compile_options(raytracercu PRIVATE "$<$<AND:$<OR:$<CONFIG:Debug>,$<CONFIG:RelWithDebInfo>>,$<COMPILE_LANGUAGE:CUDA>>:-G;-src-in-ptx>")
+target_link_libraries(raytracercu VecCore::VecCore VecGeom::vecgeom VecGeom::vecgeomcuda_static CopCore::CopCore)
+target_compile_options(raytracercu PRIVATE "$<$<COMPILE_LANGUAGE:CUDA>:--extended-lambda;-fmad=false>")
```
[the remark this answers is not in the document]
agheata: Is it still the case after fixing the color addition order? Do you get different pictures if you disable this flag?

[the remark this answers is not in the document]
antoniopetre: Yes, I tried it and the pictures are different without this flag.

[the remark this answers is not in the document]
agheata: OK, to be put on the todo list for understanding why, but IMO should not block the PR

agheata (approved):

PR #106, opened 2021-04-01
Introduce a VolumeDispatcher to query volumes without virtual calls

hahnjo:
Currently only for Box, Trd, and Tube. Requires https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/797.

### DIFF
--- a/base/inc/AdePT/VolumeDispatcher.h
+++ b/base/inc/AdePT/VolumeDispatcher.h
@@ -3,7 +3,7 @@
 
 /**
  * @file VolumeDispatcher.h
- * @brief Dispatching queries to volumes.
+ * @brief Dispatching queries to volumes without virtual calls.
  */
 
 #ifndef VOLUME_DISPATCHER_H_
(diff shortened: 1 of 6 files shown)


hahnjo:
Preliminary performance results:
`example8`: before 0.4479s, after 0.3914s (- 12.6%)
`example9`: before 0.9286s, after 0.8214s (- 11.5%)
`TestEm3` (with PbWO4): before 11.5009s, after 10.5905s (- 7.9%)

base/inc/AdePT/VolumeDispatcher.h
```diff
@@ -0,0 +1,86 @@
...
+    type output;                                                                               \
+    const auto unplacedStruct = static_cast<const vecgeom::class *>(vol)->GetUnplacedStruct(); \
+    vecgeom::impl::func(*unplacedStruct, __VA_ARGS__, output);                                 \
+    return output;                                                                             \
+  }
+
+#define DISPATCH_TO_PLACED_VOLUME(vol, type, func, ...)                                                            \
+  switch (vol->GetType()) {                                                                                        \
+    DISPATCH_TO_PLACED_VOLUME_KIND(vol, kBox, PlacedBox, BoxImplementation, type, func, __VA_ARGS__);              \
+    DISPATCH_TO_PLACED_VOLUME_KIND(vol, kTrd, PlacedTrd, TrdImplementation<vecgeom::TrdTypes::UniversalTrd>, type, \
```
sawenzel, line 33: is there an additional benefit, in comparison to (let's say for the box case): 
```
static_cast<PlacedBox*>(vol)->PlacedBox::FUNC
```

In such version, one would not need to know/add the "Implementation" type since it's known internally.

hahnjo: Calling `PlacedBox::FUNC` doesn't work because `PlacedBox` is an abstract class and doesn't provide implementations for `Contains`, `DistanceToIn`, etc. - that only comes with `SpecializedBox` (and is the reason why I brought up the complex inheritance structure in the last meeting). Qualifying to `SimpleBox` works (see #101), but I think this leaks even more internals about VecGeom.

hahnjo:
Ugh, I just added the missing dispatch for all solids in `cms2018.gdml` and the `BooleanImplementation`s call "the virtual function interface of its constituents". So much to the layering.

agheata:
> Ugh, I just added the missing dispatch for all solids in `cms2018.gdml` and the `BooleanImplementation`s call "the virtual function interface of its constituents". So much to the layering.

Hmm, I forgot that booleans dispatch to components. We would need to replace the direct calls to the virtual interface with a call to a dispatcher, which could be set up via some macro mechanism. Or template the BooleanImplementation struct by a dispatch type, but this sounds awfully complicated.

hahnjo:
Updated numbers with the approach limited to Box, Trd, and Tube (so completely without virtual calls; I won't push the other shapes until we have a solution for the boolean volumes):
`TestEm3` (with `PbWO4` and 10,000 electrons of 10 GeV): before 33.2s, after 26.8s (-19.3%)
`example9` (default settings): before 0.88s, after 0.84s (-4.5%) - less than before because the RNG sequences changed, I think

Please note that I didn't optimize the magnetic field yet, so it's likely that the (relative) numbers for `example9` will grow bigger as the time spent for the other components is reduced.

PR #114, opened 2021-04-09
Branch RNG state when generating secondaries

hahnjo:
This is more correct because static skipping leads to overlapping sequences and more efficient because there is only one `Advance`.

### DIFF
--- a/base/inc/CopCore/include/CopCore/Ranluxpp.h
+++ b/base/inc/CopCore/include/CopCore/Ranluxpp.h
@@ -31,6 +31,23 @@ private:
   static constexpr const uint64_t *kA = kA_2048;
   static constexpr int kMaxPos        = 9 * 64;
 
+protected:
+  __host__ __device__
+  void SaveState(uint64_t *state) const
+  {
+    for (int i = 0; i < 9; i++) {
+      state[i] = fState[i];
+    }
+  }
+
+  __host__ __device__
+  void XORstate(const uint64_t *state)
+  {
+    for (int i = 0; i < 9; i++) {
+      fState[i] ^= state[i];
+    }
+  }
+
   /// Produce next block of random bits
   __host__ __device__
   void Advance()
@@ -144,6 +161,20 @@ public:
 
   __host__ __device__
   uint64_t IntRndm() { return this->NextRandomBits(); }
+
+  /// Branch a new RNG state, also advancing the current one.
+  __host__ __device__
+  RanluxppDouble Branch()
+  {
+    // Save the current state, will be used to branch a new RNG.
+    uint64_t oldState[9];
+    this->SaveState(oldState);
+    this->Advance();
+    // Copy and modify the new RNG state.
+    RanluxppDouble newRNG(*this);
+    newRNG.XORstate(oldState);
+    return newRNG;
+  }
 };
 
 #endif // COPCORE_RANLUXPP_H_
--- a/examples/Example9/electrons.cu
+++ b/examples/Example9/electrons.cu
@@ -114,10 +114,12 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
         sincos(phi, &sinPhi, &cosPhi);
 
         gamma1.InitAsSecondary(/*parent=*/currentTrack);
+        gamma1.rngState = currentTrack.rngState.Branch();
         gamma1.energy = copcore::units::kElectronMassC2;
         gamma1.dir.Set(sint * cosPhi, sint * sinPhi, cost);
 
         gamma2.InitAsSecondary(/*parent=*/currentTrack);
+        gamma2.rngState = currentTrack.rngState.Branch();
         gamma2.energy = copcore::units::kElectronMassC2;
         gamma2.dir    = -gamma1.dir;
       }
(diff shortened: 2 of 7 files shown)


hahnjo:
This provides a fantastic 2.8x improvement for `TestEm3`, reducing the time for 1000 electrons in `PbWO4` / `lAr` from 11.7 to 4.13 seconds (now at least as fast as a fully multi-threaded run with Geant4, if not slightly faster) :rocket:

At the same time, the results match the ones by Geant4 more closely: [fig_CaloEDep_NoField.pdf](https://github.com/apt-sim/AdePT/files/6285917/fig_CaloEDep_NoField.pdf)

base/inc/CopCore/include/CopCore/Ranluxpp.h
```diff
@@ -144,6 +161,21 @@ class RanluxppDouble : public RanluxppEngineImpl<52> {
...
+  RanluxppDouble Branch()
+  {
+    // Save the current state, will be used to branch a new RNG.
+    uint64_t oldState[9];
+    this->SaveState(oldState);
+    this->Advance();
+    // Copy and modify the new RNG state.
+    RanluxppDouble newRNG(*this);
+    newRNG.XORstate(oldState);
+    newRNG.Advance();
```
agheata, line 176: I guess xor-ing two consecutive states branches in a cheap and reproducible way, but why do you need to advance the newRNG?

amadio: Some RNGs need to be warmed up, not sure if this is the case for this particular RNG. However, I'd argue that one of the states could continue to be used without advancing as well, while the other could be branched and advanced.

hahnjo: > [...] but why do you need to advance the newRNG?

Actually I've asked myself the same question while having a walk. I added this because the whole approach is modeled after MIXMAX, but they use an LCG to mangle only a single component whereas I think it's better to XOR the whole state (after all, the states should already be independent).

> However, I'd argue that one of the states could continue to be used without advancing as well, while the other could be branched and advanced.

No, we need to advance the original state to make sure that the next branching doesn't end up in the same state. But there's a different "reuse" possible that I've already been working on in the afternoon, stay tuned :wink:

hahnjo: I ran some tests over the weekend and the errors in the tail worry me: [fig_CaloEDep_NoField.pdf](https://github.com/apt-sim/AdePT/files/6294438/fig_CaloEDep_NoField.pdf) This could mean that there's too much correlation between the states when only XORing. This is avoided by an immediate `Advance()` and the strong guarantees of RANLUX++. I'd prefer not to go this way, at least for now, until we understand if that's a real error or just statistical.

agheata: OK, we can keep the Advance if it reduces the correlations. Anyway, I think we need to understand those in particular for the case without field. One could think of rounding effects, is there anything in float in any of the three cases (G4 native, G4HepEm, and AdePT-G4HepEm? The last two seem to be free of float for the energy deposition, but can be due also to intermediate calculations. I would first check that.

hahnjo: Look, if I get correlations when deleting that line, how can it be `float`?

agheata: But there are correlations compared to G4 results even in the initial TestEm3 version, at least this is how I read the original plots. I'm certainly not excluding some are coming from the xor.

hahnjo: There are *errors* in the permille range, and there will be always be: even between versions of Geant4, you don't get exactly the same results. What worries me, is that they grow bigger.

jonapost:
The 'original method' was proposed Kostas Saviddis (MixMax author).  It was to generate a new state of the parent in one of a set of predefined ways (for MixMax by changing one or a few bits - but rotating the bit used by incrementing a 'daughter' counter. ) Then, to ensure that there is divergence between the parent and the daughter, both must be advanced a number of times which depends on the decorrelation time of the generator.  

Using a different method to 'mix' the state, such as x-oring a constant state with the current state, will no longer use those de-correlation properties of the generator.  How can we reason about it ?

I note that the original method does not provide the iron-clad guarantees provided by skipping ahead the original state of the generator to a selected location on the progression from the initial state (as part splitting the original sequence). That method could also be adapted for GPU, initially at least in a naive implementation that apportions the states using an atomic counter, to serve as a baseline 'correct' PRNG implementation (but in that case not reproducible.)

hahnjo:
> Using a different method to 'mix' the state, such as x-oring a constant state with the current state, will no longer use those de-correlation properties of the generator. How can we reason about it ?

I'm running TestU01's BigCrush on a kind of wrapper generator that interleaves random numbers from the original and the branched RNG state. So far, it has passed for the scheme `Advance + Copy + XOR + Advance` with one warning at `p = 5.3e-5` which is probably statistical. Right now, it's choking through `Advance + Copy + XOR` without the second `Advance`.

The underlying construction that inspired me is the so-called One-Time Pad, which is a guaranteed secure encryption scheme. As the purpose of an RNG is to generate non-related blocks of random bits, this should be a solid approach.

hahnjo:
Both methods (with and without the additional `Advance`) have passed the full BigCrush when either interleaving numbers from one branched RNG state with the original sequence and when branching a new state after using up all bits from the previous. So I've update the PR with the faster alternative, without the second `Advance`.

One correction, though:

> such as x-oring a constant state with the current state

The implementation XORs two dynamic states: the current one (of the parent) and the advanced one (the new state of the parent).

hahnjo:
For future reference and completeness: The "original" approach of toggling a single bit depending on a "branch counter" also passes full BigCrush, but requires additional storage (to count) and two `Advance`. The new approach for XORing requires no extra counter and only one `Advance` and is therefore faster.

Good to merge?

agheata (approved):

PR #115, opened 2021-04-09
Work around underestimated register count

hahnjo:
The compiler cannot see that we're going to call the functions through their pointers, so it underestimates the number of required registers. By including calls to the functions we force the compiler to account for the register usage.

### DIFF
--- a/examples/Example9/example9.cuh
+++ b/examples/Example9/example9.cuh
@@ -56,10 +56,18 @@ struct Track {
   }
 };
 
+// Defined in example9.cu
+extern __constant__ __device__ int Zero;
+
 class RanluxppDoubleEngine : public G4HepEmRandomEngine {
   // Wrapper functions to call into RanluxppDouble.
-  static __host__ __device__ double FlatWrapper(void *object) { return ((RanluxppDouble *)object)->Rndm(); }
-  static __host__ __device__ void FlatArrayWrapper(void *object, const int size, double *vect)
+  static __host__ __device__ __attribute__((noinline))
+  double FlatWrapper(void *object)
+  {
+    return ((RanluxppDouble *)object)->Rndm();
+  }
+  static __host__ __device__ __attribute__((noinline))
+  void FlatArrayWrapper(void *object, const int size, double *vect)
   {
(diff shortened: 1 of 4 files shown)


hahnjo:
This doesn't have an effect right now because `nvcc` currently has to reserve the maximum of 254 registers. This changes with #106 and I was getting funny "Illegal instruction" exceptions when testing that together with #114...

examples/Example9/example9.cuh
```diff
@@ -70,6 +78,17 @@ public:
...
+#ifdef __CUDA_ARCH__
+    // This is a hack: The compiler cannot see that we're going to call the
+    // functions through their pointers, so it underestimates the number of
+    // required registers. By including calls to the (non-inlinable) functions
+    // we force the compiler to account for the register usage, even if this
+    // particular set of calls are not executed at runtime.
+    if (Zero) {
+      FlatWrapper(engine);
+      FlatArrayWrapper(engine, 0, nullptr);
+    }
```
agheata, line 90: OK, if I understand correctly the call by pointer happens in G4HepEM I where the wrapper is passed and you need to show the compiler what it's calling into, and `RanluxppDouble::Rndm` uses quite a few registers. I wonder why it doesn't optimize out the if (Zero), but if it works...

agheata (approved):
How much speed-up this change only does, have you measured?

examples/Example9/example9.cuh
hahnjo: Yes, and not all functions from G4HepEm are inlined into `gammas.cu` yet (I need to make some changes there to make that possible).

The reason that the compiler cannot optimize out `if (Zero)` is that `Zero` is a global variable that could change its value. `__constant__` just means that its value doesn't change while a kernel is running, but there's an API function to change it from the host (which we use for setting up the data structures used by G4HepEm).

hahnjo (commented):
> How much speed-up this change only does, have you measured?

As I wrote, this change has no effect right now and is neutral in performance. It will result in higher register usage once that's not fixed to the maximum due to virtual calls, so it will degrade performance at some point. But it's better not to crash :smile:

PR #116, opened 2021-04-13
Fix sync in test_sparsevector

### LINKED ISSUE #109: Test of SparseVector fails when device not completely idle
Executing two instances of `test_sparsevector` reliably fails with `Error in compact.` as the last line.

agheata:
Added missing synchronization. Closes #109


hahnjo (approved):
Synchronization is generally a bit funky in that example (e.g. not directly after `print_*` kernels, but only later), but ok for now.

PR #117, opened 2021-04-13
Optimize RNG some more

hahnjo:
#### Reuse RNG state of killed particles

When the primary is killed, we can reuse the state for a secondary instead of branching twice.

---

#### Move RNG branching to reduce thread divergence

Creating a branched RNG state is faster before threads start diverging for the discrete interactions.

---

(please don't squash, the performance development might be of interest in the future)


agheata (approved):
What is the extra gain of this?

hahnjo:
> What is the extra gain of this?

Of course depends on the parameters: For `TestEm3` without magnetic field, the time for 10,000 electrons à 10 GeV is reduced from 36.5s to 33.5s. Simulating 100 electrons à 5 GeV with `example9` (including magnetic field) is basically flat.

PR #119, opened 2021-04-14
Update G4HepEm and interface calls

hahnjo:


WitekPokorski (approved):
Looks good, as discussed.

PR #123, opened 2021-06-01
Fix (some) ODR violations in CopCore

### LINKED ISSUE #122: RaytraceBenchmark segfaults in debug builds
Here is the output log:
```
$ cuda-gdb --args BuildProducts/bin/RaytraceBenchmark -on_gpu 1
NVIDIA (R) CUDA Debugger
11.3 release
Portions Copyright (C) 2007-2021 NVIDIA Corporation
GNU gdb (GDB) 8.3.1
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Type "show copying" and "show warranty" for details.
This GDB was configured as "x86_64-pc-linux-gnu".
Type "show configuration" for configuration details.
For bug reporting instructions, please see:
<http://www.gnu.org/software/gdb/bugs/>.
Find the GDB manual and other documentation resources online at:
    <http://www.gnu.org/software/gdb/documentation/>.

For help, type "help".
Type "apropos word" to search for commands related to "word"...
Reading symbols from BuildProducts/bin/RaytraceBenchmark...
(cuda-gdb) run
Starting program: /home/amadio/src/adept/build/BuildProducts/bin/RaytraceBenchmark -on_gpu 1
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib64/libthread_db.so.1".
INFO: using default trackML.gdml for option -gdml_name
INFO: using default 0 for option -cache_depth
(II) vgdml::Frontend::Load: VecGeom millimeter is 1
[Detaching after fork from child process 1468676]
[Detaching after fork from child process 1468683]
[New Thread 0x7fffdee06000 (LWP 1468684)]
[New Thread 0x7fffde605000 (LWP 1468685)]
[New Thread 0x7fffdde04000 (LWP 1468686)]
INFO: using default 1840 for option -px
INFO: using default 512 for option -py
INFO: using default 2 for option -model
INFO: using default 1 for option -view
INFO: using default 0 for option -reflection
INFO: using default 3.5 for option -zoom
INFO: using default -5000 for option -screenx
INFO: using default 0 for option -screeny
INFO: using default 0 for option -screenz
INFO: using default 0 for option -upx
INFO: using default 1 for option -upy
INFO: using default 0 for option -upz
INFO: using default -128 for option -bkgcol
INFO: using default 0 for option -use_tiles
INFO: using default 8 for option -block_size

Thread 1 "RaytraceBenchma" received signal SIGSEGV, Segmentation fault.
0x0000555555706e30 in adept::SparseVectorInterface<cuda::Ray_t>::SparseVectorInterface 
(this=0x7fff54000000, totalCapacity=4194304)
    at /home/amadio/src/adept/base/inc/AdePT/SparseVector.h:344
344	  __host__ __device__ SparseVectorInterface(size_t totalCapacity)
```
The configuration command I used was:
```
$ cmake  -DCMAKE_BUILD_TYPE="Debug" -DCMAKE_CUDA_ARCHITECTURES="75"
  -DCMAKE_CXX_STANDARD="17" -DCMAKE_PREFIX_PATH="/tmp/vecgeom" ~/src/adept
```

hahnjo:
Some function bodies are defined based on `COPCORE_CUDA_COMPILER`. When linking compiled objects from both `.cu` and `.cpp` files (such as the RaytraceBenchmark), and if the functions were not actually inlined (for example in a Debug) build, the linker got to pick one and the wrong implementation might be used.

This fixes #122 on my machine.

### DIFF
--- a/base/inc/CopCore/include/CopCore/Global.h
+++ b/base/inc/CopCore/include/CopCore/Global.h
@@ -27,9 +27,9 @@ enum BackendType { CPU = 0, CUDA, HIP };
 
 /** @brief CUDA error checking */
 #ifndef COPCORE_CUDA_COMPILER
-inline void error_check(int, const char *, int) {}
+static inline void error_check(int, const char *, int) {}
 #else
-inline void error_check(cudaError_t err, const char *file, int line)
+static inline void error_check(cudaError_t err, const char *file, int line)
 {
   if (err != cudaSuccess) {
     ::fprintf(stderr, "CUDA ERROR at %s[%d] : %s\n", file, line, cudaGetErrorString(err));
@@ -55,9 +55,9 @@ inline void error_check(cudaError_t err, const char *file, int line)
 
 /** @brief Check if pointer id device-resident */
 #ifndef COPCORE_CUDA_COMPILER
-inline bool is_device_pointer(void *ptr) { return false; }
+static inline bool is_device_pointer(void *ptr) { return false; }
 #else
-inline bool is_device_pointer(void *ptr)
+static inline bool is_device_pointer(void *ptr)
 {
   cudaPointerAttributes attr;
   cudaError_t err = cudaPointerGetAttributes(&attr, ptr);
@@ -68,9 +68,9 @@ inline bool is_device_pointer(void *ptr)
 
 /** @brief Get number of SMs on the current device */
 #ifndef COPCORE_CUDA_COMPILER
-inline int get_num_SMs() { return 0; }
+static inline int get_num_SMs() { return 0; }
 #else
-inline int get_num_SMs()
+static inline int get_num_SMs()
 {
   int deviceId, numSMs;
   COPCORE_CUDA_CHECK(cudaGetDevice(&deviceId));


agheata (approved):
I have noticed this behavior, but could not find the reason or a cure.

base/inc/CopCore/include/CopCore/Global.h
```diff
@@ -27,9 +27,9 @@ enum BackendType { CPU = 0, CUDA, HIP };
 
 /** @brief CUDA error checking */
 #ifndef COPCORE_CUDA_COMPILER
-inline void error_check(int, const char *, int) {}
+static inline void error_check(int, const char *, int) {}
```
amadio, line 30: Why not move these to a source file instead of keeping them a the header?
These are not performance critical functions, so I think moving them to a cpp/cu file is a better solution.

hahnjo: I wouldn't really mind them being in a source file, but that would have to be the CUDA version, which implies that all executables need to link with the CUDA libraries and there can never be a HIP implementation, say.

hahnjo:
@agheata 

> I have noticed this behavior, but could not find the reason or a cure.

Any other issues that we should know about before running into weird behavior like this?

base/inc/CopCore/include/CopCore/Global.h
amadio: In such cases, the build system can choose a source file with the appropriate implementations depending on the backend (i.e. either a .cpp or a .cu). This is a common way to avoid having to have `#ifdefs` in a lot of places. As long as the prototype doesn't change, it should work. For now we can just use `static` if this is a pain to move to a source file. The annoying thing is that you cannot tell CMake to consider a .cpp file as a CUDA source file, only .cu files, otherwise you could just have a .cpp with both implementations too.

PR #124, opened 2021-06-02
Adapt Raytrace_Benchmark example to use BVH acceleration structure

amadio:
This uses a simple adaptation of the `LoopNavigator.h` (changed into just `Navigator.h` in the benchmark's directory) to use the BVH acceleration structure on the GPU. This requires VecGeom's latest tag (1.1.16) in order to work.

Here are reported run times (Release build in both cases):

**Before:** CPU:  168.456 s, GPU: 61.872 s
**After**: CPU: 4.405 s, GPU: 1.063 s

**Speedups:** CPU: ≈38.24x, GPU:  ≈58.2x

The resulting rendered images are identical as far as I can tell, so the BVH navigator is working on the GPU as expected.

For your convenience, here is a diff of the headers`LoopNavigator.h` and `Navigator.h`:
```diff
--- base/inc/AdePT/LoopNavigator.h	2021-05-21 11:46:24.473844224 +0200
+++ examples/Raytracer_Benchmark/Navigator.h	2021-06-02 14:22:50.814901356 +0200
@@ -2,17 +2,18 @@
 // SPDX-License-Identifier: Apache-2.0
 
 /**
- * @file LoopNavigator.h
+ * @file Navigator.h
  * @brief Navigation methods for geometry.
  */
 
-#ifndef RT_LOOP_NAVIGATOR_H_
-#define RT_LOOP_NAVIGATOR_H_
+#ifndef RT_NAVIGATOR_H_
+#define RT_NAVIGATOR_H_
 
 #include <CopCore/Global.h>
 
 #include <VecGeom/base/Global.h>
 #include <VecGeom/base/Vector3D.h>
+#include <VecGeom/management/BVHManager.h>
 #include <VecGeom/navigation/NavStateIndex.h>
 
 #ifdef VECGEOM_ENABLE_CUDA
@@ -21,7 +22,7 @@
 
 inline namespace COPCORE_IMPL {
 
-class LoopNavigator {
+class Navigator {
 
 public:
   using VPlacedVolumePtr_t = vecgeom::VPlacedVolume const *;
@@ -36,26 +37,22 @@
       if (!vol->UnplacedContains(point)) return nullptr;
     }
 
-    VPlacedVolumePtr_t currentvolume = vol;
+    path.Push(vol);
+
     vecgeom::Vector3D<vecgeom::Precision> currentpoint(point);
-    path.Push(currentvolume);
+    vecgeom::Vector3D<vecgeom::Precision> daughterlocalpoint;
 
-    bool godeeper;
-    do {
-      godeeper = false;
-      for (auto *daughter : currentvolume->GetDaughters()) {
-        vecgeom::Vector3D<vecgeom::Precision> transformedpoint;
-        if (daughter->Contains(currentpoint, transformedpoint)) {
-          path.Push(daughter);
-          currentpoint  = transformedpoint;
-          currentvolume = daughter;
-          godeeper      = true;
-          break;
-        }
-      }
-    } while (godeeper);
+    for (auto v = vol; v->GetDaughters().size() > 0;) {
+      auto bvh = vecgeom::BVHManager::GetBVH(v->GetLogicalVolume()->id());
+
+      if (!bvh->LevelLocate(currentpoint, v, daughterlocalpoint))
+        break;
 
-    return currentvolume;
+      currentpoint = daughterlocalpoint;
+      path.Push(v);
+    }
+
+    return path.Top();
   }
 
   __host__ __device__
@@ -97,15 +94,9 @@
 
     if (step < 0) step = 0;
 
-    for (auto *daughter : pvol->GetDaughters()) {
-      double ddistance = daughter->DistanceToIn(localpoint, localdir, step);
-
-      // if distance is negative; we are inside that daughter and should relocate
-      // unless distance is minus infinity
-      const bool valid = (ddistance < step && !vecgeom::IsInf(ddistance)) &&
-                         !((ddistance <= 0.) && in_state.GetLastExited() == daughter);
-      hitcandidate = valid ? daughter : hitcandidate;
-      step         = valid ? ddistance : step;
+    if (pvol->GetDaughters().size() > 0) {
+      auto bvh = vecgeom::BVHManager::GetBVH(pvol->GetLogicalVolume()->id());
+      bvh->CheckDaughterIntersections(localpoint, localdir, step, pvol, hitcandidate);
     }
 
     // now we have the candidates and we prepare the out_state
```


agheata (approved):
Good to go for me. I would have used BVHNavigator or something similar as name, but it's minor detail.

amadio:
I will rename it and move to the same place as `LoopNavigator.h` then. I will see what I can do about the CI as well.

agheata:
> I will rename it and move to the same place as `LoopNavigator.h` then. I will see what I can do about the CI as well.

Don't forget to increase the VecGeom required version number in CMakeLists.txt

hahnjo:
I get similar results with `trackML.gdml`, but `cms2018.gdml` crashes in `vecgeom::cuda::BVHManager::GetBVH`. I'm not sure though if the BVH is to blame because it also crashes on `master` without this PR, but somewhere else (in `vecgeom::cuda::NavStateIndex::NavIndAddr` because you cannot call that when compiling with `nvcc` for the host?).

amadio:
Ok, I will try to reproduce the problem using cms2018. However, since the materials are hard-coded, how did you change the code to be able to use cms2018.gdml?

hahnjo:
> Ok, I will try to reproduce the problem using cms2018. However, since the materials are hard-coded, how did you change the code to be able to use cms2018.gdml?

Just passing `-gdml_name /path/to/cms2018.gdml`?

amadio:
I think that the geometry may not be properly setup, given all the hard-coded stuff, so I didn't even think about trying with CMS. I will see what happens in any case.

amadio:
I figured out what is happening, at least on the CPU. A call to the navigator is made before the initialization of the BVH structures...

amadio:
On the GPU the problem is that there is a call to the navigator on the CPU, but using `vecgeom::cuda::` namespace so it's the wrong copy of the CPU code that cannot/should not be called. This is *again* due to the double compilation on the host and namespacing issues. The initialization of the BVH structures happens for the host version of the code in the `vecgeom::cxx` namespace, so if you call the code in the `vecgeom::cuda` namespace (which happens automatically from a `.cu` file, as the inline namespace there is the `vecgeom::cuda` namespace), it will try to access uninitialized variables, that's why `GetBVH` crashes. The problem is that with VecGeom headers flip flopping which namespace is inline depending on how you compile things, this is only going to get worse and worse, as the `vecgeom::cxx` namespace is not accessible from `.cu` files and `vecgeom::cuda` namespace is not accessible from `.cpp` files of external projects trying to use VecGeom. Funnily enough, I cannot use the `CudaCheckError` function from VecGeom on a `.cu` file in AdePT for that reason! So we keep having to wrap calls to VecGeom in functions (like the `InitBVH` function I added to the RaytraceBenchmark example) and placing their body on a `.cpp` file if you need to call things from `vecgeom::cxx` and on `.cu` file if you need to call `vecgeom::cuda` stuff, so that, for example the `vecgeom::cxx::BVHManager::DeviceInit()` function can be called somewhere in a `.cu` file to initialize the GPU BVH structures if needed. This makes development using VecGeom on GPUs **_very very_** painful and wastes lots of time of developers that have to debug this sort of thing.

examples/Raytracer_Benchmark/Navigator.h
```diff
@@ -0,0 +1,245 @@
...
+
+  __host__ __device__
+  static VPlacedVolumePtr_t LocatePointIn(vecgeom::VPlacedVolume const *vol,
+                                          vecgeom::Vector3D<vecgeom::Precision> const &point,
+                                          vecgeom::NavStateIndex &path, bool top)
+  {
+    if (top) {
+      assert(vol != nullptr);
+      if (!vol->UnplacedContains(point)) return nullptr;
+    }
```
amadio, line 38: Why does it work on trackML.gdml and not cms2018.gdml? I guess the answer is that the origin point of the camera is such that for trackML the point is outside the world bounding box, while the CMS geometry is bigger, so the point lies inside the geometry and the first call to the navigator moves to the rest of the body, which then tries to access uninitialized stuff...

agheata: Probably. Have you tried to call the program with -screenx putting a larger negative value (now the default is -5000)? This should not crash of course. Can vecgeom::cxx be made explicit in the code at that place (since the call is supposed to be on the host geometry) or it just does not compile?

agheata:
> ... This makes development using VecGeom on GPUs **_very very_** painful and wastes lots of time of developers that have to debug this sort of thing.

You are right and we have seen how difficult it is to integrate a new structure/type in VecGeom and move it to the device. Also, this example is a demonstration of how complicated management of the namespaces becomes, in particular, if you want to have a single executable for host/device and common code.  I think we all agree, but we will have to still live with this schema for a while until we manage to refactor it

amadio:
The question that remains is: how does it even work for trackML? I think that the answer lies here:
```

```
So, after I fixed the problems discussed above, I now get this on the GPU:
```
RaytraceBenchmark: /tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h:148: static const NavIndex_t* vecgeom::cuda::NavStateIndex::NavIndAddr(NavIndex_t): Assertion `false && "reached unimplement code"' failed.
```
I guess this is due to the navigation index table not being deep enough to cover all of the geometry.

hahnjo:
> I guess this is due to the navigation index table not being deep enough to cover all of the geometry.

I *think* this may again be an instance of the host calling CUDA code, at least that's what the comment immediately above the `assert` suggests...

amadio:
Yep, I fixed my comment, thanks.

amadio:
As a last resort, I also tried to just move the camera point outside the geometry and use the code as is (from this PR). This is what I get:
```
...
=== Allocating 93.438 MB of ray data on the BackendType::CUDA

CUDA Exception: Lane User Stack Overflow

Thread 1 "RaytraceBenchma" received signal CUDA_EXCEPTION_2, Lane User Stack Overflow.
[Switching focus to CUDA kernel 3, grid 60079, block (6,0,0), thread (123,0,0), device 0, sm 12, warp 3, lane 27]
0x0000555559bc3fd0 in vecgeom::cuda::Transformation3D::Transform<-1, -1, double> (this=<optimized out>, master=..., local=...)
    at /tmp/vecgeom/include/VecGeom/base/Transformation3D.h:661
661	void Transformation3D::Transform(Vector3D<InputType> const &master, Vector3D<InputType> &local) const
```
So, even if we fix the problems with calling code from the wrong namespace, it will probably not be enough to get something working with the full CMS geometry.

agheata:
> So, even if we fix the problems with calling code from the wrong namespace, it will probably not be enough to get something working with the full CMS geometry.

Well, not necessary, but it's bad indeed that the code is stack-hungry. I've tried and succeeded in the past to increase the CUDA thread stack size, maybe we should try this first: https://gitlab.cern.ch/VecGeom/VecGeom/-/blob/35faafd3346978ea5d5bfd0fcfed1fd5c50828e8/source/benchmarking/RaytraceBenchmark.cu#L90-97

amadio:
I think we may also need to add an initialization mode for the CMS geometry later on. This is what I get when the code "works" on the CPU (with default settings):
output

examples/Raytracer_Benchmark/Navigator.h
amadio: Yes, when I call with a point far away, it gets past the point where it broke before. For CMS, I used `-screeny -20000`.

amadio:
I've fixed some of the problems, but the code will still fail on GPU for CMS because of calling `vecgeom::cuda` stuff on the host. The stack trace is this:
```
  screen_pos(-5000, 0, 0) screen_size(1840, 512)
  light_dir(0.57735, 0.57735, 0.57735) light_color(0xffffff80)
  zoom_factor(3.5) rt_model(2) rt_view(1)
RaytraceBenchmark: /tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h:148: static const NavIndex_t* vecgeom::cuda::NavStateIndex::NavIndAddr(NavIndex_t): Assertion `false && "reached unimplement code"' failed.

Thread 1 "RaytraceBenchma" received signal SIGABRT, Aborted.
__GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
49	  return ret;
(cuda-gdb) bt
#0  __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:49
#1  0x00007ffff16dc54a in __GI_abort () at abort.c:79
#2  0x00007ffff16dc42c in __assert_fail_base (fmt=0x7ffff184f018 "%s%s%s:%u: %s%sAssertion `%s' failed.\n%n", 
    assertion=0x55555584d3a8 "false && \"reached unimplement code\"", file=0x55555584d370 "/tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h", 
    line=148, function=<optimized out>) at assert.c:92
#3  0x00007ffff16eada3 in __GI___assert_fail (assertion=0x55555584d3a8 "false && \"reached unimplement code\"", 
    file=0x55555584d370 "/tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h", line=148, 
    function=0x55555584d320 "static const NavIndex_t* vecgeom::cuda::NavStateIndex::NavIndAddr(NavIndex_t)") at assert.c:101
#4  0x00005555556fa1dc in vecgeom::cuda::NavStateIndex::NavIndAddr (nav_ind=55111779) at /tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h:148
#5  0x00005555556fa23c in vecgeom::cuda::NavStateIndex::GetLevelImpl (nav_ind=55111779) at /tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h:196
#6  0x00005555556fa363 in vecgeom::cuda::NavStateIndex::GetLevel (this=0x7fffa80040e8) at /tmp/vecgeom/include/VecGeom/navigation/NavStateIndex.h:280
#7  0x00007ffff24dbeb6 in vecgeom::cuda::NavStateIndex::Print (this=0x7fffa80040e8) at /home/amadio/src/VecGeom/build/cuda_src/NavStateIndex.cu:85
#8  0x00005555556f7e58 in cuda::RaytracerData_t::Print (this=0x7fffa8004000)
    at /home/amadio/src/adept/build/examples/Raytracer_Benchmark/cuda_src/Raytracer.cu:42
#9  0x000055555570190b in runSimulation<(copcore::BackendType)1> (volume_container=0x7fffa8000000, world=0x7fffe644ce70, logicalvolumes=..., argc=5, 
    argv=0x7fffffffd648) at /home/amadio/src/adept/examples/Raytracer_Benchmark/RaytraceBenchmark.hpp:156
#10 0x00005555556fe9cb in executePipelineGPU (volume_container=0x7fffa8000000, world=0x7fffe644ce70, logicalvolumes=..., argc=5, argv=0x7fffffffd648)
    at /home/amadio/src/adept/examples/Raytracer_Benchmark/RaytraceBenchmark.cu:204
#11 0x000055555560c413 in main (argc=5, argv=0x7fffffffd648) at /home/amadio/src/adept/examples/Raytracer_Benchmark/RaytraceBenchmark.cpp:112
(cuda-gdb) 
```
I will leave fixing these remaining errors for someone else as I don't want to waste more time on debugging issues related to the double compilation on the host. We need to find a way to move away from this setup with urgency, and from also having to repeat it on any project using VecGeom (I tried to remove the `COPCORE_IMPL` namespace and get multiple definitions for symbols when I do that, or they end up calling the wrong parts of VecGeom unless compiled in a similar way by copying some cpp files as cu files and compiling twice).

amadio:
The CI is using LCG-style externals installed with LCGCMake, so I think waiting for Pere to update VecCore/VecGeom there next Monday is best, as I don't have an easy way to run the update myself.

agheata:
> I will leave fixing these remaining errors for someone else as I don't want to waste more time on debugging issues related to the double compilation on the host. 

Is the current branch containing all fixes you mentioned? How do you currently invoke the example using CMS geometry?

amadio:
No, I am still looking at a problem the latest change introduced. Please don't merge. To use with CMS, just use the -gdml_name option and pass the cms2018.gdml.

amadio:
For the record, now it works as long as the initial point is outside the geometry world (even with CMS geometry in release mode). Otherwise, `vecgeom::cuda` code gets called and things break. As I said before, I will leave this for someone else to fix, as it's an existing problem. In any case, I added a test that compares the output on CPU/GPU and fails if they are not identical, and will post the images to CDash if the failure is not a crash.

PR #125, opened 2021-06-07
Example10, reduce number of iterations

hahnjo:
There are two levels:
1. On the device, each transportation kernel can make several steps (up to a configurable limit, currently 4) as long as the particle stays alive and inside the volume. At volume boundary, relocation still happens in a separate kernel.
2. The host submits multiple kernels (currently 4 per particle type) into the appropriate streams and only synchronizes once at the end. Proper ordering on the device is ensured via events, extending the previous approach. The code uses an upper bound on the number of particles to determine the launch configuration.

In total, this reduces the number of iterations for `trackML` from 214 with `example9` to 29 using the code from the new example with identical results (energy deposition, number of secondaries, number of hit volume boundaries). However, best performance for this setup is obtained with only one step per kernel and submitting one kernel per iteration - in other words, exactly what example9 does.

TL;DR: This example is currently meant as a proof-of-concept and for playing with the parameters (number of steps and submitted kernels).

### DIFF
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -11,6 +11,7 @@ add_subdirectory(Example5)
 add_subdirectory(Example6)
 add_subdirectory(Example8)
 add_subdirectory(Example9)
+add_subdirectory(Example10)
 add_subdirectory(TestEm3)
 add_subdirectory(ECS)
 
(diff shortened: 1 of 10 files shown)


WitekPokorski (approved):
We merge it as it is for the future reference and eventual revisit.

PR #127, opened 2021-06-09
transport rays to the next boundary at every step in RaytraceBenchmark

antoniopetre:
The rays are transported only to the next boundary at every step. Two containers of indices are used. The rays used at every step are taken based on the indices from the first container. The second one stores temporary the indices of the alive rays. Before moving to the next step, the first and second containers are swapped and the second container is cleared.
Every ray has a new parameter, called ```reflection```. The new generated reflected rays have this parameter set (only for the current step). ```reflection``` is used for getting the indices of reflected rays.

### DIFF
--- a/examples/Raytracer_Benchmark/README.md
+++ b/examples/Raytracer_Benchmark/README.md
@@ -18,12 +18,9 @@ It currently works for only 2 gdml files: "trackML.gdml" (placed in the source t
 
 ## Ray transport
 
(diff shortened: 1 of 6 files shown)


agheata:
@antoniopetre how does the performance of the current version compares to the previous one? It would be nice also to see some profiling plots.

agheata (approved):

PR #129, opened 2021-06-11
Decorrelate branched RNG state

hahnjo:
It appears I never validated the final version of commit 02f73620f6 ("Branch RNG state when generating secondaries") after dropping the call to `Advance()` for the branched RNG state with the other commits to optimize RNG usage from PR #117 (commits 127b7be676 ("Reuse RNG state of killed particles") and 08e4301a98 ("Move RNG branching to reduce thread divergence")). Testing current master reveals errors in comparison to Geant4 in the order of multiple per mill. Restore the call to `Advance()` for better decorrelation of the branched RNG.

### DIFF
--- a/base/inc/CopCore/include/CopCore/Ranluxpp.h
+++ b/base/inc/CopCore/include/CopCore/Ranluxpp.h
@@ -173,6 +173,7 @@ public:
     // Copy and modify the new RNG state.
     RanluxppDouble newRNG(*this);
     newRNG.XORstate(oldState);
+    newRNG.Advance();
     return newRNG;
   }
 };


hahnjo:
Here's the comparison of 1 million electrons in TestEm3 (PbWO4 + lAr) between Geant4 and current `master`:  [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6637671/fig_CaloEDep.pdf) (**note** the increased scale for the error graph of 1 percent!).

With this fix, the result is much better: [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6637677/fig_CaloEDep.pdf)

agheata:
> Here's the comparison of 1 million electrons in TestEm3 (PbWO4 + lAr) between Geant4 and current `master`: [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6637671/fig_CaloEDep.pdf) (**note** the increased scale for the error graph of 1 percent!).
> 
> With this fix, the result is much better: [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6637677/fig_CaloEDep.pdf)

I'm just getting the same image, even if I rename locally one of the files, is it only me?

hahnjo:
Hm, I probably uploaded the same file twice, here's the figure from this PR: [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6638101/fig_CaloEDep.pdf)

agheata:
> Hm, I probably uploaded the same file twice, here's the figure from this PR: [fig_CaloEDep.pdf](https://github.com/apt-sim/AdePT/files/6638101/fig_CaloEDep.pdf)

👍

agheata (approved):

PR #131, opened 2021-07-15
Detect charged particles that are looping

hahnjo:
This improves performance by up to a factor of 2x if a significant fraction of the runtime is spent on waiting for these particles, until eventually running into the previous limit of 1000 iterations.

---

For testing with `cms2018.gdml`, be sure to have a VecGeom with https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/831 to avoid stuck particles due to `DistanceToOut` and `Contains` disagreeing for some polyhedra...

### DIFF
--- a/examples/Example11/example11.cu
+++ b/examples/Example11/example11.cu
@@ -243,7 +243,8 @@ void example11(const vecgeom::cxx::VPlacedVolume *world, int numParticles, doubl
   timer.Start();
 
   int inFlight;
-  int iterNo = 0;
+  int iterNo = 0, loopingNo = 0;
+  int previousElectrons = -1, previousPositrons = -1;
 
   do {
     Secondaries secondaries = {
@@ -320,8 +321,20 @@ void example11(const vecgeom::cxx::VPlacedVolume *world, int numParticles, doubl
               << " number of hits: " << std::setw(4) << stats->scoring.hits;
     std::cout << std::endl;
 
+    // Check if only charged particles are left that are looping.
+    numElectrons = stats->inFlight[ParticleType::Electron];
+    numPositrons = stats->inFlight[ParticleType::Positron];
+    numGammas = stats->inFlight[ParticleType::Gamma];
+    if (numElectrons == previousElectrons && numPositrons == previousPositrons && numGammas == 0) {
+      loopingNo++;
+    } else {
+      previousElectrons = numElectrons;
+      previousPositrons = numPositrons;
+      loopingNo = 0;
+    }
+
     iterNo++;
-  } while (inFlight > 0 && iterNo < 1000);
+  } while (inFlight > 0 && loopingNo < 20);
 
   auto time_cpu = timer.Stop();
   std::cout << "Run time: " << time_cpu << "\n";
--- a/examples/Example9/example9.cu
+++ b/examples/Example9/example9.cu
@@ -249,7 +249,8 @@ void example9(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double
   timer.Start();
 
   int inFlight;
-  int iterNo = 0;
+  int iterNo = 0, loopingNo = 0;
+  int previousElectrons = -1, previousPositrons = -1;
 
   do {
     Secondaries secondaries = {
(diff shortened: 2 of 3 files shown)


examples/Example11/example11.cu
```diff
@@ -320,8 +321,20 @@ void example11(const vecgeom::cxx::VPlacedVolume *world, int numParticles, doubl
               << " number of hits: " << std::setw(4) << stats->scoring.hits;
     std::cout << std::endl;
 
+    // Check if only charged particles are left that are looping.
+    numElectrons = stats->inFlight[ParticleType::Electron];
+    numPositrons = stats->inFlight[ParticleType::Positron];
+    numGammas = stats->inFlight[ParticleType::Gamma];
+    if (numElectrons == previousElectrons && numPositrons == previousPositrons && numGammas == 0) {
```
agheata, line 328: don't we have to check also BzFieldValue>0 here?

examples/Example11/example11.cu
```diff
@@ -320,8 +321,20 @@ void example11(const vecgeom::cxx::VPlacedVolume *world, int numParticles, doubl
...
+      loopingNo++;
+    } else {
+      previousElectrons = numElectrons;
+      previousPositrons = numPositrons;
+      loopingNo = 0;
+    }
+
     iterNo++;
-  } while (inFlight > 0 && iterNo < 1000);
+  } while (inFlight > 0 && loopingNo < 20);
```
agheata, line 337: This is mostly fine because there are no secondaries produced, but we should deposit the energy of the loopers. A way is to setup a kill signal visible in the kernels, then doe the killing there.

agheata: And we maybe we should account for the "killed" energy to tune the 20 value

examples/Example11/example11.cu
hahnjo: We could, but we also need a cut-off mechanism in case there is no field. Actually, it might even be necessary to kill stuck photons (due to geometry errors)

examples/Example11/example11.cu
hahnjo: TestEm3 has this killing, but I plan to remove it because Geant4 doesn't deposit energy of killed particles either.

> And we maybe we should account for the "killed" energy to tune the 20 value

First, getting the sum of "killed" energy requires a reduction, which is expensive on GPUs. But what should we do with the value? Is 1 MeV ok or not? And what if the particles continue to loop while they are above a threshold?

examples/Example11/example11.cu
agheata: Exactly, but this type of check is typically done based on non-advancing, by setting a counter per track incremented whenever geometry returns a too small step.

examples/Example11/example11.cu
agheata: Well, it is all about making sure the errors we do by this procedure are acceptable and what category of loopers we kill. I mean, killing slow curly electrons may be OK, but killing particles having geometry errors may require a different action. Just saying it's good to understand what we kill, depending on that depositing the energy in care there is a sensitive volume may or may not be needed. Did you have a look on the kinematic properties for the loopers?

examples/Example11/example11.cu
hahnjo: I'm against having a counter per track: It increases memory usage and checking per track, in every single step, is slower than one global check on the host (that already has the data)

examples/Example11/example11.cu
hahnjo: I have looked at the deposited energy, and that remains exactly the same as before. Plus Geant4 doesn't deposit energy of looping particles either.

examples/Example11/example11.cu
agheata: True, but how do we detect stuck tracks, you still have to do some accounting, no? And now we don't have a check for that. This does not concern this PR though.

examples/Example11/example11.cu
agheata: Then we are fine for now, but there may be cases in future where this simple approach can introduce non-negligible bias, and we need to keep in mind than the number 20 used now may not be ideal in other cases.

agheata (approved):

examples/Example11/example11.cu
hahnjo: The same way this PR already works, even though we should probably allow more steps for gammas because they can go much further without interaction.

PR #132, opened 2021-07-19
Only push when relocating a track

hahnjo:
The previous code pushed on every step which introduces systematic errors. Instead only push when reaching a boundary, also removing the problematic case that a slightly pushed point is relocated and the track itself is stuck at one position.

Also correct the handling of negative values from `DistanceToIn` in `LoopNavigator`.

---

I tested all examples with the `cms2018` geometry and I think the pushing when at boundary resolves all stuck particles. For validation, I ran with 1 million electrons on TestEm3 and didn't see much of a difference, presumably because the geometry is very simple.

### DIFF
--- a/base/inc/AdePT/BVHNavigator.h
+++ b/base/inc/AdePT/BVHNavigator.h
@@ -211,17 +211,18 @@ public:
   // Relocate a state that was returned from ComputeStepAndNextVolume: It first
   // removes all volumes from the state that were left, and then recursively
   // locates the pushed point in the containing volume.
-  __host__ __device__ static void RelocateToNextVolume(vecgeom::Vector3D<vecgeom::Precision> const &globalpoint,
+  __host__ __device__ static void RelocateToNextVolume(vecgeom::Vector3D<vecgeom::Precision> &globalpoint,
                                                        vecgeom::Vector3D<vecgeom::Precision> const &globaldir,
                                                        vecgeom::NavStateIndex &state)
   {
     // Push the point inside the next volume.
-    vecgeom::Vector3D<vecgeom::Precision> pushed = globalpoint + 1.E-6 * globaldir;
+    static constexpr double kPush = 1.e-8;
+    globalpoint += kPush * globaldir;
 
     // Calculate local point from global point.
     vecgeom::Transformation3D m;
     state.TopMatrix(m);
-    vecgeom::Vector3D<vecgeom::Precision> localpoint = m.Transform(pushed);
+    vecgeom::Vector3D<vecgeom::Precision> localpoint = m.Transform(globalpoint);
 
     VPlacedVolumePtr_t pvol = state.Top();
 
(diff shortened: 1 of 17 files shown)


base/inc/AdePT/BVHNavigator.h
```diff
@@ -211,17 +211,18 @@ class BVHNavigator {
...
   // removes all volumes from the state that were left, and then recursively
   // locates the pushed point in the containing volume.
-  __host__ __device__ static void RelocateToNextVolume(vecgeom::Vector3D<vecgeom::Precision> const &globalpoint,
+  __host__ __device__ static void RelocateToNextVolume(vecgeom::Vector3D<vecgeom::Precision> &globalpoint,
                                                        vecgeom::Vector3D<vecgeom::Precision> const &globaldir,
                                                        vecgeom::NavStateIndex &state)
   {
     // Push the point inside the next volume.
-    vecgeom::Vector3D<vecgeom::Precision> pushed = globalpoint + 1.E-6 * globaldir;
+    static constexpr double kPush = 1.e-8;
```
agheata, line 219: This should read something like 10*vecgeom::kTolerance to make it vecgeom::Precision proof.

base/inc/AdePT/BVHNavigator.h
```diff
@@ -211,17 +211,18 @@ class BVHNavigator {
...
+    globalpoint += kPush * globaldir;
```
agheata, line 220: Changing the globalpoint still does a systematic push, even if only when crossing boundaries. The bias may be negligible in most cases, but isn't it enough to use just the locally pushed point for relocation, and preserve the original globalpoint? I know this may create trouble on the next ComputeStep query since the globalpoint will be on boundary and numerically converted on the wrong side, but we could apply kPush just inside the next ComputeStep and then correct the distance before returning it by adding kPush. This would hide the push ensuring that the start point for querying the geometry is passed the boundary, but in case physics wins will be free of the kPush bias on boundary.

agheata: So the idea is to keep the tracking point numerically on boundary, but inside both Relocate and ComputeStep work with a locally pushed value to get around the rounding errors.

base/inc/AdePT/BVHNavigator.h
hahnjo: Maybe, but for now this PR is only moving the relevant code and it wasn't there before, so I'd prefer not to do this in the current PR.

base/inc/AdePT/BVHNavigator.h
hahnjo: I don't think it would be easy because you'd only want to do it after you moved to a new volume, which the next `ComputeStep` doesn't know about. This gets even trickier for the magnetic field propagator, which needs to query the navigator multiple times. Overall, this PR definitely reduces the bias so I think there is merit, for the rest I'm happy to be proven wrong if you can make it work.

agheata (approved):

PR #135, opened 2021-07-23
Add a simple NVTX helper class for profiling

hageboeck:
### Add NVTX tracer to aid profiling.
Include a tracer to annotate profiling, e.g.:
```diff
+  NVTXTracer tracer("SimulationLoop");
```

The class is a bit rudimentary and can/should be extended, but it at least supports cycling through a few colours and setting tags for the different phases of the workflow.

### Include NVTX annotations for example9:
When compiling with `-DUSE_NVTX`, example 9 looks as follows in nsight systems:
image

### DIFF
--- a/base/inc/AdePT/NVTX.h
+++ b/base/inc/AdePT/NVTX.h
@@ -1,0 +1,75 @@
+// SPDX-FileCopyrightText: 2021 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @file NVTX.h
+ * @brief Helper for annotating workflows with NVTX profiling information.
+ */
+
+#ifndef NVTX_H
+#define NVTX_H
+
+#if defined USE_NVTX && !defined __CUDACC__
+
+#include "nvToolsExt.h"
+
+#include <array>
+#include <cstdint>
+#include <string>
+
+class NVTXTracer {
+  static constexpr std::array<uint32_t, 7> _colours = {0xff00ff00, 0xff0000ff, 0xffffff00, 0xffff00ff,
+                                                       0xff00ffff, 0xffff0000, 0xffffffff};
+  std::string _name;
+  nvtxRangeId_t _id;
+
+public:
+  NVTXTracer(const char *name) : _name(name)
+  {
+    nvtxEventAttributes_t eventAttrib = {0};
+    eventAttrib.version               = NVTX_VERSION;
+    eventAttrib.size                  = NVTX_EVENT_ATTRIB_STRUCT_SIZE;
+    eventAttrib.colorType             = NVTX_COLOR_ARGB;
+    eventAttrib.color                 = nextColour();
+    eventAttrib.messageType           = NVTX_MESSAGE_TYPE_ASCII;
+    eventAttrib.message.ascii         = name;
+    _id                               = nvtxRangeStartEx(&eventAttrib);
+  }
+  ~NVTXTracer() { nvtxRangeEnd(_id); }
+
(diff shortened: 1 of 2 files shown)


agheata (approved):
This is cool!

hageboeck:
CI? How do we make it run? It *should* work, since all is `#ifdef`-ed out, but let's check.

hageboeck:
@phsft-bot build

examples/Example9/example9.cu
```diff
@@ -252,13 +264,24 @@ void example9(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double
...
     };
 
+    const auto nInFlight = std::accumulate(std::begin(stats->inFlight), std::end(stats->inFlight), 0);
+    if (nInFlight > 3400) {
+      occupTracer.setTag("PeakOccup");
+    } else if (nInFlight > 1000) {
+      occupTracer.setTag("HighOccup");
+    } else {
+      occupTracer.setTag("LowOccup");
+    }
```
hahnjo, line 283: Just use `inFlight` :smiley:

base/inc/AdePT/NVTX.h
```diff
@@ -0,0 +1,77 @@
...
+        eventAttrib.size = NVTX_EVENT_ATTRIB_STRUCT_SIZE;
+        eventAttrib.colorType = NVTX_COLOR_ARGB;
+        eventAttrib.color = nextColour();
+        eventAttrib.messageType = NVTX_MESSAGE_TYPE_ASCII;
+        eventAttrib.message.ascii = name;
+        _id = nvtxRangeStartEx(&eventAttrib);
+    }
+    ~NVTXTracer() {
+        nvtxRangeEnd(_id);
+    }
```
hahnjo, line 40: Should probably run `clang-format` over this file...

agheata:
> CI? How do we make it run? It _should_ work, since all is `#ifdef`-ed out, but let's check.

What is the pipe of commands to get the profile picture, once some example is instrumented?

hageboeck:
> > CI? How do we make it run? It _should_ work, since all is `#ifdef`-ed out, but let's check.
> 
> What is the pipe of commands to get the profile picture, once some example is instrumented?

This, modulo typos:
```bash
cmake -DCMAKE_CXX_FLAGS=-DUSE_NVTX .
#build
nsys profile -o /tmp/output <executable> <args>
nsys-ui /tmp/output.qdrep
```

agheata:
> This, modulo typos:

We should be able to run these, but what is the output, an image? And how do we validate/publish?

examples/Example9/example9.cu
hageboeck: I didn't see it at this point. It's indeed there, but uninitialised on the first run. I'll set it to `numParticles`, and then it should be there for the following runs.

hageboeck:
> We should be able to run these, but what is the output, an image? And how do we validate/publish?

I think there's a misunderstanding:
- I was asking for a CI that checks the code when NVTX is not active, so we don't destroy existing workflows.
- When there's an mistake with NVTX active, the person who wants to do the profiling will fix it, but it's only one person that's affected.

hahnjo (approved):

agheata:
> * I ~was asking~ meant to ask 🙂  for a CI that checks the code when NVTX is not active, so we don't destroy existing workflows.

OK I see. The way you made the header protects for this, so it should be fine.

hageboeck:
> OK I see. The way you made the header protects for this, so it should be fine.

Where's jenkins?

PR #136, opened 2021-07-23
Store only one NavStateIndex

hahnjo:
This saves 8 bytes per track and a bunch of global memory accesses. Ideally, we would only save the `NavIndex_t` but there is no API to directly retrieve the member `fNavInd` (`GetNavIndex` expects a level and both getting and using that requires walking the hierarchy).

### DIFF
--- a/examples/Example11/electrons.cu
+++ b/examples/Example11/electrons.cu
@@ -72,11 +72,12 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
     // also need to carry them over!
 
     // Check if there's a volume boundary in between.
+    vecgeom::NavStateIndex nextState;
     double geometryStepLength = fieldPropagatorBz.ComputeStepAndPropagatedState</*Relocate=*/false, BVHNavigator>(
         currentTrack.energy, Mass, Charge, geometricalStepLengthFromPhysics, currentTrack.pos, currentTrack.dir,
-        currentTrack.currentState, currentTrack.nextState);
+        currentTrack.navState, nextState);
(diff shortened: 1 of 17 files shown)


agheata (approved):
This makes sense with NavStateIndex because it can be handled on the stack, while this was not possible using the previous NavStatePath. Indeed, we can make NavStateIndex constructible from a NavIndex_t and boundary flag since we manage the boundary flag in the track, then compose the NavStateIndex upon usage via the navigation interfaces, which will reduce the geometry state even more. To add a getter as well.

hahnjo:
> This makes sense with NavStateIndex because it can be handled on the stack, while this was not possible using the previous NavStatePath.

If `NavStatePath` cannot be on the stack, I would guess it involves dynamic memory allocation which doesn't work in `Track` anyway.

> Indeed, we can make NavStateIndex constructible from a NavIndex_t and boundary flag since we manage the boundary flag in the track, then compose the NavStateIndex upon usage via the navigation interfaces, which will reduce the geometry state even more. To add a getter as well.

`NavStateIndex` is constructible from a `NavIndex_t`, but (as written in the commit message) there is no efficient way to get the index of an existing state. And we don't need to store the boundary flag, it is only used temporarily until a track is relocated.

agheata:
The getter is of course trivial to add. The boundary flag will be needed if we have to expose the state to user code for scoring, even if we don't use it now.

agheata:
See: https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/833

PR #137, opened 2021-08-05
WIP: Made AdePT compile in single precision mode. The box_sphere geometry can be ray-traced.

agheata:
From M. Kostelnik and A. Gheata:
This can only be merged after tagging the master of VecGeom. It introduces type changes for the geometry state to use vecgeom::Precision rather than double, to allow compiling with VecGeom in SINGLE_PRECISION mode. In particular, the track state and magnetic field propagation were the most affected.

The patch modifies the ray-tracing example to add optional approaching the bounding box of daughters before computing the distance to daughters, to reduce errors for the propagated points on boundaries. One needs to use `-approach 1` as input to RaytraceBenchmark. This flag seems to have no impact however on the image produced in single precision for `box_sphere.gdml` geometry. Raytracing trackML.gdml currently hangs.


base/inc/AdePT/SparseVector.h
```diff
@@ -718,12 +718,14 @@ class SparseVector : public SparseVectorInterface<Type> {
...
   {
     if (!addr) return new SparseVector(N);
-    bool devicePtr = copcore::is_device_pointer(addr);
-    if (devicePtr) return SparseVector::ConstructOnDevice(addr);
-    return new (addr) SparseVector(N);
+    if (backend == copcore::BackendType::CUDA)
+      return SparseVector::ConstructOnDevice(addr);
+    else
+      return new (addr) SparseVector(N);
   }
```
hahnjo, line 729: This looks like a different change, can you submit this as a separate PR? Also, can we get rid of the broken `is_device_pointer` completely?

agheata: Yes, I forgot to push this separately, the idea is to get rid of this function.

agheata: Moved to #138

hahnjo:
The first commit contains the changes removed during review of #138; they should not be a separate commit

agheata:
> The first commit contains the changes removed during review of #138; they should not be a separate commit

Indeed, fixed now. Running ctest in single precision mode I noticed that the changes from double to float make some examples (Exa9) hang on few particles, so I'll need to do some extra tuning in this case.

agheata:
The LoopNavigator + Exa9 work now in single-precision mode, giving ~2.8x speedup compared to double precision with compatible output.

agheata:
I have switched the VecGeom requirement tpo 1.1.17. The test show failures only on the Raytrace comparison tets between CPU and GPU, few pixels get different color values. Debugging this, it turned out that the Release version of VecGeom with CUDA gives in some cases different values for distances compared to its Debug version. This does not happen for the CPU code, so I suspect that some optimizations enabled in the GPU compilation of VecGeom are maybe too aggressive. This is a speculation however and needs to be debugged more thoroughly.

examples/Raytracer_Benchmark/Raytracer.cpp
```diff
@@ -137,28 +139,26 @@ adept::Color_t RaytraceOne(RaytracerData_t const &rtdata, Ray_t &ray, int genera
...
     while (nextvol == ray.fVolume && nsmall < kMaxTries) {
-      snext   = BVHNavigator::ComputeStepAndPropagatedState(ray.fPos, ray.fDir, vecgeom::kInfLength, ray.fCrtState,
-                                                           ray.fNextState);
+      vecgeom::Precision dist = 0.;
+      if (rtdata.fApproach) {
+        dist = BVHNavigator::ComputeStepToApproachNextVolume(ray.fPos, ray.fDir, vecgeom::kInfLength, ray.fCrtState);
+        ray.fPos += dist * ray.fDir;
+      }
+      snext = BVHNavigator::ComputeStepAndPropagatedState(ray.fPos, ray.fDir, vecgeom::kInfLength, ray.fCrtState,
+                                                          ray.fNextState, kPush);
```
hahnjo, line 153: We should not push unconditionally for every step. If it's needed for single precision, we should do it only then.

base/inc/AdePT/BVHNavigator.h
```diff
@@ -197,7 +197,7 @@ class BVHNavigator {
       }
 
       if (out_state.Top() != nullptr) {
-        while (out_state.Top()->IsAssembly()) {
+        while (out_state.Top()->IsAssembly() || out_state.GetNavIndex() == in_state.GetNavIndex()) {
```
hahnjo, line 200: What's the reason for this? Isn't this just hiding other problems?

base/inc/AdePT/LoopNavigator.h
```diff
@@ -128,6 +128,10 @@ class LoopNavigator {
     // Otherwise it is a geometry step and we push the point to the boundary.
     out_state.SetBoundaryState(true);
 
+    if (step < 0.) {
+      step = 0.;
+    }
+
```
hahnjo, line 134: How can this ever happen? `valid` above checks `ddistance >= 0`

base/inc/AdePT/LoopNavigator.h
```diff
@@ -166,7 +174,7 @@ class LoopNavigator {
       }
 
       if (out_state.Top() != nullptr) {
-        while (out_state.Top()->IsAssembly()) {
+        while (out_state.Top()->IsAssembly() || out_state.GetNavIndex() == in_state.GetNavIndex()) {
```
hahnjo, line 177: see above

examples/Example7/electrons.cu
```diff
@@ -82,7 +83,7 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
     } else {
       geometryStepLength =
           LoopNavigator::ComputeStepAndNextVolume(currentTrack.pos, currentTrack.dir, geometricalStepLengthFromPhysics,
-                                                  currentTrack.currentState, currentTrack.nextState);
+                                                  currentTrack.currentState, currentTrack.nextState, kPush);
```
hahnjo, line 86: We should not push uncoditionally

examples/Example7/gammas.cu
```diff
@@ -55,7 +56,7 @@ __global__ void TransportGammas(Track *gammas, const adept::MParray *active, Sec
     // Check if there's a volume boundary in between.
     double geometryStepLength =
         LoopNavigator::ComputeStepAndNextVolume(currentTrack.pos, currentTrack.dir, geometricalStepLengthFromPhysics,
-                                                currentTrack.currentState, currentTrack.nextState);
+                                                currentTrack.currentState, currentTrack.nextState, kPush);
```
hahnjo, line 59: same

examples/Example9/gammas.cu
```diff
@@ -52,7 +53,7 @@ __global__ void TransportGammas(Track *gammas, const adept::MParray *active, Sec
     // Check if there's a volume boundary in between.
     vecgeom::NavStateIndex nextState;
     double geometryStepLength = LoopNavigator::ComputeStepAndNextVolume(
-        currentTrack.pos, currentTrack.dir, geometricalStepLengthFromPhysics, currentTrack.navState, nextState);
+        currentTrack.pos, currentTrack.dir, geometricalStepLengthFromPhysics, currentTrack.navState, nextState, kPush);
```
hahnjo, line 56: same

examples/Example9/example9.cuh
```diff
@@ -42,6 +42,12 @@ struct Track {
     this->pos      = parent.pos;
     this->navState = parent.navState;
   }
+
+  __host__ __device__ void Print()
+  {
+    printf("pos:(%g,%g,%g), dir:(%g,%g,%g), ene=%g ", pos[0],pos[1],pos[2],dir[0],dir[1],dir[2],energy);
+    navState.Print();
+  }
```
hahnjo, line 50: left-over from debugging?

magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -69,7 +69,9 @@ __host__ __device__ double fieldPropagatorConstBz::ComputeStepAndPropagatedState
     vecgeom::NavStateIndex &next_state)
 {
   using Precision = vecgeom::Precision;
-  double momentumMag = sqrt(kinE * (kinE + 2.0 * mass));
+  constexpr Precision kPush = 10. * vecgeom::kTolerance;
+
+double momentumMag = sqrt(kinE * (kinE + 2.0 * mass));
```
hahnjo, line 74: formatting is off here

magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -94,9 +96,9 @@ __host__ __device__ double fieldPropagatorConstBz::ComputeStepAndPropagatedState
 
   if (charge == 0) {
     if (Relocate) {
-      stepDone = Navigator::ComputeStepAndPropagatedState(position, direction, remains, current_state, next_state);
+      stepDone = Navigator::ComputeStepAndPropagatedState(position, direction, remains, current_state, next_state, kPush);
     } else {
-      stepDone = Navigator::ComputeStepAndNextVolume(position, direction, remains, current_state, next_state);
+      stepDone = Navigator::ComputeStepAndNextVolume(position, direction, remains, current_state, next_state, kPush);
```
hahnjo, line 101: should not push unconditionally

CMakeLists.txt
```diff
@@ -49,7 +49,7 @@ find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
 
 # Find VecGeom geometry headers library
-set(VecGeom_VERSION 1.1.16)
+set(VecGeom_VERSION 1.1.17)
```
hahnjo, line 52: Can you please also update the documentation in `README.md`?

hahnjo (changes requested):

agheata:
OK, I resolved the comments and in the physics examples using the push only in single-precision mode. Discovered also two other places where tolerance was hardcoded to 1e-8

agheata:
We will need VecGeom 1.1.17 as default in Jenkins before merging this.

agheata:
Looks like most of my answers for the resolved issues just disappeared... Re-answering those.

examples/Raytracer_Benchmark/Raytracer.cpp
agheata: For the geometry case, it makes sense to push unconditionally since propagation is always done always to the boundary, and before the propagation contained a push which was worse. The way it is done does not introduce biases because it just computes the distance from a pushed point, compensating for it after, and avoiding starting numerically from the wrong side of the boundary. This removes loopers also in the double-precision case in the raytracing example. In the case of the physics examples, I conditioned the push now for the single-precision case: there the danger may be only for the charged particle transport that may get wrongly pushed during the last step before crossing the boundary.

base/inc/AdePT/BVHNavigator.h
agheata: This method has the contract to cross into a different location, which in rare cases may not happen because relocation is done using the inverse transformation so a mismatch may appear making the algorithm find again the exited volume, but on the wrong side of the boundary (returning negative distance for DistanceToIn, converted to zero and making a looper). I would claim that the modification hides indeed the real fix, which would be to use the direct transformation also for relocation, but allows to recover the looping error w/o introducing a bias.

base/inc/AdePT/LoopNavigator.h
agheata: Indeed, fixed

base/inc/AdePT/LoopNavigator.h
agheata: see above

examples/Example7/electrons.cu
agheata: pushing only in SP mode

examples/Example7/gammas.cu
agheata: pushing only in SP mode

examples/Example9/gammas.cu
agheata: pushing only in SP mode

examples/Example9/example9.cuh
agheata: indeed, removed

magneticfield/inc/fieldPropagatorConstBz.h
agheata: fixed

magneticfield/inc/fieldPropagatorConstBz.h
agheata: pushing only in SP mode

CMakeLists.txt
agheata: done

WitekPokorski (approved):
As discussed, ready to merged.

WitekPokorski (approved):

PR #139, opened 2021-09-08
WIP:  Example12 - create a VecGeom geometry from Geant4

jonapost:
This example will read a geometry using the Geant4 GDML reader, and then convert it to VecGeom in memory using the G4VecGeomNav repository's G4VecGeomConverter class.

This was modified to set the Material Index in MaterialPtr of VecGeom's LogicalVolume and the (G4HepEM) Material Cuts Couple index in LogicalVolume's MateriaCutsPtr.

Current state: WORK IN PROGRESS - not guaranteed to compile or work.

I will extract the part where the geometry is converted from Geant4 to VecGeom, and try it out in example13 instead of adding this example.

### DIFF
--- a/examples/Example12/CMakeLists.txt
+++ b/examples/Example12/CMakeLists.txt
@@ -1,0 +1,13 @@
+# SPDX-FileCopyrightText: 2021 CERN
+# SPDX-License-Identifier: Apache-2.0
+
+if(NOT TARGET G4HepEm::g4HepEm)
+  message(STATUS "Disabling example11 (needs G4HepEm)")
+  return()
+endif()
+
+add_executable(example11 example11.cpp example11.cu electrons.cu gammas.cu)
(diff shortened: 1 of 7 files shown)


agheata:
@jonapost is this example still relevant/pursued?

jonapost:
Yes, it's relevant - as far as I know.

WitekPokorski:
I doubt is relevant if it was not touched for 2 years (!!!). I do not see a point keeping such old stuff which will be never completed.

agheata:
@jonapost  when I asked about being relevant I was referring to the fact that the PR is based on an old example, which itself is not so relevant anymore. I can see the merit in having the geometry read only once from GDML, but I wanted to understand if the G4 to VecGeom converter is more or less ready and if we are planning to do any work on this integration in the near future.

agheata:
Closing this since it is no longer relevant after the adoption of the `g4vg` converter

PR #140, opened 2021-09-09
Example12 featuring TestEm3-like transport in arbitrary geometry imported from GDML.

agheata:
The example implements TestEm3-like workflow, but the geometry can be taken from an arbitrary GDML file. 
* The geometry is first read by the Geant4 parser and the G4HepEm state is initialized based on this
* The geometry is then imported via vecgeom::vgdml and the geometry hierarchy is visited in parallel with the Geant4 one to match G4LogicalVolume<->vecgeom::LogicalVolume and build the LUT (logical volume index) -> (G4HepEm material-cut couple index)
* GPU data is initialized and showering/scoring is done as in TestEm3
* Scoring is done per placed volume, no sensitivity selection

I tested this with cms2018.gdml, it works and gives reasonable (at first glance) results (the energy deposit per primary equal to the primary energy) when shooting electrons of 10 GeV perpendicular to the beam axis.

### DIFF
--- a/examples/Example12/CMakeLists.txt
+++ b/examples/Example12/CMakeLists.txt
@@ -1,0 +1,42 @@
+# SPDX-FileCopyrightText: 2021 CERN
+# SPDX-License-Identifier: Apache-2.0
+
+if(NOT TARGET G4HepEm::g4HepEm)
+  message(STATUS "Disabling example12 (needs G4HepEm)")
+  return()
+endif()
+
+if(Geant4_FOUND)
+  if(NOT Geant4_gdml_FOUND)
+    message(STATUS "Disabling example12 (needs Geant4 with GDML support)")
+    return()
+  endif()
+else()
+  message(STATUS "Disabling example12 (needs Geant4)")
+  return()
+endif()
+
+# example12 is the AdePT demo example using material cuts as defined in the input gdml file
+add_executable(example12
+  example12.cpp
+  example12.cu
+  electrons.cu
+  gammas.cu)
+target_link_libraries(example12 
+  PRIVATE
+    AdePT
+    CopCore::CopCore
+    VecGeom::vecgeom
+    VecGeom::vecgeomcuda_static
+    VecGeom::vgdml
+    ${Geant4_LIBRARIES}
+    G4HepEm::g4HepEmData
+    G4HepEm::g4HepEmInit
+    G4HepEm::g4HepEmRun
+    CUDA::cudart)
+
+set_target_properties(example12 PROPERTIES CUDA_SEPARABLE_COMPILATION ON CUDA_RESOLVE_DEVICE_SYMBOLS ON)
+
+# Tests
+add_test(NAME example12
+  COMMAND $<TARGET_FILE:example12> -gdml_file ${CMAKE_BINARY_DIR}/cms2018.gdml)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,10 @@ find_package(VecGeom ${VecGeom_VERSION} REQUIRED)
 message(STATUS "Using VecGeom version ${VecGeom_VERSION}")
 # make sure we import VecGeom architecture flags - is this needed?
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${VECGEOM_CXX_FLAGS}")
+# Make sure VecGeom::vgdml is enabled
+if(NOT TARGET VecGeom::vgdml)
+  message(FATAL_ERROR "AdePT requires VecGeom compiled with GDML support")
+endif()
 
 # Find Geant4, optional for now
 find_package(Geant4 QUIET)
@@ -80,7 +84,9 @@ add_subdirectory(external)
 
 if(BUILD_TESTING)
   set(TESTING_GDML "${PROJECT_BINARY_DIR}/trackML.gdml")
+  set(CMS2018_GDML "${PROJECT_BINARY_DIR}/cms2018.gdml")
   file(DOWNLOAD https://gitlab.cern.ch/VecGeom/VecGeom/raw/master/persistency/gdml/gdmls/trackML.gdml "${TESTING_GDML}")
+  file(DOWNLOAD https://gitlab.cern.ch/VecGeom/VecGeom/raw/master/persistency/gdml/gdmls/cms2018.gdml "${CMS2018_GDML}")
 endif()
 
 # Builds...
--- a/examples/CMakeLists.txt
+++ b/examples/CMakeLists.txt
@@ -14,7 +14,7 @@ add_subdirectory(Example8)
 add_subdirectory(Example9)
 add_subdirectory(Example10)
 add_subdirectory(Example11)
+add_subdirectory(Example12)
 add_subdirectory(TestEm3)
 add_subdirectory(TestEm3MT)
 add_subdirectory(ECS)
-
--- a/examples/Example12/README.md
+++ b/examples/Example12/README.md
@@ -1,0 +1,53 @@
+<!--
+SPDX-FileCopyrightText: 2021 CERN
+SPDX-License-Identifier: CC-BY-4.0
+-->
+# Example12 example
+Example demonstrating particle transportation on GPUs in arbitrary geometry read from a GDML file, with the same GPU workflow as Example 11.
+
+ * arbitrary geometry via gdml file (tested with cms2018.gdml from VecGeom persistency/gdml/gdmls folder) and optionally a magnetic field with constant Bz,
+ * geometry read as Geant4 geometry, reading in regions and cuts, to initialize G4HepEm data
+ * geometry read then into VecGeom, and synchronized to GPU
+ * G4HepEm material-cuts couple indices mapped to VecGeom logical volume id's
+ * physics processes for e-/e+ and gammas using G4HepEm.
+ * scoring per placed volume, no sensitive detector feature
+
+Electrons, positrons, and gammas are stored in separate containers in device memory.
+Free positions in the storage are handed out with monotonic slot numbers, slots are not reused.
+Active tracks are passed via three queues per particle type (see `struct ParticleQueues`).
+Results are reproducible using one RANLUX++ state per track.
+
+Additionally, the kernels score energy deposit and the charged track length per volume.
+
+### Kernels
+
+This example uses one stream per particle type to launch kernels asynchronously.
+They are synchronized via a forth stream using CUDA events.
+
+#### `TransportElectrons<bool IsElectron>`
+
+1. Determine physics step limit.
+2. Call magnetic field to get geometry step length.
+3. Apply continuous effects; kill track if stopped.
+4. If the particle reaches a boundary, perform relocation.
+5. If not, and if there is a discrete process:
+ 1. Sample the final state.
+ 2. Update the primary and produce secondaries.
+
+#### `TransportGammas`
+
+1. Determine the physics step limit.
+2. Query VecGeom to get geometry step length (no magnetic field for neutral particles!).
+3. If the particle reaches a boundary, perform relocation.
+4. If not, and if there is a discrete process:
+ 1. Sample the final state.
+ 2. Update the primary and produce secondaries.
+
+#### `FinishIteration`
+
+Clear the queues and return the tracks in flight.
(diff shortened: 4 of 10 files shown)


examples/CMSapp/CMSapp.cpp
```diff
@@ -0,0 +1,266 @@
...
+  {
+    std::cerr << "Failed to read geometry from GDML file '" << gdml_file << "'" << std::endl;
+    return nullptr;
+  }
+
+  const vecgeom::VPlacedVolume *world = vecgeom::GeoManager::Instance().GetWorld();
+  if (world == nullptr) {
+    std::cerr << "GeoManager world volume is nullptr" << std::endl;
+    return nullptr;
+  }
```
hahnjo, line 90: one of the two `if` statements is not correctly formatted, please run `clang-format` on newly added code

examples/CMSapp/CMSapp.cpp
```diff
@@ -0,0 +1,266 @@
...
+    region->UsedInMassGeometry(true); // make sure all regions are marked as used
+
+  // - UPDATE COUPLES
+  std::cout << "updating material-cut couples based on " << G4RegionStore::GetInstance()->size() << " regions ...\n";
+  G4ProductionCutsTable *theCoupleTable = G4ProductionCutsTable::GetProductionCutsTable();
+  theCoupleTable->UpdateCoupleTable(world);
+  return world;
+}
+
+const vecgeom::cxx::VPlacedVolume *InitVecGeom(std::string gdml_file, int cache_depth)
```
hahnjo, line 74: `const std::string &gdml_file`

examples/CMSapp/CMSapp.cpp
```diff
@@ -0,0 +1,266 @@
...
+#include <VecGeom/base/Stopwatch.h>
+#include <VecGeom/management/GeoManager.h>
+#include <VecGeom/management/BVHManager.h>
+#include <VecGeom/gdml/Frontend.h>
+#include <VecGeom/management/CudaManager.h>
+
+#include <CopCore/SystemOfUnits.h>
+#include <AdePT/ArgParser.h>
+
+const G4VPhysicalVolume *InitGeant4(std::string gdml_file)
```
hahnjo, line 34: `const std::string &gdml_file`

examples/CMSapp/CMSapp.cpp
```diff
@@ -0,0 +1,266 @@
...
+  int NumVolumes = vecgeom::GeoManager::Instance().GetRegisteredVolumesCount();
+  int NumPlaced  = vecgeom::GeoManager::Instance().GetPlacedVolumesCount();
+
+  // Scoring is done per placed volume (for now...)
+  double *chargedTrackLength = new double[NumPlaced];
+  double *energyDeposit      = new double[NumPlaced];
+  ScoringPerVolume scoringPerVolume;
+  scoringPerVolume.chargedTrackLength = chargedTrackLength;
+  scoringPerVolume.energyDeposit      = energyDeposit;
+  GlobalScoring globalScoring;
```
hahnjo, line 210: @jonapost and me had thought about scoring per volume. That would drastically reduce the number of numbers to output and compare and gives a reasonable abstraction that can be checked against Geant4.

examples/CMSapp/CMSapp.cu
```diff
@@ -0,0 +1,452 @@
...
+  dataOnDevice.fTheSBTableData_gpu  = nullptr;
+  dataOnDevice.fTheGammaData_gpu    = nullptr;
+
+  COPCORE_CUDA_CHECK(cudaMemcpyToSymbol(g4HepEmData, &dataOnDevice, sizeof(G4HepEmData)));
+}
+
+static void FreeG4HepEm(G4HepEmState *state)
+{
+  FreeG4HepEmData(state->fData);
+}
```
hahnjo, line 77: AFAICT G4HepEm is now initialized in `CMSapp.cpp`, so memory should also be freed there.

examples/CMSapp/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+# Also need VecGeom 1.1.17 or newer
+if(VecGeom_VERSION VERSION_LESS 1.1.17)
+  message(STATUS "Disabling CMSapp (requires VecGeom 1.1.17 or newer to support full GDML parsing)")
+  return()
+else()
+  if(NOT TARGET VecGeom::vgdml)
+    message(STATUS "Disabling CMSapp (requires VecGeom with GDML support)")
+    return()
+  endif()
+endif()
```
hahnjo, line 28: I think this already checked in the top-level `CMakeLists.txt`

examples/CMSapp/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
...
+
+# CMSapp is the AdePT application using Geant4 to initialize G4HepEm data structures for the CMS geometry
```
hahnjo, line 30: I think this application does more than just "initialize G4HepEm data structures", right?

hahnjo (commented):
As far as I understand, CI fails because `cms2018.gdml` is not downloaded / found.

agheata:
@hahnjo the test is now passing

agheata:
Renamed to Example12 as discussed, since the example is not CMS-dependent.

examples/Example12/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+# Needs VecGeom::vgdml
+if(NOT TARGET VecGeom::vgdml)
+  message(STATUS "Disabling Example12 (requires VecGeom with GDML support)")
+  return()
+endif()
+
+# Example12 is the AdePT demo example using material cuts as defined in the input gdml file
+add_executable(Example12
+  Example12.cpp
+  Example12.cu
```
hahnjo, line 28: Lower-case `example12` please for consistency - I just spend too much of my time checking and re-checking why I have no binary `example12`...

examples/Example12/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+else()
+  message(STATUS "Disabling Example12 (needs Geant4)")
+  return()
+endif()
+
+# Needs VecGeom::vgdml
+if(NOT TARGET VecGeom::vgdml)
+  message(STATUS "Disabling Example12 (requires VecGeom with GDML support)")
+  return()
+endif()
```
hahnjo, line 23: I think this should be moved to the top-level: All existing examples already require GDML support, `example12` is not special about that.

examples/Example12/Example12.cpp
```diff
@@ -0,0 +1,268 @@
...
+  int nphysical = 0;
+
+  int nvolumes   = vecgeom::GeoManager::Instance().GetRegisteredVolumesCount();
+  int *mcc_index = new int[nvolumes];
+  memset(mcc_index, 0, nvolumes * sizeof(int));
+
+  // recursive geometry visitor lambda matching one by one Geant4 and VecGeom logical volumes
+  // (we need to make sure we set the right MCC index to the right volume)
+  typedef std::function<void(G4LogicalVolume const *, vecgeom::LogicalVolume const *)> func_t;
+  func_t visitAndSetMCindex = [&](G4LogicalVolume const *g4vol, vecgeom::LogicalVolume const *vol) {
```
hahnjo, line 107: Just `auto`, that's what it was invented for

examples/Example12/Example12.cpp
```diff
@@ -0,0 +1,268 @@
...
+    int nd         = g4vol->GetNoDaughters();
+    auto daughters = vol->GetDaughters();
+    assert(nd == daughters.size() && "Mismatch in number of daughters");
```
hahnjo, line 110: Would it make sense to convert these `assert`s to checks that are also compiled into `Release` builds? This code is not performance critical, but if somebody tries a new GDML they probably want to know if the matching worked...

examples/Example12/Example12.cu
```diff
@@ -0,0 +1,452 @@
...
+__constant__ __device__ struct G4HepEmData g4HepEmData;
+
+__constant__ __device__ int *MCIndex = nullptr;
+
+__constant__ __device__ int Zero = 0;
+
+// struct G4HepEmState {
+//  G4HepEmData data;
+//  G4HepEmParameters parameters;
+//};
```
hahnjo, line 45: remove

examples/Example12/Example12.cpp
```diff
@@ -0,0 +1,268 @@
...
+    std::cout << std::endl;
+    std::cout << "  Layer   Charged-TrakL [mm]   Energy-Dep [MeV]" << std::endl;
+    for (int i = 0; i < NbOfLayers; i++) {
+      int layerVolume        = 2 + i * (1 + NbOfAbsorbers);
+      double chargedTrackLen = chargedTrackLength[layerVolume + 1] + chargedTrackLength[layerVolume + 2];
+      double energyDep       = energyDeposit[layerVolume + 1] + energyDeposit[layerVolume + 2];
+      std::cout << std::setw(5) << i << std::setw(20) << chargedTrackLen / copcore::units::mm << std::setw(20)
+                << energyDep / copcore::units::MeV << std::endl;
+    }
+  */
```
hahnjo, line 263: remove if not used (I still think that accumulating per material would be good for comparison with Geant4)

examples/Example12/Example12.cpp
```diff
@@ -0,0 +1,268 @@
...
+
+  visitAndSetMCindex(g4world->GetLogicalVolume(), world->GetLogicalVolume());
+  std::cout << "Visited " << nphysical << " matching physical volumes\n";
+  return mcc_index;
+}
+
+void FreeG4HepEm(G4HepEmState *state)
+{
+  FreeG4HepEmData(state->fData);
+}
```
hahnjo, line 141: should also be called in this file!

examples/Example12/Example12.cu
```diff
@@ -0,0 +1,452 @@
...
+  dataOnDevice.fTheSBTableData_gpu  = nullptr;
+  dataOnDevice.fTheGammaData_gpu    = nullptr;
+
+  COPCORE_CUDA_CHECK(cudaMemcpyToSymbol(g4HepEmData, &dataOnDevice, sizeof(G4HepEmData)));
+}
+
+static void FreeG4HepEm(G4HepEmState *state)
+{
+  FreeG4HepEmData(state->fData);
+}
```
hahnjo, line 77: should not be in this file

examples/Example12/Example12.cu
```diff
@@ -0,0 +1,452 @@
...
+    COPCORE_CUDA_CHECK(cudaFree(particles[i].slotManager));
+
+    COPCORE_CUDA_CHECK(cudaFree(particles[i].queues.currentlyActive));
+    COPCORE_CUDA_CHECK(cudaFree(particles[i].queues.nextActive));
+
+    COPCORE_CUDA_CHECK(cudaStreamDestroy(particles[i].stream));
+    COPCORE_CUDA_CHECK(cudaEventDestroy(particles[i].event));
+  }
+
+  FreeG4HepEm(state);
```
hahnjo, line 451: should not be called from the `.cu` file

examples/Example12/Example12.h
```diff
@@ -0,0 +1,33 @@
...
+  unsigned long long numElectrons;
+  unsigned long long numPositrons;
+};
+
+struct ScoringPerVolume {
+  double *energyDeposit;
+  double *chargedTrackLength;
+};
+
+static void FreeG4HepEm(G4HepEmState *state);
```
hahnjo, line 27: forward-declaring a `static` function in a header makes no sense

examples/Example12/Example12.cpp
agheata: not yet for recursive lambdas unfortunatelly:
```/home/agheata/geant_src/AdePT/examples/Example12/example12.cpp:129:7: error: use of ‘visitAndSetMCindex’ before deduction of ‘auto’
       visitAndSetMCindex(g4pvol->GetLogicalVolume(), pvol->GetLogicalVolume());
       ^~~~~~~~~~~~~~~~~~
```

examples/Example12/Example12.cpp
agheata: Yes it would, because it is in the initialization part anyway

agheata:
Thanks @hahnjo this should be ready now

hahnjo (approved):

PR #141, opened 2021-09-29
Update G4HepEm and implement photoelectric effect

hahnjo:


agheata:
Looks good, besides the compilation errors that I could not see explicitly. What is the impact of tracking the photo-electrons to the timing? If we merge #140 first, the same changes will have to be applied there.

hahnjo:
> What is the impact of tracking the photo-electrons to the timing?

Not much different (I think gammas will, on average, travel less far due to the cross section for higher energies), but I didn't look in detail because it's completely irrelevant until we have MSC.

hahnjo:
@agheata I worked around the compilation problem with a change in G4HepEm (that Mihaly and me wanted to do anyway), should be good to merge now.

agheata (approved):
LGTM

PR #142, opened 2021-10-05
Integrate Multiple Scattering

hahnjo:
... into TestEm3(MT) and add new example.


hahnjo:
Validation of TestEm3 without magnetic field looks good; with `B_z = 1T` the charged track length and the energy deposit are too low. The reason is that the field propagator sums up the straight line segments instead of integrating the step length along the curve.

examples/Example13/electrons.cu
```diff
@@ -0,0 +1,341 @@
...
+          reducedSafety = sFact * safety;
+
+          // 1b. Far away from geometry boundary:
+          if (reducedSafety > 0.0 && dispR <= reducedSafety) {
+            currentTrack.pos += displacement;
+            // 2. Push to boundary:
+          } else if (reducedSafety > kGeomMinLength) {
+            currentTrack.pos += displacement * (reducedSafety / dispR);
+          }
+          // 3. Very small safety: do nothing.
```
agheata, line 160: As far as I can read from the code, this branch has also all cases where the displacement exceeds the reduced safety. That means applying MSC depends on the accuracy geometry estimates the safety, at the limit, if geometry returns 0 safety everywhere, there will be no displacement. Shouldn't we try to apply the displacement in all cases, and only check if the boundary gets crossed when we are in this condition? Or maybe I am missing something?

examples/Example13/electrons.cu
```diff
@@ -0,0 +1,341 @@
...
+    theTrack->SetOnBoundary(currentTrack.navState.IsOnBoundary());
+    theTrack->SetCharge(Charge);
+    G4HepEmMSCTrackData *mscData = elTrack.GetMSCTrackData();
+    mscData->fIsFirstStep        = currentTrack.initialRange < 0;
+    mscData->fInitialRange       = currentTrack.initialRange;
+
+    // Compute safety, needed for MSC step limit.
+    double safety = 0;
+    if (!currentTrack.navState.IsOnBoundary()) {
+      safety = BVHNavigator::ComputeSafety(currentTrack.pos, currentTrack.navState);
```
agheata, line 68: I see that the safety is cached in the track (good!) but I don't see where the cached value is reused , since it seems to be systematically recomputed here. It seems to me we should do:
```
safety = track->GetSafety();
if (safety < vecgeom::kTolerance)
  safety = BVHNavigator::ComputeSafety(currentTrack.pos, currentTrack.navState);
```

examples/Example13/electrons.cu
```diff
@@ -0,0 +1,341 @@
...
+    if (!nextState.IsOnBoundary()) {
+      const double *mscDisplacement = mscData->GetDisplacement();
+      vecgeom::Vector3D<Precision> displacement(mscDisplacement[0], mscDisplacement[1], mscDisplacement[2]);
+      const double dLength2            = displacement.Length2();
+      constexpr double kGeomMinLength  = 5 * copcore::units::nm;          // 0.05 [nm]
+      constexpr double kGeomMinLength2 = kGeomMinLength * kGeomMinLength; // (0.05 [nm])^2
+      if (dLength2 > kGeomMinLength2) {
+        const double dispR = std::sqrt(dLength2);
+        // Estimate safety by subtracting the geometrical step length.
+        safety -= geometryStepLength;
```
agheata, line 140: I think we should subtract from the cached safety value here

base/inc/AdePT/BVHNavigator.h
```diff
@@ -158,6 +158,26 @@ class BVHNavigator {
...
+  {
+    VPlacedVolumePtr_t pvol = state.Top();
+    vecgeom::Transformation3D m;
+    state.TopMatrix(m);
+    vecgeom::Vector3D<vecgeom::Precision> localpoint = m.Transform(globalpoint);
+
+    // need to calc DistanceToOut first
+    vecgeom::Precision safety = pvol->SafetyToOut(localpoint);
+
+    if (safety > 0 && pvol->GetDaughters().size() > 0) {
```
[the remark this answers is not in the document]
hahnjo: Of course the check for the mother safety is needed because it may be the closest boundary, even if not on it. In general, this method does exactly do what @amadio implemented in `BVHSafetyEstimator`.

examples/Example13/electrons.cu
hahnjo: No, HepEm needs the accurate safety so it must be recomputed (at least that is what `G4HepEmProcess` does and all MSC models in Geant4). There is no point in caching the value and using more memory, if we cannot use it anyway.

examples/Example13/electrons.cu
```diff
@@ -0,0 +1,341 @@
...
+    // double physicalStepLength = elTrack.GetPStepLength();
+    int winnerProcessIndex = theTrack->GetWinnerProcessIndex();
+    // Leave the range and MFP inside the G4HepEmTrack. If we split kernels, we
+    // also need to carry them over!
+
+    // Check if there's a volume boundary in between.
+    double geometryStepLength;
+    vecgeom::NavStateIndex nextState;
+    if (BzFieldValue != 0) {
+      geometryStepLength = fieldPropagatorBz.ComputeStepAndPropagatedState</*Relocate=*/false, BVHNavigator>(
```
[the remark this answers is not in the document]
hahnjo: No, because the particle needs to travel across the curved path. There may be some optimizations where we can avoid querying the geometry if within the safety, but that wasn't even worth it with the `LoopNavigator` so I doubt it will be with the `BVHNavigator`

examples/Example13/electrons.cu
hahnjo: `safety` is the value computed at the beginning of the step, I don't understand where I should subtract from otherwise. (This case is only to avoid calling `BVHNavigator::ComputeSafety` in cases where the safety is much longer than the step length and we can apply the displacement right away.)

examples/Example13/electrons.cu
hahnjo: All I can say is that this is exactly what Geant4 does in the default case. There are more accurate algorithms for the step limitation, but they won't be implemented in G4HepEm.

examples/Example13/electrons.cu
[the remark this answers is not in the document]
agheata: I don't get the argument with the curved path, which would allow even longer geometry steps to stay within safety. I mean if the physics step is smaller than safety, there is no way that a boundary crossing may happen in this range, magnetic field or not. So why compute the distance?  And I would not extrapolate based on some old observation with LoopNavigator and a single test case. If physics is more dense, I would expect many physics steps in the same volume not needing to know the precise distance to boundary. So I would definitely try it, I don't agree it is a minor optimization.

examples/Example13/electrons.cu
agheata: @mnovak42 any comment on this issue?

examples/Example13/electrons.cu
agheata: So I get from this that MSC is indeed applied only within the safety limit, trying not to get out of the current volume, and limiting the step to basically the safety. I'll wait maybe for @mnovak42 comment on the above discussion. I would claim that even if this is the case, working with the reduced safety would be equivalent from MSC point of view, and if this is true it can get rid of several calls to the actual distance. Anyway, `accurate` safety is not a requirement for geometry, which does the best it can returning an overestimate.

mnovak42: MSC is not limiting the step such that it can be only at most safety (we would have difficulties crossing any boundary in that case). However, MSC (at least the currently employed step limit algorithm) limits the step based on some information regarding the safety and contains some optimisation based on the safety as well. 

What we do in G4 is:
- we don't use immediately the safety re-computed in the current pre-step point but instead 
- we try to see if the current step limit, given by all other physics, is inside the updated previous safety 
-  this updated safety (called pre-safety) is the previous pre-step point safety minus the previous geometrical step length and displacement if any)    
- the exact safety is computed only if the current physics step would be longer than the updated safety 
- we can do this in G4 since we can interact with the transportation inside the MSC step limit using the `SafetyHelper`
 
When using G4HepEm, the run-time library do not contain any G4 dependence. Therefore, we need to assume the second case above and recompute the safety at the pre-step point (still in the connection layer to G4) before we would ask G4HepEmRun library to compute the step limit. This do not lead to any differences in the results (e.g. the step limit is eventually exactly the same in G4HepEm and native G4) while causing some overhead (especially that we do not even have the physics step length, given by all other physics, at this point.)

So this is why we do not cache and update the safety by subtracting the step lengths done after the exact computation.

mnovak42: >  Anyway, `accurate` safety is not a requirement for geometry, which does the best it can returning an overestimate.

As far as I knew safety should always be an underestimate (<=) to the distance to the closest boundary and never an overestimate. This is why I say above that in theory (according to definition) it could always be zero.

base/inc/AdePT/BVHNavigator.h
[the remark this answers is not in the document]
agheata: Up to you @hahnjo to apply this change here or not and resolve the conversation, I will leave a general comment in this PR to remember to test if the `safety > tolerance` case gives a real reduction of the full daughter safety checks.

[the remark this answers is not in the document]
hahnjo: I think I would prefer to keep the initial implementation free of micro-optimizations like this. Let's do the fully correct thing for now and check back later if there would be a gain.

PR #148, opened 2021-10-15
Field propagation correction and optimization

agheata:
Implementing a better estimation of the last chord step before crossing the boundaries in field. Using the safety computed for charged particles in the helix propagator. This reduces the run time for exa13 with a field of 3.8T and default gun by a factor of 3.

### DIFF
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -135,10 +135,14 @@ __host__ __device__ double fieldPropagatorConstBz::ComputeStepAndNextVolume(
         //   Currently it's controlled/limited by the acceptable step size ie. 'safeLength' )
         position = position + move * chordDir;
 
-        // Primitive approximation of end direction ...
+        // Primitive approximation of end direction and move to the crossing point ...
         double fraction = chordLen > 0 ? move / chordLen : 0.0;
         direction       = direction * (1.0 - fraction) + endDirection * fraction;
         direction       = direction.Unit();
+        // safeMove is how much the track would have been moved if not hitting the boundary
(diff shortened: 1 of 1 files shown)


agheata:
The problem comes from the fact that the current field propagator forcefully exits the loop if the number of chord iterations is more than 10 and the physics or geometry limits are not hit. This chops the step of some low energy electrons without notification. Setting the limit to 1000 steps makes the output matching much better with or w/o safety optimization:
```
no safety:
Run time: 463.632
Mean energy deposit          9.97569 GeV
Mean number of gamma         3631.88
Mean number of e-            12053.5
Mean number of e+            316.475
Mean number of charged steps 34435.7
Mean number of neutral steps 34347.9
Mean number of hits          33385.1

safety-optimized:
Run time: 425.073
Mean energy deposit          9.97504 GeV
Mean number of gamma         3632.57
Mean number of e-            12058.2
Mean number of e+            316.532
Mean number of charged steps 34862.7
Mean number of neutral steps 34340.6
Mean number of hits          33793.9
```

I think the proper approach is to accept that field propagation may take too long  for e.g. low-energy electrons curling near a boundary, so to pass a success flag via the interface, leaving the caller the responsibility to kill the track. I prefer doing this in a separate PR. The safety optimization becomes less efficient if not limiting the iterations, because the sampled safety point before propagation is in average too close to the boundary. Working on a better version resampling the safety during propagation in certain conditions.

agheata:
While implementing the safety optimization I noticed that the cut on the number of magnetic field iterations has a huge impact on performance. Depending on the cut, I plotted the simulation time (with and w/o safety optimization), but also other things. It seems the current algorithm deserves extra attention, and we need to understand the impact of the real CMS cuts.
Screenshot 2021-10-21 at 18 16 30

PR #152, opened 2021-10-20
Changed field propagator interface to account for error and modified examples to use this.

agheata:
This addresses a bug in the field propagator, currently hard-limited to 10 iterations. An interface change was needed to:
- allow the user define the cutoff in number of iterations, if not used set to 100
- return the propagation success flag, set to false if the cutoff is being hit, giving the responsibility to the caller to kill the track (likely a slow electron)
- require the user to call with a safety value computed at the starting point. However, if the value is not computed, the method will accept 0.0, with the meaning (not using it). This feature will be preserved in the further optimizations

The safety value is not used in this PR, but allows further optimizations based on safety without changing the interface.

After this change, the output for the examples doing propagation in field is numerically identical (checked for example9 to 12 and TestEM3). This addresses the discussion opened on #148

### DIFF
--- a/examples/Example10/electrons.cu
+++ b/examples/Example10/electrons.cu
@@ -77,9 +77,17 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
       // also need to carry them over!
 
       // Check if there's a volume boundary in between.
+      bool propagated = true;
       double geometryStepLength = fieldPropagatorBz.ComputeStepAndNextVolume(
           currentTrack.energy, Mass, Charge, geometricalStepLengthFromPhysics, currentTrack.pos, currentTrack.dir,
-          currentTrack.currentState, currentTrack.nextState);
+          currentTrack.currentState, currentTrack.nextState, propagated);
+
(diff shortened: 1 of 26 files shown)


agheata:
I'm fine with the change of the name of the method (no strong opinion about it). I can also accommodate returning the distance as before and pushing the safety parameter to the end to get a default value, but the examples need to be changed even if not using safety, to account for the possible error in propagation. Let's try to merge first #149 and adapt this after.

hahnjo (approved):

hahnjo:
FWIW the conditions are met 77 times in `TestEm3` with magnetic field when simulating 1000 electrons (the other examples are indeed not affected, possibly because they don't have materials yet and `example12` and `example13` don't enable magnetic fields).

agheata:
> Yes, 321 tracks are killed in propagation when simulating 100 electrons in `example13` with a magnetic field of Bz=1T.

Indeed, if in a particular example this becomes significant, we have the means to increase the default number of propagation sub-steps tolerated (default is 100)

PR #156, opened 2021-11-03
Use intrinsic for 128 bit multiplication in RANLUX++

hahnjo:
This yields single digit percent improvements, depending on the ratio of time spent in the RNG; results are identical.


hahnjo:
ping @agheata @WitekPokorski

agheata (approved):

PR #157, opened 2021-11-19
Optional NVTX annotations for example11 and 13

hageboeck:
This adds NVTX annotations for example 11 and 13, e.g.
image

They only have an effect, though, if the executable is linked against NVTX, e.g. by doing:
```bash
cmake -DNVTX=On ...
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -103,6 +103,14 @@ target_include_directories(AdePT
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 target_link_libraries(AdePT INTERFACE CopCore::CopCore)
 
+# Optional library to activate NVTX annotations for profiling:
+set(NVTX OFF CACHE BOOL "Add NVTX instrumentation for profiling (only for annotated examples)")
+add_library(NVTX INTERFACE)
+if(NVTX)
+  target_link_libraries(NVTX INTERFACE nvToolsExt)
+  target_compile_definitions(NVTX INTERFACE -DUSE_NVTX)
+endif()
+
 add_subdirectory(tracking)
 add_subdirectory(test)
 add_subdirectory(examples)
--- a/examples/Example13/example13.cu
+++ b/examples/Example13/example13.cu
@@ -7,6 +7,7 @@
 #include <AdePT/Atomic.h>
 #include <AdePT/BVHNavigator.h>
 #include <AdePT/MParray.h>
+#include <AdePT/NVTX.h>
 
 #include <CopCore/Global.h>
 #include <CopCore/PhysicalConstants.h>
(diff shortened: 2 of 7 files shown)


CMakeLists.txt
```diff
@@ -103,6 +103,11 @@ target_include_directories(AdePT
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>)
 target_link_libraries(AdePT INTERFACE CopCore::CopCore)
 
+# Optional library to activate NVTX annotations for profiling:
+add_library(NVTX INTERFACE)
```
amadio, line 107: Shouldn't you have an `if` around this to make it optional?

hageboeck: I don't know. It's still unused if you merged this right now.

I hadn't decided where to put the option to enable/disable NVTX. I just add `NVTX` to the list of libraries for the example I'm interested in. Alternatively, one could enable it for the entire project using
```
cmake -DNVTX=On <sourceDir>
```
and then the `if` would be useful. It may boil down to where you want to enable the tracing:
- In the `CMakeLists.txt` of the respective example
- Globally when configuring
- Or somewhere else

What do you think?

amadio: I like the idea of the interface library. I thought it was used unconditionally. Having a CMake switch to turn this stuff on would be nice indeed. A global switch + per-example support should be fine, as not all examples have the tags anyway.

hageboeck: Here's the switch:
```bash
cmake -DNVTX=On .
```

examples/Example13/example13.cu
```diff
@@ -275,12 +279,14 @@ void example13(int numParticles, double energy, int batch, const int *MCIndex_ho
     stats->inFlight[ParticleType::Gamma]    = 0;
 
     constexpr int MaxBlocks        = 1024;
-    constexpr int TransportThreads = 32;
+    constexpr int TransportThreads = 128;
```
hahnjo, line 282: I don't think this should be in here

examples/Example13/example13.cu
```diff
@@ -275,12 +279,14 @@ void example13(int numParticles, double energy, int batch, const int *MCIndex_ho
...
     int transportBlocks;
 
     int inFlight;
     int loopingNo         = 0;
     int previousElectrons = -1, previousPositrons = -1;
+    int maxInFlight = 0, localMinInFlight = 0;
```
hahnjo, line 288: You should maybe initialize both of them to `chunk` in the beginning, which is the initial number of particles in flight?

examples/Example13/example13.cu
hageboeck: Ouch, thanks for spotting this. 😅

amadio (approved):
LGTM

PR #158, opened 2021-11-19
Add option to distribute particles over eta,phi in example13.

hageboeck:
Implement a simple rotating particle gun that generates flat
distributions in eta and phi between -5 and 5.
This option is off by default.

### DIFF
--- a/examples/Example13/example13.cu
+++ b/examples/Example13/example13.cu
@@ -107,7 +107,7 @@ __global__ void InitParticleQueues(ParticleQueues queues, size_t Capacity)
 
 // Kernel function to initialize a set of primary particles.
 __global__ void InitPrimaries(ParticleGenerator generator, int startEvent, int numEvents, double energy,
-                              const vecgeom::VPlacedVolume *world, GlobalScoring *globalScoring)
+                              const vecgeom::VPlacedVolume *world, GlobalScoring *globalScoring, bool rotatingParticleGun)
 {
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < numEvents; i += blockDim.x * gridDim.x) {
     Track &track = generator.NextTrack();
@@ -120,7 +120,17 @@ __global__ void InitPrimaries(ParticleGenerator generator, int startEvent, int n
     track.initialRange = -1.0;
 
     track.pos = {0, 0, 0};
-    track.dir = {1.0, 0, 0};
+    if (rotatingParticleGun) {
+      // Generate particles flat in phi and in eta between -5 and 5. We'll lose the far forwards ones, so no need to simulate.
+      const double phi = 2. * M_PI * track.rngState.Rndm();
+      const double eta = -5. + 10. * track.rngState.Rndm();
+      track.dir = {
+        cos(phi) / cosh(eta),
+        sin(phi) / cosh(eta),
+                   tanh(eta)};
+    } else {
+      track.dir = {1.0, 0, 0};
+    }
     track.navState.Clear();
     BVHNavigator::LocatePointIn(world, track.pos, track.navState, true);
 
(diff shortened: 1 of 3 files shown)


examples/Example13/example13.cu
```diff
@@ -120,7 +120,13 @@ __global__ void InitPrimaries(ParticleGenerator generator, int startEvent, int n
     track.initialRange = -1.0;
 
     track.pos = {0, 0, 0};
-    track.dir = {1.0, 0, 0};
+    if (rotatingParticleGun) {
+      const double phi = 2. * 3.14159 * i / numEvents;
+      const double eta = -5. + 10. * i / numEvents;
+      track.dir = {cos(phi), sin(phi), sinh(eta)};
```
agheata, line 126: Is this normalized anywhere?

examples/Example13/example13.cu
```diff
@@ -120,7 +120,13 @@ __global__ void InitPrimaries(ParticleGenerator generator, int startEvent, int n
     track.initialRange = -1.0;
 
     track.pos = {0, 0, 0};
-    track.dir = {1.0, 0, 0};
+    if (rotatingParticleGun) {
+      const double phi = 2. * 3.14159 * i / numEvents;
+      const double eta = -5. + 10. * i / numEvents;
```
agheata, line 125: This introduces a rather weird correlation eta-phi. How about a good old uniform theta/phi gun: 
```
phi = 2*M_PI*rnd();
theta= acos(1.-2.*rnd());
track.dir ={ sin(theta)*cos(phi), sin(theta)*sin(phi), cos(theta) };
```
This is normalized, de-correlated, and reproducible if we start the rng from the same seed.

examples/Example13/example13.cu
hageboeck: I was wondering if it has to be normalised. I interpreted this as a direction. Momentum should follow from the energy and the mass, so it shouldn't matter what the normalisation is. So tell me: Do I have to normalise (and why if you happen to know)?

examples/Example13/example13.cu
hageboeck: I preferred a weird correlation over shooting everything in positive x, but I'm willing to introduce more randomness. 🙂

hageboeck: I wouldn't go uniform in theta, though. Collisions are "uniform" in eta, no in theta. We would be skewing the probing of the geometry towards the centre, where we don't hit that often in the real world.

agheata: Sure, generating uniform eta is better

examples/Example13/example13.cu
agheata: Not normalizing direction generates havoc for geometry since all propagations are made like: `point +=  dir*distance` assuming dir is a unit vector. Also, if you need the momentum components (used in field propagation) they will be wrong. So yes, normalization has to be done.

hageboeck: Confirmed (and now fixed):
```
Normalised                                                    | Normalised * 5.

Simulating particles 1 ... done!				Simulating particles 1 ... done!
Run time: 112.829					      |	Run time: 99.9913


Mean energy deposit          8.13637 GeV		      |	Mean energy deposit          8.12393 GeV
Mean number of gamma         3506.79			      |	Mean number of gamma         3496.55
Mean number of e-            8167.55			      |	Mean number of e-            8205.67
Mean number of e+            359.98			      |	Mean number of e+            357.56
Mean number of charged steps 21586.8			      |	Mean number of charged steps 20658.6
Mean number of neutral steps 30942.1			      |	Mean number of neutral steps 27520.2
Mean number of hits          27120.3			      |	Mean number of hits          23151.3
Killed in propagation:       867			      |	Killed in propagation:       740
```

hageboeck:
@phsft-bot build

agheata (approved):

PR #161, opened 2021-12-06
Add AdePT logos

graeme-a-stewart:
As this seems to have become the defacto logo, here are the pdf, png and the original GIMP xcf files from which I sketched the logo


agheata (commented):
Thanks for adding these Graeme! I don't know which tool you used to make the logo, but it seems that some of the resolution was lost in the export process (
in particular visible for the text below AdePT). So would it be possible to export a hi-res pdf version to start with? We could then use this to produce few lower resolution ones, one with transparent background and maybe one for dark background presentations?

graeme-a-stewart:
Hi @agheata - it was done in GIMP, which is a raster based editor and unfortunately the orginal is also rather low resolution. I will work on it again as a vector image soon. I set this as a WIP for noe.

agheata:
By the way, we need to figure out how to disable the REUSE compliance check for binary files, maybe disable them by file extension?

graeme-a-stewart:
This is ready for another look now @agheata @WitekPokorski. There is a master SVG file, with the logo in vector format, plus exports of that to PDF and PNG.

The font has been updated to Ariel, mainly as it is so ubiquitous that there's no danger of font-rot because the fonts on someone's laptop evolve and things dissapear.

One slight design difference is that the lower text (Accelerated Particle Transport) is now in italics, giving a more dynamic feel.

graeme-a-stewart:
> By the way, we need to figure out how to disable the REUSE compliance check for binary files, maybe disable them by file extension?

The fix is to include a *sidecar* file for any binary `foo` called `foo.license`.

graeme-a-stewart:
AdePT-logo

agheata:
Very nice, sorry for the delay

agheata (approved):

PR #162, opened 2021-12-07
Use fewer threads when not enough particles

hahnjo:
*Note*: not necessarily a good idea since it wastes resources that other host threads might fill with their work, but makes `example13` 40% and `TestEm3` 5% faster in the current setup. This should not be merged until we can test with more than one thread on the CPU, once we have integration with Geant4.

Credits to @amadio for making me (re)think about using less threads, but the approach implemented here is beneficial overall since it only applies the strategy if there are few particles.

### DIFF
--- a/examples/Example13/example13.cu
+++ b/examples/Example13/example13.cu
@@ -289,9 +289,10 @@ void example13(int numParticles, double energy, int batch, const int *MCIndex_ho
     stats->inFlight[ParticleType::Positron] = 0;
     stats->inFlight[ParticleType::Gamma]    = 0;
 
-    constexpr int MaxBlocks        = 1024;
-    constexpr int TransportThreads = 32;
-    int transportBlocks;
+    constexpr int MaxBlocks              = 1024;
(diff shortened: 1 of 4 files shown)


amadio:
Although this change is fine for improving a bit the tails when batching a few particles at a time, using less threads is not actually what I am suggesting we do. Just batching more particles makes things 4x faster, which means that we should look into adapting the workflow to inject particles at regular intervals so they have staggered showers in terms of iteration cycles to actually not have these long tails where the GPU is almost entirely idle.

hahnjo:
As expected, using multiple threads for `TestEm3` (via `TestEm3MT`) reduces to negates the performance improvement at already 2-3 threads. For `cms2018`, now via `example14`, the benefit shrinks down to something in the order of 10-20% at four threads. The much larger benefit just comes from using multiple threads to submit work to the GPU, which is not actually over at four threads yet, so it's likely that both versions eventually approach about the same performance at higher thread counts.

PR #166, opened 2021-12-14
Japost/rk integration

jonapost:
Development of Runge-Kutta integration in 'kernel' code and example15 to demonstrate its use. 

New classes that implement the RK integration include:
- RkIntegrationDriver
- DormandPrince45
- MagneticFieldEquation
- fieldPropagatorRK

The development is not yet fully mature.

Runge-Kutta integration has been checked in the unit test vs helix for a uniform field.  The classes to integrate it into this example are created, and initial tests done.


hahnjo:
I tried to do some testing on my end, here are the results:

 * A full build of this PR / branch fails, as also reported by Jenkins:
```
In file included from /home/jhahnfel/AdePT/src-rk-integration/test/test_magfieldRK.cpp:14:
/home/jhahnfel/AdePT/src-rk-integration/magneticfield/inc/PrintFieldVectors.h: In instantiation of ‘void PrintFieldVectors::PrintSixvecAndDyDx(const Real_t*, int, const Real_t*, const Real_t*) [with Real_t = float]’:
/home/jhahnfel/AdePT/src-rk-integration/test/test_magfieldRK.cpp:86:40:   required from ‘bool TestEquation(const Field_t&) [with Real_t = float; Field_t = UniformMagneticField]’
/home/jhahnfel/AdePT/src-rk-integration/test/test_magfieldRK.cpp:457:83:   required from here
/home/jhahnfel/AdePT/src-rk-integration/magneticfield/inc/PrintFieldVectors.h:106:12: error: no matching function for call to ‘Print3vec(const float&, const char [25])’
   Print3vec( dydx[3], " dy/dx [3-5] (=dP/ds) = " );  printf("\n");
   ~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
 * `example15` runs (with quite verbose output), but the simulated results for 10 GeV electrons in CMS with B_z = 1T (all default values) look very wrong:
```
Mean energy deposit          0.247005 GeV
Mean number of gamma         90.23
Mean number of e-            301.51
Mean number of e+            4.96
Mean number of charged steps 1471.25
Mean number of neutral steps 10911.3
Mean number of hits          11318.4
Killed in propagation:       675
```
compare that to `example13` with also B_z = 1T enabled:
```
Mean energy deposit          9.87802 GeV
Mean number of gamma         3563.5
Mean number of e-            11807.4
Mean number of e+            309.65
Mean number of charged steps 29275.1
Mean number of neutral steps 32100.3
Mean number of hits          26898.8
Killed in propagation:       392
```
I tried increasing `max_iterations = 1000`, but not a huge difference.
 * The changes to the helix propagator driver to find more accurate intersections yield different results in TestEm3, but still not very good: In particular the first layers have a much lower energy deposit than `master` and Geant4, going as low as 2 per-mille relative error for the second layer. The later layers have a higher mean energy deposit (compared to `master` or my attempts to fix it), but still 0.05% below G4HepEm on the CPU. This is probably caused by less energy being deposited in the first few layers and therefore more energetic particles reaching the rest of the detector and "correcting" the underestimated energy deposit.

magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -11,21 +11,25 @@
...
-  __host__ __device__ void stepInField(double kinE, double mass, int charge, double step,
+  inline
+  VECCORE_ATT_HOST_DEVICE void stepInField(double kinE, double mass, int charge, double step,
                                        vecgeom::Vector3D<Precision> &position, vecgeom::Vector3D<Precision> &direction);
 
+
   template <class Navigator = LoopNavigator>
-  __host__ __device__ double ComputeStepAndNextVolume(double kinE, double mass, int charge, double physicsStep,
+  inline  
+  VECCORE_ATT_HOST_DEVICE double ComputeStepAndNextVolume(double kinE, double mass, int charge, double physicsStep,
```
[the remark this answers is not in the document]
jonapost: Correcting.  Thnkx

jonapost:
Thanks for the feedback. 
In the last day or two I am having difficulties using cuda-gdb on titanx to investigate.  This has slowed me down.
Unless it works I need to move to (or use in parallel) a different machine where I can use it.

magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -66,91 +71,173 @@ __host__ __device__ void fieldPropagatorConstBz::stepInField(double kinE, double
...
+        vecgeom::Vector3D<Precision> midPosition  = position;
+        vecgeom::Vector3D<Precision> midDirection = direction;
+        helixBz.DoStep<vecgeom::Vector3D<Precision>,Precision,int>(position, direction, charge, momentumMag, curveLen,
+                                                                   midPosition, midDirection);
+        // Instead we could create a method to re-use calculated helix parameters ...
+
+        // Also - should we just re-use endPosition / endDirection (in place of midPosition etc ) ??
+        
+        // Check
+        accurateIntersection = (midPosition - position).Mag2() < deltaIntersectSq;
```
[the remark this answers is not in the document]
jonapost: Yes, you are correct! Well spotted.  I will fix it.

hahnjo:
FWIW the very low energy deposit I reported in https://github.com/apt-sim/AdePT/pull/166#issuecomment-994936796 is still there.

jonapost:
I have fixed a key bug in the RK glue code for GPU.  I get almost perfect agreement between helix and Runge-Kutta now.
There is some cleanup to do, deleting the check code, before running further tests.

examples/Example15/electrons.cu
```diff
@@ -0,0 +1,486 @@
...
+    // also need to carry them over!
+
+    // Check if there's a volume boundary in between.
+    bool propagated = true;
+    double geometryStepLength;
+    vecgeom::NavStateIndex nextState;
+    
+    float BzFieldValue= *gPtrBzFieldValue_dev;   // Use vecgeom::Precision ?
+    if (BzFieldValue != 0.0) {
+      stepLengthFromPhysics= 0.1;
```
hahnjo, line 187: I suppose this is from debugging?

hahnjo (commented):
I gave this another try, and after fixing the build failure related to `kB2C` in `fieldPropagatorConstBz`, it's at least running. However, the energy deposit is still very wrong: With `stepLengthFromPhysics = 0.1`, I see around 0.5 GeV for a 10 GeV primary. With that line removed, I get only  3-4 MeV (!).

I also think the `CompareResponseVector3D` isn't suitable for comparing directions, which are / should be normalized so their magnitude is always (close to) 1.

jonapost:
I confirm that I have problems on steps which require multiple integration steps at least.
I am investigating.

I note that the function that compared  vectors checked the change from the previous to the next vector - not the magnitude of each vector. So it was not a complete check, but it was a valid check.

examples/Example15/electrons.cu
jonapost: Yes, of course.  I am making revisions - I had hoped to finish and push them already yesterday, but got sidelined.

jonapost:
I am still investigating the cause of the errors. They occur when multiple steps are needed to complete an integration. 

I pushed the code including debugging prints, so that it is in this repository.

hahnjo:
It appears this example is just generally broken, even without magnetic field: Both for `trackML.gdml` and `cms2018.gdml`, there is a energy deposit of only 5 MeV, no matter the energy set for the primaries (tested from 1 GeV to 100 GeV). For `testEm3.gdml`, the energy deposit seems about right (in agreement with `example13`).

agheata:
> It appears this example is just generally broken, even without magnetic field: Both for `trackML.gdml` and `cms2018.gdml`, there is a energy deposit of only 5 MeV, no matter the energy set for the primaries (tested from 1 GeV to 100 GeV). For `testEm3.gdml`, the energy deposit seems about right (in agreement with `example13`).

But wasn't this just copied from one of the previous examples replacing just the field propagator? @jonapost ?

magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -142,6 +149,21 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
         }
       }
 
+      static constexpr int ReduceIters  = 5;
+// #ifdef VERBOSE
+      if( safeLength < 1.0e-03 * physicsStep ) {
+         printf( "fpConstBz: very small safeMove = %10.5g  - vs physicsStep= %10.5g  \n", safeLength, physicsStep );
+         printf("%4s oneStep-Check track (id= %3d)  e_kin= %8.4g stepLen= %12.9g chord-iter= %5d\n ",
+             "Short", indx, kinE, safeLength, chordIters);
+      }
```
hahnjo, line 158: @jonapost can you remove these debugging printouts?

jonapost:
> 

In example15 I am getting energy deposition which is too perfect when I run with cms2018.gml - a 10 GeV primary deposits exactly 10 GeV, and a 5 GeV primary exactly 5 GeV.

hahnjo:
Ooohhhh, I think I see my mistake: All other recent examples default to `cms2018.gdml`, while `example15` uses `trackML.gdml`. That's a bit unfortunate because it is mostly filled with Air and only a bit of Silicon, so there is only very little physics happening. Can we change this back?

jonapost:
Yes, I agree and will change it back.  I had changed it for my testing, but it is simpler and better to have the same default.

jonapost:
I am working on the rebase.  It is a complicated because of the changes introduced in example13, from the use of the RNG to more local variables in place of data structures ...

agheata:
> I am working on the rebase. It is a complicated because of the changes introduced in example13, from the use of the RNG to more local variables in place of data structures ...

@jonapost any progress with the rebase?

agheata:
@jonapost looking at the modified code, I see that it is well-localized in `magneticfield/inc` and `example15`, example13 is barely touched. I can help with rebasing this branch, or if it turns out difficult we can make a new one that just adds the relevant files.

agheata:
Actually @jonapost I did the simple exercise and checked out your branch and rebase on the master, which gave no conflict. Then I could even make it compile by changing `fieldPropagatorConstBz.h` which seems to contain some debugging code compared to the master.
```
-    vecgeom::NavStateIndex &next_state, bool &propagated, const Precision safety,  int indx, // Slot index 
+    vecgeom::NavStateIndex &next_state, bool &propagated, Precision safety, int indx, // Slot index
```

jonapost:
I can run it now successfully with the CMS 2018 geometry for 100,000 initial particles.
So I think it's ready to merge.  Can you please clarify what's wrong, if anything?

agheata:
> I can run it now successfully with the CMS 2018 geometry for 100,000 initial particles. So I think it's ready to merge. Can you please clarify what's wrong, if anything?

Thanks @jonapost , do you have an estimate timing RK versus uniform field for CMS?

jonapost:
> The CI shows the example15 failing at the moment: https://lcgapp-services.cern.ch/spi-jenkins/job/AdePT-CI/510/console

Oops, I forgot to commit the change of Stack Limit to 8K.  Now done.

jonapost:
> > I can run it now successfully with the CMS 2018 geometry for 100,000 initial particles.  [....]
> 
> [..], do you have an estimate timing RK versus uniform field for CMS?

I am adding the numbers is overleaf:

For CMS-2018 and 10,000 particles the 
|                              | B= 1.0 T | B= 3.8 T  |  Example Name | 
|  :-------------- | -------: |  ---------:  | -----:       |
| helix                     |  255.4  |   571.4(5)  |   ex13   |
| helix/short steps |   231.0  |   580.5(9) |   ex18  |
| Runge-Kutta       |   413.0  |  1296 (1)    |   ex15  |

agheata:
@japost, I also measured that Exa15 is slower than Exa13 with a factor of 2 for smaller field values and a bit less for 3.8T. I think we can merge this, but we should be careful in quoting the results from this early version before we understand better the main bottlenecks and if we can easily remove some.

agheata (approved):

PR #167, opened 2021-12-16
Example14 - AdePT integration with Geant4

agheata:
This introduces an integration interface based on the GPU workflow adopted in Example13 and combined in a standard Geant4 CPU workflow:
* arbitrary geometry via gdml file (`/example14/detector/filename` command) and optionally a magnetic field with constant Bz (`/example14/detector/setField` command)
* physics list on the CPU, either HepEm or G4Em. HepEm physics on GPU.
* a GPU region declared by the user (`/example14/detector/regionname` command) where entering particles are killed in the Geant4 application and handed over to the GPU. Particles are buffered to a threshold for a given event, then the whole batch is showered on GPU. The particles escaping the GPU region are put back to the Geant4 stack.
* scoring per placed volume, based on a list of sensitive volumes declared by the user (`/example14/detector/addsensitivevolume` command. The hits produced on GPU by every batch are copied back to host and converted to Geant4 application hits (feature still being worked out)
* a first attempt of abstracting the user scoring, by using a scoring type providing some basic interface, and typedef-ing AdeptScoring to this type in a header file. The user can provide then alternative implementations and only change the typedef. 
* the GPU integration is done via a specific G4VFastSimulationModel that instantiates one AdeptIntegration object per CPU worker thread. The different AdeptIntegration share on GPU the geometry and physics data structures, but work with independent track containers and streams.
* The integration can be (de)activated using the command `/param/(De)ActivateModel model`
* The example requires a change in Geant4 main transportation loop to flush the fast sim model buffer, so this will compile but not run correctly without the change


agheata:
This is still WIP, it still has bugs in the mechanism of handing over tracks from GPU to host, but we've shared it so people can already have a look and comment. There are still a few areas to be cleaned up (related to hits, docs)

hahnjo:
I *tried* running `example14`, but it wasn't overly successful:
 * It seems that you need the patch for Geant4 to add `G4VFastSimulationModel::Flush`. Not that anything during the build would complain because the method is only marked `virtual ... final` and not `override`, but I guess otherwise particles are left in the buffer forever?
 * When trying to run regardless, I'm buried with tons of debugging output. Is there a way to turn this off / down to an acceptable level?
 * I'm not even sure `example14.in` runs to completion; I've let it run for 5 minutes, it used 15 GB of memory, and produced 3.6 GB of log output!

agheata:
It is expected to have this behaviour, I've made the PR even if the code is not yet in a working state so that comments on code can be made. The debugging output will be of course removed once the bug is fixed.

hahnjo (commented):
Okay, I didn't understand that the example is not expected to work at all. So all I can do for now is browsing a bit through the code and leave a few comments. In general, I think `examples/Par04` should be removed.

agheata:
> Andrei, did you put the Geant4 patch that I send you anywhere in Git? Maybe, for the time being we can simply put it in the example directory, so people can easily apply it.

I just did

examples/Example14/AdeptIntegration.cu
```diff
@@ -0,0 +1,501 @@
...
+                           const vecgeom::VPlacedVolume *world, AdeptScoring *userScoring)
+{
+  GlobalScoring *globalScoring = userScoring->fGlobalScoring_dev;
+  constexpr unsigned long kMaxTracks = 10000;
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < ntracks; i += blockDim.x * gridDim.x) {
+    ParticleGenerator *generator = nullptr;
+    switch (trackinfo[i].pdg) {
+    case 11:
+      generator = &secondaries.electrons;
+      atomicAdd(&globalScoring->numElectrons, 1);
```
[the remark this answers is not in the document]
agheata: Probably we should count only particles produced by AdePT transport, and count separately all tracks produced in Geant4, then we can add the 2 counts and compare them with Geant4 standalone. Anyway, I'll remove indeed the counting here.

examples/Example14/src/Example14DetectorConstruction.cc
```diff
@@ -0,0 +1,186 @@
...
+
+  // - UPDATE COUPLES
+  std::cout << "Updating material-cut couples based on " << G4RegionStore::GetInstance()->size() << " regions ...\n";
+  G4ProductionCutsTable *theCoupleTable = G4ProductionCutsTable::GetProductionCutsTable();
+  theCoupleTable->UpdateCoupleTable(world);
+
+  // --- Set MSC range factor to match G4HepEm physics lists.
+  G4EmParameters *param = G4EmParameters::Instance();
+  param->SetDefaults();
+  param->SetMscRangeFactor(0.06);
```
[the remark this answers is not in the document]
agheata: Indeed, this needs to be improved, but for the moment it is needed. We'll be adding a custom physics list, probably some of the code has to be moved there

examples/Example14/src/Example14EMShowerModel.cc
```diff
@@ -0,0 +1,145 @@
...
+{
+  // Remove particle from further processing by G4
+  aFastStep.KillPrimaryTrack();
+  aFastStep.SetPrimaryTrackPathLength(0.0);
+  G4double energy = aFastTrack.GetPrimaryTrack()->GetKineticEnergy();
+  // No need to create any deposit, it will be handled by this model (and
+  // G4FastSimHitMaker that will call the sensitive detector)
+  aFastStep.SetTotalEnergyDeposited(0);
+  auto particlePosition  = aFastTrack.GetPrimaryTrackLocalPosition();
+  auto particleDirection = aFastTrack.GetPrimaryTrackLocalDirection();
```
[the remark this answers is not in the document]
agheata: Indeed, to be fixed. Now it works just because the GPU region volume is has id transformation (at least in TestEm3)

agheata:
Screenshot 2022-01-14 at 10 32 13
Now we can compare scoring per placed volume in G4+AdePT vs. G4 standalone. There are big differences at this point per layer (up to 10%) pointing to remaining bugs. This is in magnetic field 0.1T.

agheata:
Fixing the random seed sequence makes the comparison with Geant4 look much better
fixed
.

hahnjo:
Thanks for the UI command to disable AdePT, now it's possible to test fully multi-threaded Geant4 and running on the GPU using the same binary, just using different macros. However, with a bit of testing on the CPU, I'm quickly getting crashes in `Example14SensitiveDetector::RetrieveAndSetupHit`. It looks a lot like memory corruption, sometimes segfaulting and some other times a  call to `free` failing with "invalid pointer"...

examples/Example14/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/Example14PrimaryGeneratorAction.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/Example14PrimaryGeneratorMessenger.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/Example14RunAction.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/Example14SensitiveDetector.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/PhysListHepEm.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/PhysListEmStandard.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/PhysListG4Em.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/PhysicsList.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/PhysicsListMessenger.cc
+  ${CMAKE_CURRENT_SOURCE_DIR}/src/TrackingAction.cc
```
hahnjo, line 36: No need to qualify this with `${CMAKE_CURRENT_SOURCE_DIR}`, paths are by default relative to the current source directory

examples/Example14/src/PhysicsList.cc
```diff
@@ -0,0 +1,291 @@
...
+  itr != (G4VMPLsubInstanceManager.offset[g4vmplInstanceID]).physicsVector->end();
+      ++itr)
+  {
+    (*itr)->ConstructProcess();
+  }
+}
+
+//....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
+
+void PhysicsList::AddPhysicsList(const G4String& name)
```
hahnjo, line 149: As said before, I don't think we need all this switching, we want G4HepEm on the host for comparison. However, I do think it would be critical to have an `FTFP_BERT` like physics list, including hadronic interactions, with only the EM physics for e-/e+/gamma replaced with G4HepEm or AdePT.

examples/Example14/src/Example14DetectorConstruction.cc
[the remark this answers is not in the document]
hahnjo: Why is the updating of regions and couples needed in the detector construction? It was needed when we weren't using the Geant4 kernel, but if it's still required there is an issue somewhere else - this should be taken care of automatically. And yes, the physics settings belong into the physics list.

examples/Example14/include/Example14EMShowerModel.hh
```diff
@@ -0,0 +1,112 @@
...
+  /// There are no kinematics constraints. True is returned.
+  virtual G4bool ModelTrigger(const G4FastTrack &) final;
+  /// Model is applicable to electrons, positrons, and photons.
+  virtual G4bool IsApplicable(const G4ParticleDefinition &) final;
+
+  /// Take particle out of the full simulation (kill it at the entrance
+  /// depositing all the energy). Simulate the full shower using AdePT library.
+  virtual void DoIt(const G4FastTrack &, G4FastStep &) final;
+
+  virtual void Flush() final;
```
hahnjo, line 71: Should be `override` instead of `virtual` - compilation should hard-error if the necessary interfaces aren't there!

agheata:
> However, with a bit of testing on the CPU, I'm quickly getting crashes in `Example14SensitiveDetector::RetrieveAndSetupHit`. It looks a lot like memory corruption, sometimes segfaulting and some other times a call to `free` failing with "invalid pointer"...

This is hopefully fixed with my last commit. For G4 scoring and finding the hit id in the example, we rely on a mapping between the runtime `G4VPhysicalVolume` pointer and the VecGeom placed volume id. This requires the AdePT interface to be instantiated and initialized for the common data (G4HepEm physics + auxiliary volume data), even if we don't allocate GPU storage for tracks and everything else. It seems to work in my simple test, but please let me know if I missed something.

examples/Example14/src/Example14EMShowerModel.cc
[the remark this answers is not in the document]
agheata: OK, fixed this.

agheata:
> Why is the updating of regions and couples needed in the detector construction?
I guess because we load region settings from the GDML, triggered by the detector construction, so these are intrinsically coupled with geometry. I'm not sure if the GDML parser does more than storing temporarily the auxiliary volume info and letting the user deal with it. Probably this can be done much better, but I guess we'll need to postpone some of the design problems of the integration after merging. I would try to merge this as soon as we are at least happy with correctness and we are able to run in all modes that we want to test, and postpone the non-critical fixes.

hahnjo:
> I guess because we load region settings from the GDML, triggered by the detector construction, so these are intrinsically coupled with geometry. I'm not sure if the GDML parser does more than storing temporarily the auxiliary volume info and letting the user deal with it. Probably this can be done much better, but I guess we'll need to postpone some of the design problems of the integration after merging. I would try to merge this as soon as we are at least happy with correctness and we are able to run in all modes that we want to test, and postpone the non-critical fixes.

Did you test this? Loading GDML also loads all regions and if return the `G4VPhysicalVolume` from `DetectorConstruction::Construct`, the Geant4 kernel will take care of the rest.

hahnjo:
> > However, with a bit of testing on the CPU, I'm quickly getting crashes in `Example14SensitiveDetector::RetrieveAndSetupHit`. It looks a lot like memory corruption, sometimes segfaulting and some other times a call to `free` failing with "invalid pointer"...
> 
> This is hopefully fixed with my last commit. For G4 scoring and finding the hit id in the example, we rely on a mapping between the runtime `G4VPhysicalVolume` pointer and the VecGeom placed volume id. This requires the AdePT interface to be instantiated and initialized for the common data (G4HepEm physics + auxiliary volume data), even if we don't allocate GPU storage for tracks and everything else. It seems to work in my simple test, but please let me know if I missed something.

Now it works *if* you call `/param/InActivateModel AdePT`. The provided `testEm3G4.in` doesn't do this right now and this fails during `Flush`.

hahnjo:
> > Did you test this? Loading GDML also loads all regions and if return the `G4VPhysicalVolume` from `DetectorConstruction::Construct`, the Geant4 kernel will take care of the rest.
> 
> No, what I tested is that removing this logic from the detector construction doesn't work any more. So unless you can point me to a quick fix, I propose postponing this.

Indeed, it crashes because G4HepEm doesn't find any material-cuts couples. The reason is that you're trying to initialize physics in `Example14DetectorConstruction::ConstructSDandField` which is before the Geant4 kernel sets everything up in `/run/initialize`. The problem is that this is wrong because users can change physics settings via the macro and call `/run/initialize` again. So the correct fix here is to build physics tables only when Geant4 tells us to do that, which may happen multiple times during execution.

agheata:
> So the correct fix here is to build physics tables only when Geant4 tells us to do that, which may happen multiple times during execution.

The current schema of bundling integration with the fast sim model and initializing at an inappropriate phase is clearly generating problems and will need to be eventually fixed, but the fix is non-trivial AFAICT and I proposed postponing it to be able to do tests with the integration. Any objection to that?

hahnjo:
> > So the correct fix here is to build physics tables only when Geant4 tells us to do that, which may happen multiple times during execution.
> 
> The current schema of bundling integration with the fast sim model and initializing at an inappropriate phase is clearly generating problems and will need to be eventually fixed, but the fix is non-trivial AFAICT and I proposed postponing it to be able to do tests with the integration. Any objection to that?

The important question here is: Can this be solved with the design centered around `G4VFastSimulationModel`? I don't see hooks that would allow us to properly build physics tables at the right time.

agheata:
> The important question here is: Can this be solved with the design centered around G4VFastSimulationModel? I don't see hooks that would allow us to properly build physics tables at the right time.
I think we diverge from what we try to do now. This is definitely not the final solution to be given to the user, and it has clear limitations, but it was started like this to demonstrate the functionality and to be able to do some performance measurements. So let's keep this discussion in mind, but move on to the current May goal.

agheata:
I've added the conditional compilation of this example only if the Geant4 `Flush` feature is detected. Now the macros and geometry are placed in a separate `macros` directory and installed in the build directory, and the example comes with 2 tests. I did also a bit of extra cleanup and removed the G4 patch files. I think this is ready for a final review.

agheata:
I think for merging we should squash.

hahnjo:
> I did also a bit of extra cleanup and removed the G4 patch files.

I think the patch files must be kept as long as there is no public version of Geant4 with the added interface.

agheata:
> > I did also a bit of extra cleanup and removed the G4 patch files.
> 
> I think the patch files must be kept as long as there is no public version of Geant4 with the added interface.

Right... I'll put them back in a `patches` folder.

examples/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ add_subdirectory(Example10)
 add_subdirectory(Example11)
 add_subdirectory(Example12)
 add_subdirectory(Example13)
+#add_subdirectory(Par04)
```
hahnjo, line 19: delete

examples/Example14/AdeptIntegration.cuh
```diff
@@ -0,0 +1,170 @@
...
+  cudaStream_t stream;                ///< all-particle sync stream
+  int fNumFromDevice{0};              ///< number of tracks in the fromDevice buffer
+  TrackData *toDevice_dev{nullptr};   ///< toDevice buffer of tracks
+  TrackData *fromDevice_dev{nullptr}; ///< fromDevice buffer of tracks
+  Stats *stats_dev{nullptr};          ///< statistics object pointer on device
+  Stats *stats{nullptr};              ///< statistics object pointer on host
+};
+
+// Constant data structures from G4HepEm accessed by the kernels.
+// (defined in TestEm3.cu)
```
hahnjo, line 159: `AdeptIntegration.cu`

examples/Example14/AdeptIntegration.cpp
```diff
@@ -0,0 +1,290 @@
...
+{
+  vecgeom::cxx::BVHManager::Init();
+  vecgeom::cxx::BVHManager::DeviceInit();
+}
+
+void AdeptIntegration::Cleanup()
+{
+  if (!fInit) return;
+  // if (!fCleanup.test_and_set()) {
+  //  std::cout << "=== Cleaning up GPU ...\n";
```
hahnjo, line 117: Is this intentionally commented out and `fCleanup` not used at all?

examples/Example14/src/Example14EventAction.cc
```diff
@@ -0,0 +1,113 @@
...
+  for(size_t iHit = 0; iHit < hitsCollection->entries(); iHit++)
+  {
+    hit     = static_cast<Example14Hit*>(hitsCollection->GetHit(iHit));
+    hitEn   = hit->GetEdep();
+
+    const char *vol_name = vecgeom::GeoManager::Instance().FindPlacedVolume(iHit)->GetLogicalVolume()->GetName();
+    const char *type = (hit->GetType()==1) ? "AdePT" : "G4";
+    if (hitEn)
+      std::cout << "EndOfEventAction " << eventId << " : id " << std::setw(5) << iHit
+                << "  edep " << std::setprecision(2) << std::setw(12) << std::fixed << hitEn/MeV << " [MeV] logical " << vol_name << std::endl;
```
hahnjo, line 105: This is still very verbose, in particular when running a decent number of events

hahnjo (commented):
All output should use `G4cout` instead of `std::cout` so that output isn't intermixed.

examples/Example14/src/Example14DetectorConstruction.cc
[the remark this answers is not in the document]
agheata: We agreed to solve design issues in a different PR

examples/Example14/AdeptIntegration.cpp
agheata: Indeed, not needed any more. Fixed.

examples/Example14/src/Example14EventAction.cc
agheata: Fixed now, verbosity for event actions can be changed via: /example14/event/verbose, default is 0, 1 prints just total energy deposit per event, >2 prints hits per placed volume

examples/Example14/src/PhysicsList.cc
agheata: This is done like this now.

examples/Example14/include/EMShowerModel.hh
```diff
@@ -0,0 +1,112 @@
...
+  ~EMShowerModel();
+
+  /// There are no kinematics constraints. True is returned.
+  virtual G4bool ModelTrigger(const G4FastTrack &) final;
+  /// Model is applicable to electrons, positrons, and photons.
+  virtual G4bool IsApplicable(const G4ParticleDefinition &) final;
+
+  /// Take particle out of the full simulation (kill it at the entrance
+  /// depositing all the energy). Simulate the full shower using AdePT library.
+  virtual void DoIt(const G4FastTrack &, G4FastStep &) final;
```
hahnjo, line 69: All of them should be `final override`, without the `virtual`

examples/Example14/AdeptIntegration.cpp
```diff
@@ -0,0 +1,280 @@
+// SPDX-FileCopyrightText: 2021 CERN
```
hahnjo, line 1: We've arrived in 2022. Some other files in this PR still have 2020, no sure where this was copied from (`example13` is correctly marked with 2021)...

hahnjo (approved):
Other than the two comments, I'm ok with merging this as-is (even though I didn't take a very close look after the last updates)

PR #168, opened 2022-02-03
Define `Geant4_flush_FOUND` if Flush exists in Geant4

peremato:
It tries to compile and link the following code:
```
  #include \"G4VFastSimulationModel.hh\"
  class testflush_ : public G4VFastSimulationModel {
    G4bool IsApplicable(const G4ParticleDefinition&) {return true;}
    G4bool ModelTrigger(const G4FastTrack &) {return true;}
    void DoIt(const G4FastTrack&, G4FastStep&) {}
    void Flush() final override {}
    public: testflush_() : G4VFastSimulationModel(\"\") {}
  };
  int main() { testflush_ model; return 0; }
```


peremato:
@agheata would be nice to squash the commits.

agheata:
> would be nice to squash the commits

I tried this locally but it seems to define Geant4_flush_FOUND even if I change the inlined test code so that it obviously cannot compile. I wonder if there was a way to see the output of the internal compilation command.

peremato:
> I tried this locally but it seems to define Geant4_flush_FOUND even if I change the inlined test code so that it obviously cannot compile. I wonder if there was a way to see the output of the internal compilation command.

Did you clean the build area? The result is cached.

agheata:
> I tried this locally but it seems to define Geant4_flush_FOUND even if I change the inlined test code so that it obviously cannot compile. I wonder if there was a way to see the output of the internal compilation command.

Actually I was wrong, it is just that unless one removed CMakeCache.txt, any changes in the inlined source file is ignored. One can actually check in build/CMakeFiles/CMakeError.log (in case of compilation failure) what the compiler does. So the patch is working nicely!

agheata (dismissed):
LGTM

agheata:
@peremato I don't understand why the build apparently failed although the console output shows that all tests passed. Is it the latest build that is shown here as failed?

hahnjo:
https://lcgapp-services.cern.ch/spi-jenkins/job/AdePT-CI/228/console shows that there were two compilation errors, details here: http://cdash.cern.ch/viewBuildError.php?buildid=143916

peremato:
Yes. Exactly!

agheata:
OK, so alpaka has likely some new dependency libboost-fiber-dev, which is somehow not available in the boost distribution. Is it enough to upgrade the boost bersion?

peremato:
Added an issue to SPI https://sft.its.cern.ch/jira/browse/SPI-2084

agheata:
> Added an issue to SPI https://sft.its.cern.ch/jira/browse/SPI-2084

Looks like Andre merged a patch: https://gitlab.cern.ch/sft/lcgcmake/-/merge_requests/1092

peremato:
@agheata I re-run the checks with the new version of the externals and works now.

agheata (approved):

PR #170, opened 2022-02-11
Reduce safeMove in case of no progress

### LINKED ISSUE #145: Tracks get stuck on boundaries in field propagation
In `TestEm3` with magnetic field and 10 GeV electrons, we see too short charged track lengths and lower energy deposit in most layers. Most importantly, the problem is worst for the middle layers with the most particles (so highest statistics) and highest energy deposit, that should agree the most with Geant4.

---

See https://github.com/apt-sim/AdePT/issues/145#issuecomment-1034784802 for a description of the underlying problem.

hahnjo:
If propagation in magnetic field doesn't make any progress, start reducing the `safeMove` to try smaller chord steps. Geant4 only does it after a number of zero-steps, but this would require a counter per track that we generally want to avoid.

Fixes #145

### DIFF
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -102,14 +102,16 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
     stepDone = Navigator::ComputeStepAndNextVolume(position, direction, remains, current_state, next_state, kPush);
     position += stepDone * direction;
   } else {
-    bool fullChord = false;
+    bool continueIteration = false;
+
+    Precision maxNextSafeMove = safeLength;
 
     //  Locate the intersection of the curved trajectory and the boundaries of the current
     //    volume (including daughters).
     do {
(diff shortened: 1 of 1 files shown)


hahnjo:
This finally gives correct results compared to Geant4:

fig_CaloEDep

It's also quite a bit faster (up to 30% in TestEm3) since we don't have to wait for the stuck particles, until they're eventually killed by the "looping" detection.

agheata:
Does this reduce the number of killed tracks, and by how much?

agheata:
I mean, can we lower the number of steps to kill threshold to reduce even more divergence while keeping the results correct?

hahnjo:
> Does this reduce the number of killed tracks, and by how much?

We don't have statistics about looping tracks since the printouts were deemed too verbose in the past (while actually pointing to a real problem). This does not affect tracks that reach the `max_iterations` limit.

> I mean, can we lower the number of steps to kill threshold to reduce even more divergence while keeping the results correct?

In a second step, yes, but as said yesterday morning we shouldn't kill them but let them progress a bit and try again in the next iteration. I already have code for that.

PR #172, opened 2022-02-13
Fix for field safe move estimation.

### LINKED ISSUE #171: Step length is wrongly computed in field propagation
As described in #145:
> PR https://github.com/apt-sim/AdePT/pull/95 introduced and used `momentumXYMag` as follows:
> https://github.com/apt-sim/AdePT/blob/fba5bd2373a67ffcf2478aec1d93a0d9e07171ba/magneticfield/inc/fieldPropagatorConstBz.h#L78-L91
> If there is no z-direction, ie `direction[2] = 0`, then `momentumXYMag = momentumMag`. Otherwise, and especially as `direction[2]` approaches 1 or -1, `momentumXYMag` will get smaller and smaller. This leads to an increasing value for `curv` and a decreasing value of `safeLength`. So in essence, the more a track is flying in z-direction and the less it is influenced by the constant field in z-direction, the shorter `safeLength` will get and the more chord steps magnetic field propagation will do.

Contrary to what I wrote back then, this is not only a problem for performance, but also correctness since we kill particles that cannot be propagated in `max_iterations` chord steps.

agheata:
Fixes #171

### DIFF
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -75,21 +75,20 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
 #else
   const Precision kPush = 0;
 #endif
-  Precision momentumMag = sqrt(kinE * (kinE + 2.0 * mass));
-  Precision momentumXYMag =
-      momentumMag * sqrt((1. - direction[2]) * (1. + direction[2])); // only XY component matters for the curvature
-
-  Precision curv = std::fabs(ConstBzFieldStepper::kB2C * charge * BzValue) / (momentumXYMag + 1.0e-30); // norm for step
-
+  // Maximum allowed error made by approximating step along helix with step along straight line
   constexpr Precision gEpsilonDeflect = 1.E-2 * copcore::units::cm;
 
-  // acceptable lateral error from field ~ related to delta_chord sagital distance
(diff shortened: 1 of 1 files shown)


magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -75,21 +75,20 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
...
-      momentumMag * sqrt((1. - direction[2]) * (1. + direction[2])); // only XY component matters for the curvature
-
-  Precision curv = std::fabs(ConstBzFieldStepper::kB2C * charge * BzValue) / (momentumXYMag + 1.0e-30); // norm for step
-
+  // Maximum allowed error made by approximating step along helix with step along straight line
   constexpr Precision gEpsilonDeflect = 1.E-2 * copcore::units::cm;
 
-  // acceptable lateral error from field ~ related to delta_chord sagital distance
+  Precision momentumMag = sqrt(kinE * (kinE + 2 * mass));
+  // Direction projection in plane perpendiculat to field vector
```
hahnjo, line 82: "perpendicular"

hahnjo (approved):
The code seems fine to me, at least corner cases should be handled without exploding. Please see https://github.com/apt-sim/AdePT/issues/171#issuecomment-1038816300 for a question on the mathematical derivation.

PR #173, opened 2022-02-16
Continue tracking of strongly curling particles

hahnjo:
Before, if a track did not arrive at its interaction point due to an error in field propagation, it was simply killed on the basis that it was likely a low-energetic electron. Instead keep it alive, apply the continuous energy loss up to that intermediate point and enqueue the particle for the next iteration step. In case it doesn't make progress, the track will eventually be killed by the general infrastructure that stops tracking if only electrons are left and no new secondary appeared for 200 steps.

This approach also opens the door for optimizations such as limiting the number of chord steps attempted in one tracking step, imposing a maximum step limit, or only operating within the geometrical safety sphere, if possible. However, experiments indicate that at least the first approach, limiting the number of attempted chord steps, is not effective or even counter-productive for more complex geometry setups such as cms2018, so it is not attempted here.

### DIFF
--- a/examples/Example10/electrons.cu
+++ b/examples/Example10/electrons.cu
@@ -82,17 +82,8 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
           currentTrack.energy, Mass, Charge, geometricalStepLengthFromPhysics, currentTrack.pos, currentTrack.dir,
           currentTrack.currentState, currentTrack.nextState, propagated);
 
-      if (!propagated) {
-        // error condition from field propagator. Just kill the track here and account for it explicitly.
-        atomicAdd(&scoring->killedInPropagation, 1);
-        // Particles are killed by not enqueuing them into the new activeQueue.
-        continue;
-      }
-
-      if (currentTrack.nextState.IsOnBoundary()) {
-        theTrack->SetGStepLength(geometryStepLength);
-        theTrack->SetOnBoundary(true);
-      }
+      theTrack->SetGStepLength(geometryStepLength);
+      theTrack->SetOnBoundary(currentTrack.nextState.IsOnBoundary());
 
       // Apply continuous effects.
       bool stopped = G4HepEmElectronManager::PerformContinuous(&g4HepEmData, &g4HepEmPars, &elTrack, nullptr);
@@ -153,6 +144,10 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
         // Cannot continue for now: either the particles left the world, or we
         // need to relocate it to the next volume.
         break;
+      } else if (!propagated) {
+        // Did not yet reach the interaction point due to error in the magnetic
+        // field propagation. Try again next time.
+        continue;
       } else if (winnerProcessIndex < 0) {
         // No discrete process, move on.
         continue;
--- a/examples/Example10/example10.cu
+++ b/examples/Example10/example10.cu
@@ -369,8 +369,6 @@ void example10(const vecgeom::cxx::VPlacedVolume *world, int numParticles, doubl
               << " energy deposition: " << std::setw(10) << stats->scoring.energyDeposit / copcore::units::GeV
               << " number of secondaries: " << std::setw(5) << stats->scoring.secondaries
               << " number of hits: " << std::setw(4) << stats->scoring.hits;
-    if (stats->scoring.killedInPropagation)
-      std::cout << " killed in propagation: " << std::setw(4) << stats->scoring.killedInPropagation;
     std::cout << std::endl;
 
     iterNo++;
--- a/examples/Example10/example10.cuh
+++ b/examples/Example10/example10.cuh
@@ -95,7 +95,6 @@ public:
 struct GlobalScoring {
   int hits;
   int secondaries;
-  int killedInPropagation;
   double energyDeposit;
 };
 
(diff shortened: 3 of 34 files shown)


hahnjo:
This appears to not cause much of a difference for `TestEm3`, both in terms of performance and observables (even though I didn't run a full validation yet). For `Example13` with `cms2018.no-scale.gdml` (to work around the known problems in VecGeom's GDML reader), simulation now takes three times as long. For the observables, I previously saw too few secondary particles compared to Geant4 (because killing was too aggressive) and now too many secondaries, with this change. I'd like to understand the difference here before I can confidently say that this "improves" the situation...

hahnjo:
`TestEm3` with `Bz=1T` continues to be fine (I ran with higher statistics yesterday) and `example13` with `Bz=1T` improves with this change, now very good agreement with Geant4. `Bz=3.8T` was off before (see above) and is still not right, likely a different cause. Unfortunately, I cannot really do runs with higher statistics because it's so slow (maybe the same problem, again stuck particles?). In any case, I think this one is good to merge.

agheata:
As far as I understand the code, physics gets re-sampled now in case of partial propagations (rather than resuming at the next step). This is probably fine, because of the 'ever-young' property, but is it efficient, in particular for a strong field, when a large fraction of particles may do only partial steps? In particular, we may call MSC much more frequently in this new scenario. How is the total number of steps affected by this change? About the precision increase, is this equivalent to increasing the number of chords cut in the previous approach?

hahnjo:
> As far as I understand the code, physics gets re-sampled now in case of partial propagations (rather than resuming at the next step). This is probably fine, because of the 'ever-young' property, but is it efficient, in particular for a strong field, when a large fraction of particles may do only partial steps?

Before a particle was simply killed if it didn't finish propagation. Now it is kept, which obviously means more work.

So I assume your question is about what happens if we decrease the `max_iteration` argument and attempt less chord steps within one tracking step? In that case, nothing is resampled for the `HowFar` method. The `number-of-interaction-left` is reduced according to the partial step as it was always done. What is different is that we have to interpolate the cross-sections more often, but there's no way around this because of the continuous energy loss that happens even for partial propagations.

> In particular, we may call MSC much more frequently in this new scenario.

Yes, and this is the costly part as mentioned on Tuesday.

> How is the total number of steps affected by this change?

Right now, without changing `max_iteration`, we get the additional steps of particles that were previously killed and their secondaries. For `TestEm3` and `Bz=1T`, this has a negligible effect, less than 0.01%. For `Example13` and `Bz=1T`, the number of charged steps increases by around 0.7% and the number of neutral steps by less than 0.1%.

> About the precision increase, is this equivalent to increasing the number of chords cut in the previous approach?

Likely, but I didn't test.

agheata:
I did some tests with this version, there is a clear improvement due to fewer killed tracks. The new version does not print the number of killed tracks, I think this is the minimum we need to add, if possible also their total energy. We should also understand the source of the factor of 3 slow-down at 3.8T, which for 1T is much smaller. After we merge this we may also try to record the current partial step for the non-propagated tracks, push them immediately to the next iteration, and only apply the continuous processes to fully propagated tracks, using the cumulated geometry step length. This may reduce the number of MSC calls to the same level we have in the current master.
```
rotatingParticleGun, B = 3.8 T
master:
Run time: 21.5418


Mean energy deposit          9.91628 GeV
Mean number of gamma         4130.55
Mean number of e-            10571.2
Mean number of e+            413.06
Mean number of charged steps 24180.2
Mean number of neutral steps 35029.7
Mean number of hits          27604.5
Killed in propagation:       1071

this branch:
Run time: 59.7667

Mean energy deposit          9.99987 GeV
Mean number of gamma         4154.53
Mean number of e-            10899.9
Mean number of e+            414.33
Mean number of charged steps 25722.6
Mean number of neutral steps 35470.2
Mean number of hits          28276.4
```

agheata:
> We decided back in October that we do not want to print detailed diagnostics about looping tracks, also because Geant4 doesn't do this. If you insist, I can add it in this PR but I'm not sure what changed in that regard...

We are trialling algorithms and we know we do errors because we kill particles systematically and these errors depend on the magnetic field and other parameters, so we better know when we get a given result how big these errors are. This was also the spirit of the previous `Killed in propagation` printouts that now will be gone. Protecting the printout with an input argument is also fine.
> 
> > After we merge this we may also try to record the current partial step for the non-propagated tracks, push them immediately to the next iteration, and only apply the continuous processes to fully propagated tracks, using the cumulated geometry step length.
> 
> After talking to Mihaly, this is indeed what we would _have_ to do because the conversion from a (shorter) geometrical step length of an intermediate step to the true step length is only an approximation. However, accumulating geometry steps is not as easy as it sounds because the MSC step limit depends on the safety of the starting position (which we could remember) and is randomized on boundary. So in essence, we would have to store all internal state of G4HepEm, which is currently kept locally in the `G4HepEmTrack` object) to global memory, which is quite heavy.
> 

OK I see, there is G4HepEm state set by the HowFar call, and to 'freeze' this for the partially propagated track requires caching this state...
> A different approach, that would also be "correct" from a physics perspective, is to impose a step limitation before calling into G4HepEm. Then MSC would do the geometry -> true step length conversion (which is accurate) of the step limit, and we can propagate up to the returned geometry step length. The step limit could be chosen such that field propagation will always stay below a certain number of chord steps.

This step limit could be estimated for the current track based on the helix parameters.
> 
> So, to summarize, the code in this PR _does not_ (yet) allow to decrease the number of chord steps while preserving physics. But it is definitely better to track particles and make errors than simply killing them, so this PR is already an improvement on its own.

Agreeing with that.

agheata (approved):

Issue #146, opened 2021-10-15
Optimize field propagation with safety


hahnjo:
Since the integration of Multiple Scattering (see https://github.com/apt-sim/AdePT/pull/142), we compute the safety for charged particle tracks which is needed for the true step length limit. It should be used in the magnetic field propagation to reduce geometry queries while still within safety.

PR #174, opened 2022-02-21
Fix navigation for zero steps

hahnjo:
If `step_limit <= 0`, the step is not limited by geometry and there is no need to ask any solid. In fact, some solids return NaN if passed a `step_max = 0`.


hahnjo:
This condition is extremely rare to hit, I only get it for two tracks when running `example13` with `Bz = 3.8T`. The reason I found this is that, working on #146, I expected identical results, but using the safety and skipping geometry queries doesn't run into the issue.

agheata (approved):

PR #175, opened 2022-02-21
Optimize field propagation with safety

### LINKED ISSUE #146: Optimize field propagation with safety

hahnjo:
Use the safety already available from MSC, and recompute if necessary to avoid querying geometry in multiple chord steps. In combination, the performance improvement for `example13` with `cms2018` and `Bz=1T` is around 27% and more than a factor of 2x for `Bz=3.8T`.

Closes #146

### DIFF
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -129,7 +129,17 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
       if (currentSafety > chordLen) {
         move = chordLen;
       } else {
-        move = Navigator::ComputeStepAndNextVolume(position, chordDir, chordLen, current_state, next_state, kPush);
+        Precision newSafety = 0;
+        if (stepDone > 0) {
+          newSafety = Navigator::ComputeSafety(position, current_state);
+        }
+        if (newSafety > chordLen) {
+          move         = chordLen;
+          safetyOrigin = position;
+          safety       = newSafety;
+        } else {
+          move = Navigator::ComputeStepAndNextVolume(position, chordDir, chordLen, current_state, next_state, kPush);
+        }
       }
 
       if (move == chordLen) {
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -67,7 +67,7 @@ template <class Navigator>
 __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
     double kinE, double mass, int charge, Precision physicsStep, vecgeom::Vector3D<vecgeom::Precision> &position,
     vecgeom::Vector3D<vecgeom::Precision> &direction, vecgeom::NavStateIndex const &current_state,
-    vecgeom::NavStateIndex &next_state, bool &propagated, const Precision /*safety*/, const int max_iterations)
+    vecgeom::NavStateIndex &next_state, bool &propagated, const Precision safety, const int max_iterations)
 {
   using Precision = vecgeom::Precision;
   #ifdef VECGEOM_FLOAT_PRECISION
@@ -103,6 +103,11 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
   } else {
     bool continueIteration = false;
 
+    Vector3D safetyOrigin = position;
+    // Prepare next_state in case we skip navigation inside the safety sphere.
+    current_state.CopyTo(&next_state);
+    next_state.SetBoundaryState(false);
+
     Precision maxNextSafeMove = safeLength;
 
     //  Locate the intersection of the curved trajectory and the boundaries of the current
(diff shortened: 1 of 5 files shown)


magneticfield/inc/fieldPropagatorConstBz.h
```diff
@@ -103,6 +103,11 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
   } else {
     bool continueIteration = false;
 
+    Vector3D safetyOrigin = position;
+    // Prepare next_state in case we skip navigation inside the safety sphere.
+    current_state.CopyTo(&next_state);
+    next_state.SetBoundaryState(false);
+
     Precision maxNextSafeMove = safeLength;
```
agheata, line 111: Can't we now use here: `Precision maxNextSafeMove = max(safeLength, safety);`?

hahnjo: In principle, maybe, but I exactly do *not* want to do that because it only works with the helix propagator and constant magnetic field.

agheata: Why is that? If safety is very large it does not matter how curly is the electron, and I don't see why you don't want to use that for the first step. And I don't see why we need to have the same algorithm supporting both helix and the RK.

hahnjo: > And I don't see why we need to have the same algorithm supporting both helix and the RK.

My understanding of AdePT's goals is to develop demonstrators to assess the performance that particle transport simulation could achieve on GPUs. For LHC experiments, we don't care how fast we can propagate charged particles in a constant magnetic field because that's not their setup. Thus, it doesn't make sense to optimize helix propagation to death using tricks that we know for sure don't apply for RK.

The additional benefit of using safety one level up, ie only to skip geometry queries, is that we get binary identical results (if we do things right). This makes it very easy to reason about correctness.

agheata: > My understanding of AdePT's goals is to develop demonstrators to assess the performance that particle transport simulation could achieve on GPUs. For LHC experiments, we don't care how fast we can propagate charged particles in a constant magnetic field because that's not their setup. Thus, it doesn't make sense to optimize helix propagation to death using tricks that we know for sure don't apply for RK.

Why do you say we cannot do larger integration steps with RK? What means "optimization to death", where is your threshold? What you apply here is optimization, if the helix is not needed by experiments,  why don't we leave the algorithm as it was? Getting binary identical results RK vs helix is a difficult target, having a good comparison to Geant4 is I think good enough. But all this starts going beyond the scope of the current PR, which does the job much better than before, so LGTM and discuss all this later, when we'll have the RK.

agheata (approved):

magneticfield/inc/fieldPropagatorConstBz.h
hahnjo: > What you apply here is optimization, if the helix is not needed by experiments, why don't we leave the algorithm as it was?

Because we still don't have RK integration in AdePT, and we have to start somewhere. My estimate is that, if helix propagation is already prohibitively slow, RK won't be better in terms of performance. But "cheating" our way out for helix propagation won't give us clues how to speed up RK.

> Getting binary identical results RK vs helix is a difficult target

I mean identical results before and after this optimization.

I just tried applying
```diff
diff --git a/magneticfield/inc/fieldPropagatorConstBz.h b/magneticfield/inc/fieldPropagatorConstBz.h
index d86b382..3065b84 100644
--- a/magneticfield/inc/fieldPropagatorConstBz.h
+++ b/magneticfield/inc/fieldPropagatorConstBz.h
@@ -115,7 +115,9 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
     do {
       Vector3D endPosition  = position;
       Vector3D endDirection = direction;
-      Precision safeMove    = min(remains, maxNextSafeMove);
+      Precision currentSafety = safety - (position - safetyOrigin).Length();
+      Precision safeMove    = min(remains, max(maxNextSafeMove, currentSafety));
+      // Precision safeMove    = min(remains, maxNextSafeMove);

       helixBz.DoStep<Vector3D, Precision, int>(position, direction, charge, momentumMag, safeMove, endPosition,
                                                endDirection);
@@ -124,7 +126,6 @@ __host__ __device__ Precision fieldPropagatorConstBz::ComputeStepAndNextVolume(
       Precision chordLen = chordVec.Length();
       Vector3D chordDir  = (1 / chordLen) * chordVec;

-      Precision currentSafety = safety - (position - safetyOrigin).Length();
       Precision move;
       if (currentSafety > chordLen) {
         move = chordLen;
```

and it is slightly faster (199s vs 187s for 1000 primary particles with `Bz=3.8T`), but produces less secondary electrons so the 5% improvement could also be due to statistics. Overall not worth it IMO.

PR #177, opened 2022-02-24
Add check for VecGeom being built as static library

sethrj:
@stognini discovered (and @drbenmorgan confirmed) that AdePT requires VecGeom to be built as static to link correctly. This PR adds a small configure-time check to ensure this is the case. If it fails, it prints a message:
```
CMake Error at CMakeLists.txt:65 (message):
  AdePT requires VecGeom to be built a static library, but it is of type
  'SHARED_LIBRARY'
```


drbenmorgan:
@phsft-bot build

sethrj:
Thanks @drbenmorgan !

agheata:
> @agheata, the Jenkins CI doesn't seem to have triggered. Is there a command/permissions I'm missing to activate this?

No, it should trigger automatically

PR #182, opened 2022-03-09
adding LHCb geometry

WitekPokorski:
Added LHCb geometry and a macro file with LHCb sensitive detector. Also added the possibility of some angular smearing in the particle gun.

### DIFF
--- a/examples/Example14/src/PrimaryGeneratorAction.cc
+++ b/examples/Example14/src/PrimaryGeneratorAction.cc
@@ -12,7 +12,7 @@
 //....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
 
 PrimaryGeneratorAction::PrimaryGeneratorAction(DetectorConstruction *det)
-    : G4VUserPrimaryGeneratorAction(), fParticleGun(0), fDetector(det), fRndmBeam(0.), fGunMessenger(0)
+    : G4VUserPrimaryGeneratorAction(), fParticleGun(0), fDetector(det), fRndmBeam(0.), fRndmDirection(0.), fGunMessenger(0)
 {
   G4int n_particle = 1;
   fParticleGun     = new G4ParticleGun(n_particle);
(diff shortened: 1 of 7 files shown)


examples/Example14/src/PrimaryGeneratorAction.cc
```diff
@@ -36,6 +36,17 @@ void PrimaryGeneratorAction::GeneratePrimaries(G4Event *aEvent)
...
+  G4ThreeVector oldDirection = fParticleGun->GetParticleMomentumDirection();
+  // randomize direction if requested
+  if(fRndmDirection > 0.) {
+   
+    G4double rdir            = 0.5 * fRndmDirection;
+    G4double dx = oldDirection.x() + (2 * G4UniformRand() - 1.) * rdir;
+    G4double dy = oldDirection.y() + (2 * G4UniformRand() - 1.) * rdir;
+    G4double dz = oldDirection.z();
+    fParticleGun->SetParticleMomentumDirection(G4ThreeVector(dx,dy,dz));
+  }
```
agheata, line 49: This looks quite odd as direction randomization keeping the Z components and randomizing X and Y.  Also the input in double precision as magnitude of the random on top of the old direction is weird. Why not using a standard eta-phi gun as in [example13](https://github.com/apt-sim/AdePT/blob/a50453948843ea19a17a6029899d00fa856cbab3/examples/Example13/example13.cu#L125)?

agheata, line 49: This looks quite odd as direction randomization keeping the Z components and randomizing X and Y.  Also the input in double precision as magnitude of the random on top of the old direction is weird. Why not using a standard eta-phi gun as in [example13](https://github.com/apt-sim/AdePT/blob/a50453948843ea19a17a6029899d00fa856cbab3/examples/Example13/example13.cu#L125)?

WitekPokorski: Indeed, it was a bit of a nonsense. Now is replaced by proper smearing in eta and phi.

agheata:
The macro lhcb.mac.in and new gdml geometry files are currently not copied in the bin directory, as the others files in the `macros` directory, which is convenient since it produces a macro that can be run w/o copying files. Please check how this is done for testEm3 geometry and macro in the CMakeLists.txt

WitekPokorski:
added now

agheata:
There is still an annoyingly long list of names printed by the example, but it is not the example itself to blame but the G4GDMLRead during the solids section.

agheata (approved):

agheata:
The GDML files are missing the license header, you can add it as for teastEM3.gdml in the same folder

WitekPokorski:
The long list of names comes from a forgotten std::cout left in G4GDMLReadSolids::PropertyRead. I will fix it in the next Geant4 patch.

PR #183, opened 2022-03-23
Fix for assigning sensitive reflected volumes in Example14

agheata:
In addition, this adds a thorough check for local transformation consistency, comparing each local transformation while crawling in parallel through the Geant4 and VecGeom volume hierarchies. The check is done for both translation  and rotation elements, with a tolerance of 1e-8;

This can only be merged after merging VecGeom [#884](https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/884) and bumping the VecGeom tag (to be added in this PR)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -50,7 +50,7 @@ find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
 
 # Find VecGeom geometry headers library
-set(VecGeom_VERSION 1.1.17)
+set(VecGeom_VERSION 1.1.20)
 find_package(VecGeom ${VecGeom_VERSION} REQUIRED)
 message(STATUS "Using VecGeom version ${VecGeom_VERSION}")
 # make sure we import VecGeom architecture flags - is this needed?
--- a/README.md
+++ b/README.md
@@ -15,7 +15,7 @@ The following packages are a required to build and run:
 - C/C++ Compiler with C++14 support
 - CUDA Toolkit (tested 10.1, min version TBD)
 - VecCore [library](https://github.com/root-project/veccore) 0.7.0 (recommended, but older versions >= 0.5.0 also work)
-- VecGeom [library](https://gitlab.cern.ch/VecGeom/VecGeom) >= 1.1.17
+- VecGeom [library](https://gitlab.cern.ch/VecGeom/VecGeom) >= 1.1.20
 
 To configure and build VecCore, simply run:
 ```console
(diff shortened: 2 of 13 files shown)


CMakeLists.txt
```diff
@@ -50,7 +50,7 @@ find_package(VecCore ${VecCore_VERSION} REQUIRED COMPONENTS ${VecCore_BACKEND})
 message(STATUS "Using VecCore version ${VecCore_VERSION}")
 
 # Find VecGeom geometry headers library
-set(VecGeom_VERSION 1.1.17)
+set(VecGeom_VERSION 1.1.20)
```
hahnjo, line 53: Should also update `README.md` to not lack behind.

hahnjo:
@agheata can we please keep the watermark changes out of this PR? As discussed, this poses significant questions for reproducibility, and to be honest it's unrelated to the reflected solids

agheata:
Besides the sensitive volumes fix, I have now added a couple of new features:
- better memory management: the track slots limit can be set per device using the command: `/example14/adept/milliontrackslots [8]`. This will evenly divide the slots among the used CPU threads, allowing to increase the buffer size and consequently get better performance
- the maximum number of tracks in flight on GPU can be now set using the command: `/example14/adept/maxinflight [N]`. This is supposed to avoid saturation of the GPU when using many CPU threads. So far I observe that even if the number of tracks gets better under control, there is no observable benefit. The feature has a large default value which preserves the old behaviour.

I could raise the buffer threshold in the CMS example to `/example14/adept/threshold 300` compared to the previous value of 50. Using 24 CPU threads (12 physical cores) I'm seeing a 50% speedup compared to Geant4 (compiled in release mode now...). It would be good if this could be confirmed independently.

hahnjo:
As said above, I disagree with putting everything in one PR.

hahnjo:
> I could raise the buffer threshold in the CMS example to `/example14/adept/threshold 300` compared to the previous value of 50. Using 24 CPU threads (12 physical cores) I'm seeing a 50% speedup compared to Geant4 (compiled in release mode now...). It would be good if this could be confirmed independently.

I cannot reproduce this, at least in the setups that I tested: 10 GeV electrons (with 24 threads and `/example14/adept/threshold 300`) are still factors slower, 100 GeV electrons are better but still at least 60% slower in the two configurations I'm able to run (8 threads with `/example14/adept/threshold 300`, 24 threads with `/example14/adept/threshold 20`). The baseline are 24 threads on the CPU, with still a pretty unoptimized version of Geant4 (no static libraries, no tracking manager for G4HepEm, no Vecgeom on the CPU).

agheata:
Probably the fact that I'm running locally on a Haswell E5-2630, which became rather old and slow makes the difference.

agheata:
@hahnjo I left in only the memory usage improvement. The build finally passes, can you have a quick look?

hahnjo (approved):
Setting the capacity globally and dividing it by the number of threads has the disadvantage that you also need to adapt the `threshold` when changing the number of threads. But it's actually not worse, before you had to change the capacity depending on the number of threads...

PR #185, opened 2022-04-01
Hepmc

WitekPokorski:
Adding the interface to HepMC3 ascii reader. One can chose now between running with the particle gun or with the HepMC data file.


agheata (changes requested):
As a general comment, we should try to make the feature optional, activating only if found. We need to add the copyright and licence to all new files, and preferably they should get clang-formated. I would put the data file in a separate directory `data` rather than in `macros`. Speaking of that, I noticed that the LHCb geometry files and HepMC3 file have together about 70MB, would it make sense to download them if not present from some location?

agheata:
In case HepMC3 is not found, there is still a compilation error:
```
/home/agheata/geant_src/AdePT/examples/Example14/include/HepMC3G4Interface.hh:14:10: fatal error: HepMC3/GenEvent.h: No such file or directory
 #include "HepMC3/GenEvent.h"
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
```

agheata:
If the data file is not found, currently the simulation continues with 0 hits at the end. Probably better if the messenger checks the existence of the file and aborts the simulation.

agheata:
The data files need also license info, which can be added in the same folder as: filename.license (containing the common license header), see an example in the /logo folder

WitekPokorski:
fixed the CMakeLists.txt file

On 6 Apr 2022, at 14:37, Andrei Gheata ***@***.******@***.***>> wrote:



In case HepMC3 is not found, there is still a compilation error:

/home/agheata/geant_src/AdePT/examples/Example14/include/HepMC3G4Interface.hh:14:10: fatal error: HepMC3/GenEvent.h: No such file or directory
 #include "HepMC3/GenEvent.h"
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.

WitekPokorski:
Actually, the simulation is aborted and it says so.
The 0 hits come from the end of run action which is still called (and displays the counter of hits to be 0).
I think, we can leave it for the time being.

On 6 Apr 2022, at 14:39, Andrei Gheata ***@***.******@***.***>> wrote:



If the data file is not found, currently the simulation continues with 0 hits at the end. Probably better if the messenger checks the existence of the file and aborts the simulation.

agheata:
Checks are still failing because:
- the branch needs to be rebased on master
- the extension for the license files is misspelled (.licence) 
Also, we should use 2022 instead of 2021 in all files

agheata:
> Actually, the simulation is aborted and it says so. The 0 hits come from the end of run action which is still called (and displays the counter of hits to be 0). I think, we can leave it for the time being.

OK, I see, I was mislead by the hits printing.

agheata (approved):

PR #188, opened 2022-04-05
Link Raytrace benchmark against vecgeomcuda_static.

hageboeck:
If an executable uses vecgeom device symbols, i.e. does more than
exclusively deal with vecgeom's c++ interface, it needs to link
statically.
This fixes a failing kernel launch in RaytraceBenchmark.

### DIFF
--- a/examples/Raytracer_Benchmark/CMakeLists.txt
+++ b/examples/Raytracer_Benchmark/CMakeLists.txt
@@ -45,7 +45,7 @@ target_include_directories(RaytraceBenchmark PUBLIC
   $<BUILD_INTERFACE:${CMAKE_SOURCE_DIR}/base/inc>
   $<INSTALL_INTERFACE:base>
 )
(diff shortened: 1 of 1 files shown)


agheata (approved):

drbenmorgan:
To quickly confirm the cause of this error - it is introduced in VecGeom 1.1.19. The ultimate cause is that the move to native CMake support no longer automatically links the shared `vecgeomcuda` target to the shared CUDA Runtime Library. Even though the end application links to the static version of this, it looks like this doesn't/can't replace the shared version (it's not clear if the linking of _both_ the static and shared versions of the CUDA Runtime into the `RaytraceBenchmark` is o.k. in every use case. 

So the fix by @hageboeck is correct (consistent use of static libs), though we should look at a patch to VecGeom if needed. I want to confirm the requirements/etc for use/linking/mixing of the static and shared CUDA Runtime Libraries to make sure we handle this correctly, or document the correct linking requirements for VecGeom.

PR #189, opened 2022-04-06
Fix calls to template functions in G4HepEm

bernhardmgruber:
This PR prepares AdePT for upcoming changes to G4HepEM where some functions will turn into templates.


hahnjo:
see discussion in https://github.com/apt-sim/AdePT/pull/192

PR #190, opened 2022-04-06
Update G4HepEM

bernhardmgruber:
This PR updates the G4HepEM git submodule to its current master.


bernhardmgruber:
@hahnjo just DMed me that the G4HepEM update would

> introduce energy loss fluctuations which are horribly slow on the GPU, so we currently don't want that. I'll work on a method to disable that in G4HepEm tomorrow.

I will wait with this PR then.

bernhardmgruber:
Great, thx!

PR #192, opened 2022-04-06
Avoid G4HepEmRandomEngine

bernhardmgruber:
TODO

This PR depends on a few patches currently upstreamed to G4HepEM. In the end, it should give a speedup of around 1.7 on TestEm3.


hahnjo:
I gave this set of pull requests a try, in particular with respect to performance. The baseline is 28.3 seconds for 10k electrons à 10 GeV in `TestEm3`, built with current `master` of both AdePT and VecGeom. Please keep in mind that I tested on my RTX 2070 SUPER, which is a consumer card. Numbers might be different for other hardware (but I'd be surprised if the general trend is radically different).

First I tested the changes to VecGeom from the merge requests !897, !900, and !901 together with the current `master` of AdePT - these patches should be independent, right? On my hardware, I see the runtime for 10k electrons dropping to 26.8 seconds, an improvement of about 5%. Very nice :partying_face: it might be worth testing the individual changes to understand what is the biggest help to the compiler and potentially look for other opportunities.

Then I merged https://github.com/mnovak42/g4hepem/pull/75 and https://github.com/mnovak42/g4hepem/pull/76 to `external/g4hepem`, but reverted the addition of energy loss fluctuations. As mentioned in #190 I'll work on a "proper" method for this tomorrow (likely similar to `NOMSC` that we already have right now). With no other changes to AdePT, I see identical performance which matches the expectation since the compiler sees more or less the same code.

Finally I apply the RNG adaptor to `TestEm3` in AdePT (the second commit in this PR), which, as far as I understand, is expected to improve performance. My measurements cannot confirm this, if anything I see slightly longer runtimes but this could also be random jitter. Afterwards, increasing `TransportThreads` to 256 is definitely slower with around 31.3 seconds.

The initial version of this PR also increased `Capacity` 16-fold, to `4 * 1024 * 1024`. This indeed shows a factor of 1.7x speedup compared to the baseline. However the same can be seen by applying only this single modification to `master`. Moreover this advantage mostly vanishes when using 3-4 threads on the CPU to submit work to the GPU.

bernhardmgruber:
Register footprint change from before (but with VecGeom !901)/after this PR:
image

hahnjo:
Some more numbers of actual runtime (reducing register usage is irrelevant as long as it doesn't improve performance):
```
TestEm3.master.10000.txt:Run time: 18.1538
TestEm3.master.mem12.10000.txt:Run time: 16.3318
TestEm3.master.rng.10000.txt:Run time: 18.3598
TestEm3.master.rng.mem12.10000.txt:Run time: 16.526
TestEm3.reg128.10000.txt:Run time: 18.2864
TestEm3.reg128.mem12.10000.txt:Run time: 15.6588
TestEm3.reg128.rng.10000.txt:Run time: 18.6395
TestEm3.reg128.rng.mem12.10000.txt:Run time: 16.227
```
(`TestEm3` with `-particles 10000`; `master` is with the default register count, `reg128` is when compiling both VecGeom and AdePT with `--maxrregcount 128`; `mem12` is when increasing the Capacity 12-fold (compared to current `master` where it was already bumped by a factor of 4 since this PR was posted), and `rng` are the two PRs to G4HepEm and the one in AdePT merged locally)

One more experiment when playing with `TransportThreads` (number of threads per block, `t256` for 256 threads) and `MaxBlocks` (`b32` for increasing by a factor 32):
```
TestEm3.reg128.mem12.10000.txt:Run time: 15.6588
TestEm3.reg128.mem12.b32.t256.10000.txt:Run time: 13.5977
TestEm3.reg128.mem12.t256.10000.txt:Run time: 14.0769
TestEm3.reg128.rng.10000.txt:Run time: 18.6395
TestEm3.reg128.rng.mem12.b32.t256.10000.txt:Run time: 14.0575
TestEm3.reg128.rng.mem12.t256.10000.txt:Run time: 14.4771
```

PR #194, opened 2022-04-12
Testem3 compact

agheata:
This introduces a new TrackManager<Track> helper able to reuse track memory slots. It also demonstrates its use in a new example TestEm3Compact. The implementation uses a circular buffer of pre-allocated tracks, an atomic pointing to the next free slot and two MParray with active and next track indices. After each stepping iteration, the helper estimates the number of slots needed to compact the tracks in-flight and launches a compacting kernel when the needed space is larger than a user-defined fraction of the total size. The compression for a buffer of 1M track slots when reaching 90% occupancy is about x25 for TestEm3 with 10GeV electrons. This is of course highly dependent on the setup.

TrackManager

This utility introduces an overhead compared to the existing approach that I measured to ~3% in case the buffer is large enough so no defragmentation is needed. The overhead of the defragmentation itself is negligible within few measurements error bars.
The performance scaling with increasing the batch size can be measured up to much larger values, as shown below for `testem3-compact`.

scaling-testem3-compact

The results are numerically identical to those of TestEm3, independent of the batch size.


agheata:
Anyone knows how to cleanup the CI machine?
```
hudson.plugins.git.GitException: Command "git checkout -f 0b23b798885d348f1f6f3019819c4f652149b7b8" returned status code 128:
stdout: 
stderr: error: unable to write file examples/Example14/data/ppminbias.hepmc3
error: unable to write file examples/Example14/data/ppttbar.hepmc3
error: unable to write file examples/Example14/macros/LHCb_Upgrade_fullLHCb.gdml
error: unable to write file examples/Example14/macros/LHCb_Upgrade_onlyECALandHCAL.gdml
error: unable to create file examples/Example14/patches/geant411_adept.patch (No space left on device)
error: unable to create file examples/Example14/patches/geant411_adept.patch.license (No space left on device)
```

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+template <typename Track>
+struct TrackManager {
+
+  /// @brief Statistics of the track manager to be updated and copied to host between iterations
+  struct Stats {
+    int fStart{0};     ///< Index of first used track in the buffer
+    int fNextStart{0}; ///< Index where the tracks will be compacted
+    int fInFlight{0};  ///< Number of tracks sill in flight
+
+    int GetNused() { return fNextStart - fStart; }
```
hahnjo, line 95: Does this work / need to work if the buffer wraps around? I'd expect some `+capacity` and `%capacity` going on here...

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+  adept::MParray *fActiveTracks{nullptr}; ///< Array of active (input) track slots (device pointer)
+  adept::MParray *fNextTracks{nullptr};   ///< Array of rack slots for the next iteration (device pointer)
+  Track *fBuffer{nullptr};                ///< Storage for the circular buffer of tracks (device pointer)
+
+  /// @brief Construction done on host but holding device pointers.
+  __host__ __device__ TrackManager(size_t capacity) : fCapacity(capacity)
+  {
+    // The Track type must be copy constructible
+    if (!std::is_copy_constructible<Track>::value)
+      COPCORE_EXCEPTION("TrackManager ctor: The track type must be copy constructible");
```
hahnjo, line 112: I think you want this as a static check (probably as part of the `template`), not during runtime.

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+
+/**
+ * @file TrackManager.h
+ * @brief A track manager using a circular buffer with slot compression capability.
+ * @details The track manager is templated by the track type and holds a circular
+ *          buffer of pre-allocated tracks. It also holds two MParray with track
+ *          slots for the current and next iteration.
+ *
+ *          The compression of the sparse active slots is triggered when the sum of the
+ *          number of used slots and 3 times the number of inflight tracks becomes larger
```
hahnjo, line 12: should be "2 times the number of inflight tracks" (correctly done in the code below)

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+    int inFlight = fStats.fInFlight;
+    assert(used >= 0 && used < fCapacity);
+    // Estimate maximum space needed if we DON'T compress now
+    int needed = used + 2 * inFlight;
+    if (needed < compact_threshold * fCapacity || needed > fCapacity) {
+      device_impl_trackmgr::swap_active<Track><<<1, 1, 0, stream>>>(fInstance_d);
+      if (needed > fCapacity) {
+        std::cout << "TrackManager::SwapAndCompact  ALERT: not enough space left to compress " << used
+                  << " used slots. Consider increasing TrackManager capacity.\n";
+      }
```
hahnjo, line 168: Not sure this check is correct: `needed` is the space we need if we don't compress. For compressing, we actually only need an additional `inflight` slots, right?

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+
+  /// @brief Construction done on host but holding device pointers.
+  __host__ __device__ TrackManager(size_t capacity) : fCapacity(capacity)
+  {
+    // The Track type must be copy constructible
+    if (!std::is_copy_constructible<Track>::value)
+      COPCORE_EXCEPTION("TrackManager ctor: The track type must be copy constructible");
+    // The queue size must be a power of 2 (for fast access in the circular buffer)
+    bool pow_two = (capacity & (capacity - 1)) == 0;
+    if (!pow_two) COPCORE_EXCEPTION("TrackManager ctor: The buffer size has to be a power of 2");
```
hahnjo, line 115: Is this really needed? Just do a modulo operation instead.

base/inc/AdePT/TrackManager.h
```diff
@@ -0,0 +1,263 @@
...
+  /// @brief Host static call to clear the container.
+  template <typename Stream>
+  void Clear(Stream stream)
+  {
+    fStats.fStart     = 0;
+    fStats.fNextStart = 0;
+    fStats.fInFlight  = 0;
+#ifdef COPCORE_CUDA_COMPILER
+    device_impl_trackmgr::clear_trackmanager<Track><<<1, 1, 0, stream>>>(fInstance_d);
+#endif
```
hahnjo, line 200: Hrm, this is is *really* bad because the compiler will generate different code depending on where this is called and then the linker gets to decide which version to choose. We already had this problem in the past, please don't get us back to that...

base/inc/AdePT/TrackManager.h
agheata: Yes, fNextStart is always larger than fStart, it may exceed the buffer size. Slot numbers are systematically adjusted to modulo fCapacity

base/inc/AdePT/TrackManager.h
agheata: Indeed, fixed in the upcoming version.

base/inc/AdePT/TrackManager.h
agheata: Fixed

base/inc/AdePT/TrackManager.h
agheata: Yes, it is, `needed` takes into account also the space taken by the uncompressed tracks in the scenario where the compression will not be done.

base/inc/AdePT/TrackManager.h
agheata: Getting slots is called very frequently, which needs to do systematically a modulo, and modulo is expensive, so I prefer this for the time being. We can hide it behind a macro to make code more readable, maybe in the next fixup.

base/inc/AdePT/TrackManager.h
agheata: Yes, I initially thought using the utility host and device, but dropped this. I'll make the header `.cuh` removing the unnecessary ifdefs

base/inc/AdePT/TrackManager.h
hahnjo: > modulo is expensive

I don't agree on that, do you have numbers?

base/inc/AdePT/TrackManager.h
hahnjo: I understand, but we still only need `used + inFlight`, not twice `inFlight`.

base/inc/AdePT/TrackManager.h
agheata: Indeed, the difference is at per-mil level, not justifying limiting the buffer size to be a power of 2. Fixed that.

base/inc/AdePT/TrackManager.h
agheata: The worst-case scenario is that each particle `inFlight` will produce 2 extra particles (e.g. annihilation). This would require in total `2 * inFlight` extra slots. So to survive the next iteration conservatively, the buffer must have at least: `used + 2 * inflight` slots (because the current `inFlight` slots are already embedded in the `used` ones)

agheata: Indeed I revisited the "not enough space to compress" condition, should be OK in the last version

hahnjo (commented):
The data structure looks good to me, only a couple of comments for the new example.

hahnjo (approved):
Looks good. Please squash before / while merging.

PR #195, opened 2022-04-13
Making use of TrackManager in example13.

agheata:
This should be merged after #194, it allows running with larger batch size
example13_bs
.


hahnjo:
Hm, I'd actually prefer to have this in a new example so that we can compare. Can you move the changes to a new `Example16` (skipping number 15 which John is working on)?

PR #196, opened 2022-04-13
New example16 making use of TrackManager in example13.

agheata:
This should be merged after https://github.com/apt-sim/AdePT/pull/194, it allows running with a larger batch size.

scaling_exa16


hahnjo (approved):
I didn't (re)check `example16` in depth, but went over the diff in https://github.com/apt-sim/AdePT/pull/195 :+1:

PR #197, opened 2022-04-14
Adjust buffer capacity and launch configuration for TestEm3 based on device properties

amadio:


examples/TestEm3/TestEm3.cu
```diff
@@ -193,10 +193,19 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double
...
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, 0);
+
   // Capacity of the different containers aka the maximum number of particles.
-  constexpr int Capacity = 1024 * 1024;
+  // Use 25% of GPU memory for each of e+/e-/gammas, leaving 25% for the rest.
+  int Capacity = deviceProp.totalGlobalMem / sizeof(Track) / 4;
+
+  int MaxRegistersPerThread = 256;
+  int ThreadsPerBlock = deviceProp.regsPerBlock / MaxRegistersPerThread;
```
hahnjo, line 204: This is actually not a good default, spending all available registers on one block. Normally you want to schedule multiple blocks per SM. In this case, the previous default of 32 threads is faster (at least on my machine).

hahnjo: > You cannot force more blocks into the SM just by changing the `ThreadsPerBlock` here.

Sure you can: Instead of 1 block with 256 threads, I can run 8 blocks à 32 threads. And this is beneficial if blocks take different time, as we have it.

> For small batches, it does make the code slower, but for larger batches (which we should aim for, to give the GPU enough to work on), it makes it better.

No, it doesn't: As I wrote, it's slower for the "new" default batch size proposed by this PR. Which is *very* large.

amadio: I changed somewhat how the `ThreadsPerBlock` get set, and added an explanation to the commit message about the tradeoffs involved.

hahnjo: Okay, but simply documenting the choice doesn't change the fact that it's slower!

amadio: Here are the numbers for 512 threads and 1 block per SM (requires forcing VecGeom to use max 128 registers):
```
batch time
10    8.82290
25    4.43577
50    3.12839
75    2.74347
100   2.48954
200   2.12174
250   2.04410
300   2.03968
400   1.96644
500   1.88576
750   1.89316
1000  1.81481
```
and here are the numbers for 32 threads and 16 blocks per SM:
```
batch time
10    5.01077
25    3.46367
50    2.88466
75    2.69932
100   2.56502
200   2.36028
250   2.35947
300   2.33717
400   2.27771
500   2.27287
750   2.30716
1000  2.23630
```
Please post your numbers. As you can see, with more threads and batch=1000, 1.81 is faster than 2.23s, although for small batches it is much slower. As you can also see, with batch >= 100, using more threads is already faster.

hahnjo: My numbers are:
```
10  5.91442
25  3.19493
50  2.40819
100  1.97531
250  1.71683
500  1.62653
750  1.62299
1000  1.57864
```
and with 32 threads, it is faster for ever single batch size:
```
10  4.41618
25  2.74391
50  2.17194
100  1.84505
250  1.64203
500  1.5693
750  1.56786
1000  1.52848
```

And when forcing 128 registers (which isn't part of this PR!)
```
10  5.9522
25  3.19099
50  2.40863
100  1.97024
250  1.71969
500  1.62509
750  1.62484
1000  1.579
```
with 32 threads:
```
10  4.43849
25  2.74141
50  2.17499
100  1.84385
250  1.6464
500  1.5745
750  1.56922
1000  1.53323
```

amadio: Well, I cannot force 128 registers for VecGeom as part of this PR, I just ran with what I already have compiled here. What I'm currently running also includes the changes from Bernhard for AdePT/G4HepEm for inlining more things, so you may want to try with that too. In any case, I still think it's better to merge this, even if we set the defaults to 32 threads and 8 blocks rather than 256 threads and 1 block.

hahnjo: As I commented in length in https://github.com/apt-sim/AdePT/pull/192#issuecomment-1090752104, I see no change from that. And please, can we stop evaluating a whole pile of changes at once?

PR #198, opened 2022-04-19
Added Example17 identical to Example14 with compacting track manager.

agheata:
This introduces the modificaions needed for example14 to operate with the TrackManager helper.


hahnjo (approved):
Not sure if we need all the HepMC files another time. I think we discussed this, but I forgot what we decided. Hopefully `git` is clever and deduplicates them internally...

agheata:
> Not sure if we need all the HepMC files another time. I think we discussed this, but I forgot what we decided. Hopefully `git` is clever and deduplicates them internally...

We said we go with dummy duplication for the moment, we will clean-up anyway at some point

PR #199, opened 2022-04-21
Added sensitivegroup command allowing to group scoring by name prefix.

agheata:
This introduces the command: `/example17/detector/sensitivegroup prefix`  (and the same for example14) allowing to sum-up hits for all placed volumes having the name starting with `prefix`.

Added also 2 new macros in Example17: example17_validation.in and lhcb_validation.in to be used for physics validation

### DIFF
--- a/examples/Example14/include/DetectorConstruction.hh
+++ b/examples/Example14/include/DetectorConstruction.hh
@@ -44,6 +44,7 @@ public:
   void SetGDMLFile(G4String &file) { fGDML_file = file; }
   void SetRegionName(G4String &reg) { fRegion_name = reg; }
   void AddSensitiveVolume(G4String volume) { fSensitive_volumes.push_back(volume); }
+  void AddSensitiveGroup(G4String group) { fSensitive_group.push_back(group); }
 
   // Set uniform magnetic field
   inline void SetMagField(const G4ThreeVector &fv) { fMagFieldVector = fv; }
@@ -64,6 +65,8 @@ public:
   // Set total number of track slots on GPU
   void SetTrackSlots(int value) { fTrackSlotsGPU = value; }
 
+  std::vector<G4String> &GetSensitiveGroups() { return fSensitive_group; }
+
 private:
   int fVerbosity{0};        ///< Actually verbosity for AdePT integration
   int fBufferThreshold{20}; ///< Buffer threshold for AdePT transport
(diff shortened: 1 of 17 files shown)


agheata:
For validation with more statistics, one can use directly `example17_validation.in` and `lhcb_validation.in` and change `/gun/number` to a larger number of tracks. There are 10 events by default on one thread and hits are printed, so based on the event-by-event scoring we can compute error bars running each macro just once per configuration. As configuration, we should at least have field on/off (the default is now off) and /gun/energy 10/100 GeV (default now is 10).

examples/Example17/AdeptIntegration.cu
```diff
@@ -234,6 +249,7 @@ void AdeptIntegration::InitializeGPU()
     gpuState.allmgr_d.trackmgr[i]  = gpuState.allmgr_h.trackmgr[i]->ConstructOnDevice();
     gpuState.particles[i].trackmgr = gpuState.allmgr_d.trackmgr[i];
     COPCORE_CUDA_CHECK(cudaMalloc(&gpuState.allmgr_d.leakedTracks[i], QueueSize));
+    InitLeakedQueues<<<1, 1>>>(gpuState.allmgr_d, kCapacity);
```
hahnjo, line 252: If you want to call this here, the kernel must only construct the current leaked queue, not all three of them.

hahnjo (commented):
Now that I think about this, I'm not so sure anymore about using queues for leaked particles in `Example17`: How do you prevent the compaction from re-using the slots? (In `Example14` this could never happen because slots weren't re-used, so just storing indices was fine)

agheata:
> Now that I think about this, I'm not so sure anymore about using queues for leaked particles in `Example17`: How do you prevent the compaction from re-using the slots? (In `Example14` this could never happen because slots weren't re-used, so just storing indices was fine)

Right, compaction should only happen after copying away the leaked tracks somewhere else. Which becomes more complicated...

agheata:
@hahnjo I removed the "fix" part which is invalid due to the way leaked tracks may be overridden by compaction, the crash may also be related to this.

examples/Example17/AdeptIntegration.cu
agheata: removed from this PR

hahnjo (approved):

hahnjo:
Oh, but CI failed...

agheata:
should be fixed now, I will squash

PR #200, opened 2022-04-22
Sorting tracks coming from GPU to make integration example reproducible.

agheata:
To make the integration example reproducible, we have to do a full sort of the tracks leaked from the device back to Geant4. This is because the order of the tracks depends on device concurrency, and the random seed sequence on the host depends on the order since it is per CPU thread, not per track.

This change sorts tracks by the product of PDG number, energy, position and momentum components to ensure to a reasonable level the uniqueness of the sorting metric (energy is not sufficient for the gun generator with exa14). 

I am now getting numerically identical results when running multiple times with multiple threads.

### DIFF
--- a/examples/Example14/AdeptIntegration.cu
+++ b/examples/Example14/AdeptIntegration.cu
@@ -494,7 +494,7 @@ void AdeptIntegration::ShowerGPU(int event, TrackBuffer &buffer) // const &buffe
     fSorted.reserve(numLeaked);
     std::iota(fSorted.begin(), fSorted.begin() + numLeaked, 0); // Fill with 0, 1, ...
     std::sort(fSorted.begin(), fSorted.begin() + numLeaked,
-              [&](int i, int j) {return fBuffer.fromDevice[i].FullSort() > fBuffer.fromDevice[j].FullSort();});
+              [&](int i, int j) {return fBuffer.fromDevice[i] < fBuffer.fromDevice[j];});
     fBuffer.fromDevice_sorted.clear();
     fBuffer.fromDevice_sorted.reserve(numLeaked);
     for (auto i = 0; i < numLeaked; ++i)
--- a/examples/Example14/AdeptIntegration.cu
+++ b/examples/Example14/AdeptIntegration.cu
@@ -32,6 +32,9 @@
 #include <iostream>
 #include <iomanip>
 #include <stdio.h>
+#include <vector>
+#include <numeric>
+#include <algorithm>
 
 #include "electrons.cuh"
 #include "gammas.cuh"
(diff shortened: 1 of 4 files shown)


examples/Example14/CommonStruct.h
```diff
@@ -29,12 +29,18 @@ struct TrackData {
       : position{x, y, z}, direction{dirx, diry, dirz}, energy{ene}, pdg{pdg_id}
   {
   }
+  
+  inline double FullSort()
+  {
+    return pdg * energy * position[0] * position[1] * position[2] * direction[0] * direction[1] * direction[2];
+  }
```
hahnjo, line 36: This isn't very robust, if one of the value happens to be zero the product will be as well. Why not implement a proper `operator<`?

hahnjo (commented):
We should do the same for `Example17`

agheata:
> We should do the same for `Example17`

I will add this to Example17 when I will fix the other problems, for the moment we will not use it for validation

examples/Example14/CommonStruct.h
agheata: I wanted to test different sorting criteria, so it can stay as a lambda for a while. But the problem with 0 is indeed an issue, I'll fix this

hahnjo: But why the multiplication? `operator<` can compare a number of members one after the other.

examples/Example14/AdeptIntegration.cu
```diff
@@ -487,6 +490,15 @@ void AdeptIntegration::ShowerGPU(int event, TrackBuffer &buffer) // const &buffe
         cudaMemcpyAsync(gpuState.stats, gpuState.stats_dev, sizeof(Stats), cudaMemcpyDeviceToHost, gpuState.stream));
     COPCORE_CUDA_CHECK(cudaMemcpyAsync(fBuffer.fromDevice, gpuState.fromDevice_dev, numLeaked * sizeof(TrackData),
                                        cudaMemcpyDeviceToHost, gpuState.stream));
+    // Sort by energy the tracks coming from device to ensure reproducibility
+    fSorted.reserve(numLeaked);
+    std::iota(fSorted.begin(), fSorted.begin() + numLeaked, 0); // Fill with 0, 1, ...
+    std::sort(fSorted.begin(), fSorted.begin() + numLeaked,
+              [&](int i, int j) {return fBuffer.fromDevice[i].FullSort() > fBuffer.fromDevice[j].FullSort();});
```
hahnjo, line 497: needs to be updated

examples/Example14/CommonStruct.h
```diff
@@ -29,12 +29,34 @@ struct TrackData {
...
+    if (position[2] < t.position[2]) return true;
+    if (position[2] > t.position[2]) return false;
+    if (direction[0] < t.direction[0]) return true;
+    if (direction[0] > t.direction[0]) return false;
+    if (direction[1] < t.direction[1]) return true;
+    if (direction[1] > t.direction[1]) return false;
+    if (direction[2] < t.direction[2]) return true;
+    if (direction[2] > t.direction[2]) return false;
+    return false;
+  }
```
hahnjo, line 52: That's a bit un-idiomatic to write the `operator<` (I think I've seen `if (pdg != t.pdg) return pdg < t.pdg;` more often), but ok for now.

hahnjo (approved):

examples/Example14/CommonStruct.h
agheata: Indeed nicer this way, adopted it

PR #202, opened 2022-04-26
New macro allowing to run integration TestEm3 with interlaced CPU-GPU regions

agheata:
We can now use the testEm3_interlaced.mac macro in example14 to test the impact of interlacing CPU with GPU simulation regions. Layer volumes are now separate and can be added to the `caloregion` region in testEm3_interlaced.gdml. The default now is that the first 10 and last 10 layers are controlled by the GPU.

### DIFF
--- a/examples/Example14/AdeptIntegration.cpp
+++ b/examples/Example14/AdeptIntegration.cpp
@@ -112,7 +112,7 @@ void AdeptIntegration::Cleanup()
 
 void AdeptIntegration::Shower(int event)
 {
-  constexpr double tolerance = 10. * vecgeom::kTolerance;
+  constexpr double tolerance = 0.1 * vecgeom::kTolerance;
 
   int tid = G4Threading::G4GetThreadId();
   if (fDebugLevel > 0 && fBuffer.toDevice.size() == 0) {
@@ -197,7 +197,7 @@ void AdeptIntegration::Shower(int event)
   evAct->number_killed    = evAct->number_killed + fScoring->fGlobalScoring.numKilled;
 
   fBuffer.Clear();
-  fScoring->ClearGPU();
+  //fScoring->ClearGPU();
 }
 
 adeptint::VolAuxData *AdeptIntegration::CreateVolAuxData(const G4VPhysicalVolume *g4world,
--- a/examples/Example14/AdeptIntegration.cu
+++ b/examples/Example14/AdeptIntegration.cu
@@ -475,8 +475,10 @@ void AdeptIntegration::ShowerGPU(int event, TrackBuffer &buffer) // const &buffe
       gpuState.fNumFromDevice = numLeaked;
       delete[] fBuffer.fromDevice;
       fBuffer.fromDevice = new TrackData[numLeaked];
+      // From 11.6.2 we can start using the Async versions for malloc/free avoiding full device synchronization
       COPCORE_CUDA_CHECK(cudaFree(gpuState.fromDevice_dev));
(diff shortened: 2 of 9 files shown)


agheata:
I just added a commit that exposes the bug in a different way: the GPU region is composed of ALL layers (1 to 50) in testEm3_interlaced.gdml, as opposed to just the Calorimeter volume that contains all layers. So the results are supposed to be the same. However, when having the layers added separately to the GPU region, there are particles that leak to Geant4 BETWEEN the layers (which are touching each other). This relocation issue pushes some tracks to Geant4 and these come back at the next step to AdePT in the next layer and this creates a systematic difference in the output (number of electrons/gammas and energy deposit) when shooting like 10K electrons:
`example14 -m testEm3_interlaced.mac`
```
G4WT0 > EndOfEventAction 0: electrons 89640950
G4WT0 > EndOfEventAction 0: positrons 5388144
G4WT0 > EndOfEventAction 0: gammas    52594082
G4WT0 > EndOfEventAction 0Total energy deposited: 99001349.78 MeV
```

`example14 -m testEm3.mac`

```
G4WT0 > EndOfEventAction 0: electrons 88586670
G4WT0 > EndOfEventAction 0: positrons 5384377
G4WT0 > EndOfEventAction 0: gammas    51992282
G4WT0 > EndOfEventAction 0Total energy deposited: 98703791.70 MeV
```

This is reproducible by running the 2 macros in the build directory

drbenmorgan:
Assigned to self to start looking at this problem

examples/Example14/AdeptIntegration.cpp
```diff
@@ -197,7 +197,7 @@ void AdeptIntegration::Shower(int event)
   evAct->number_killed    = evAct->number_killed + fScoring->fGlobalScoring.numKilled;
 
   fBuffer.Clear();
-  fScoring->ClearGPU();
+  //fScoring->ClearGPU();
```
hageboeck, line 200: I suggest to remove it instead of commenting it out.

examples/Example14/electrons.cuh
```diff
@@ -39,7 +39,7 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
 #else
   const Precision kPush = 0.;
 #endif
-  constexpr Precision kPushOutRegion = 10 * vecgeom::kTolerance;
+  //constexpr Precision kPushOutRegion = 10 * vecgeom::kTolerance;
```
hageboeck, line 42: Also suggest to remove it.

examples/Example14/AdeptIntegration.cpp
hageboeck:
```suggestion
```

examples/Example14/electrons.cuh
hageboeck:
```suggestion
```

examples/Example14/electrons.cuh
```diff
@@ -241,7 +241,7 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
           leakedQueue->push_back(slot);
           // To be safe, just push a bit the track exiting the GPU region to make sure
           // Geant4 does not relocate it again inside the same region
-          currentTrack.pos += kPushOutRegion * currentTrack.dir;
+          //currentTrack.pos += kPushOutRegion * currentTrack.dir;
```
hageboeck, line 244: And here, but I cannot add a suggestion.

agheata:
Sorry @hageboeck I forgot to put this in draft mode, since it's only intended as a bug reproducer and not for merging

agheata:
Not needed anymore

PR #203, opened 2022-04-28
Run interactions in separate kernels for TestEm3

hageboeck:
Demo of how one can split interactions from the rest of the workflow, i.e. splitting off highly divergent code.

There's a few things that should still be worked on, namely
- [ ] Split off the build system/NVTX and related things in a different PR
- [ ] Count how much work is actually scheduled into each kernel to compute optimal number of blocks
- [ ] Don't use hard-coded buffers (probably in conjunction with track buffer), but this won't change timings


examples/TestEm3/electrons.cu
```diff
@@ -28,6 +28,9 @@
 #include <G4HepEmElectronEnergyLossFluctuation.icc>
 #endif
 
+__device__ char g_nextInteractionForEl[8'000'000];
+__device__ char g_nextInteractionForPos[8'000'000];
+
```
hahnjo, line 33: As you write in the summary, this isn't very nice. What is the plan here, move the `winnerProcessIndex` into the `Track` structure?

Another approach (that I think we had in the past? not sure) is having queues for the discrete processes. That would even save us from determining which `Track` need to run in which of the (split) kernels.

examples/TestEm3/electrons.cu
```diff
@@ -258,11 +265,83 @@ void TransportElectrons(Track *electrons, const adept::MParray *active,
...
+    auto volume         = currentTrack.navState.Top();
+    int volumeID        = volume->id();
+    int theMCIndex      = MCIndex[volumeID];
+
     const double energy   = currentTrack.energy;
     const double theElCut = g4HepEmData.fTheMatCutData->fMatCutData[theMCIndex].fSecElProdCutE;
 
-    switch (winnerProcessIndex) {
-    case 0: {
+    RanluxppDouble newRNG(currentTrack.rngState.BranchNoAdvance());
```
hahnjo, line 340: This *must* use `Branch()` (with `Advance`ing) or `newRNG` has correlations to the primary's RNG state, see #129. In turn, the calls to `newRNG.Advance();` and `currentTrack.rngState.Advance();` in the `Transport` kernel above are *not* needed anymore.

examples/TestEm3/gammas.cu
```diff
@@ -104,15 +109,57 @@ void TransportGammas(Track *gammas, const adept::MParray *active, Secondaries se
...
+
+  const int activeSize = active->size();
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < activeSize; i += blockDim.x * gridDim.x) {
+    const auto winnerProcess = g_nextInteractionForGamma[i];
+
+    if (winnerProcess == ProcessIndex) {
+      const auto destination = atomicInc(&counter, (unsigned int)-1);
+      candidates[destination % sharedSize] = i;
+    } else atomicInc(&noopCounter, (unsigned int)-1);
+  }
```
hahnjo, line 139: This is a race condition, there needs to be a `__syncthreads()` in between. This is at least one of the reasons I was seeing non-reproducible results on my machine (when forcing `-batch 104`, the previous default)

examples/TestEm3/electrons.cu
```diff
@@ -258,11 +265,83 @@ void TransportElectrons(Track *electrons, const adept::MParray *active,
...
+  }
+  __syncthreads();
+
+  const int activeSize = active->size();
+  assert(activeSize < sizeof(g_nextInteractionForEl)/sizeof(g_nextInteractionForEl[0]));
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < activeSize; i += blockDim.x * gridDim.x) {
+    const auto winnerProcess = IsElectron ? g_nextInteractionForEl[i] : g_nextInteractionForPos[i];
+
+    if (winnerProcess == ProcessIndex) {
+      const auto destination = atomicInc(&counter, (unsigned int)-1);
```
hahnjo, line 318: I find `atomicAdd(&counter, 1)` much easier to understand. For `atomicInc`, you need to know that the second argument is the "maximum" where the atomic increment is clamped to...

base/inc/AdePT/NVTX.h
```diff
@@ -22,6 +23,8 @@ class NVTXTracer {
                                                        0xff00ffff, 0xffff0000, 0xffffffff};
   std::string _name;
   nvtxRangeId_t _id;
+  std::array<unsigned long, 5> _lastOccups;
```
bernhardmgruber, line 26:
```suggestion
  std::array<unsigned long, 5> _lastOccups{};
```

base/inc/AdePT/NVTX.h
```diff
@@ -34,10 +37,11 @@ class NVTXTracer {
     eventAttrib.messageType           = NVTX_MESSAGE_TYPE_ASCII;
     eventAttrib.message.ascii         = name;
     _id                               = nvtxRangeStartEx(&eventAttrib);
+    _lastOccups.fill(0);
```
bernhardmgruber, line 40:
```suggestion
```

base/inc/AdePT/NVTX.h
```diff
@@ -54,6 +58,22 @@ class NVTXTracer {
     _id                               = nvtxRangeStartEx(&eventAttrib);
   }
 
+  void setOccupancy(unsigned long occupancy)
+  {
+    *_occupIt = occupancy;
+    if (++_occupIt == _lastOccups.end()) _occupIt = _lastOccups.begin();
+
+    const auto meanOccup = double(std::accumulate(_lastOccups.begin(), _lastOccups.end(), 0)) / _lastOccups.size();
```
bernhardmgruber, line 66:
```suggestion
    const auto meanOccup = double(std::reduce(_lastOccups.begin(), _lastOccups.end())) / _lastOccups.size();
```

examples/TestEm3/TestEm3.cu
```diff
@@ -215,9 +224,9 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double
   //  * objects to manage slots inside the memory,
   //  * queues of slots to remember active particle and those needing relocation,
   //  * a stream and an event for synchronization of kernels.
-  constexpr size_t TracksSize  = sizeof(Track) * Capacity;
-  constexpr size_t ManagerSize = sizeof(SlotManager);
-  const size_t QueueSize       = adept::MParray::SizeOfInstance(Capacity);
+  size_t TracksSize  = sizeof(Track) * Capacity;
+  size_t ManagerSize = sizeof(SlotManager);
+  size_t QueueSize   = adept::MParray::SizeOfInstance(Capacity);
```
bernhardmgruber, line 229: Why did you make these variables mutable? I do not see where you need to change them later.

examples/TestEm3/TestEm3.cuh
```diff
@@ -28,6 +30,7 @@ struct Track {
   double initialRange;
   double dynamicRangeFactor;
   double tlimitMin;
+  double fPEmxSec;
```
bernhardmgruber, line 33: I am not experienced in this domain, but even then, this looks like a very cryptic variable name.

examples/TestEm3/electrons.cu
```diff
@@ -24,15 +24,22 @@
 #include <G4HepEmElectronInteractionIoni.icc>
 #include <G4HepEmElectronInteractionUMSC.icc>
 #include <G4HepEmPositronInteractionAnnihilation.icc>
+#ifndef NOFLUCTUATION
+#include <G4HepEmElectronEnergyLossFluctuation.icc>
+#endif
+
+__device__ char g_nextInteractionForEl[8'000'000];
+__device__ char g_nextInteractionForPos[8'000'000];
```
bernhardmgruber, line 32: Could you quickly elaborate to me please, why we cannot add this information to the `Track` itself? Then we would not need this additional global state and the checking of whether you have allocated enough of it.

examples/TestEm3/electrons.cu
```diff
@@ -247,18 +258,86 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
...
+}
+
+template<bool IsElectron, int ProcessIndex>
+__global__
+void ComputeInteraction(Track *electrons, const adept::MParray *active, Secondaries secondaries,
+                        adept::MParray *activeQueue, GlobalScoring *globalScoring,
+                        ScoringPerVolume *scoringPerVolume)
+{
+  constexpr unsigned int sharedSize = 12250;
+  __shared__ int candidates[sharedSize];
```
bernhardmgruber, line 298: Where is this number coming from?  The size of candidates is 49KiB, which seems like an odd size to me.

examples/TestEm3/electrons.cu
```diff
@@ -247,18 +258,86 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
...
+      candidates[destination % sharedSize] = i;
+    } else atomicInc(&noopCounter, (unsigned int)-1);
+  }
+
+  __syncthreads();
+  assert(counter < sharedSize);
+  if (threadIdx.x == 0 && counter >= sharedSize) {
+    printf("Error: Shared queue for %d exhausted to %d in %s:%d\n", ProcessIndex, counter, __FILE__, __LINE__);
+    asm("trap;");
+  }
```
bernhardmgruber, line 325: Could you not just wrap another loop around the entire kernel that fills a chunk of shared memory, processes it, and then starts with the next chunk?

bernhardmgruber:
The current branch has a build failure on my machine, which I fixed upstream:
```
/home/bgruber/dev/AdePT/magneticfield/inc/fieldPropagatorConstBz.h(139): error: expression must be a modifiable lvalue
```
Could you please rebase onto a more recent master? Thank you!

bernhardmgruber:
For the reason given [here](https://github.com/apt-sim/AdePT/pull/207#issuecomment-1153690475) I am still super interested in having these changes on TestEm3. But sure, they should also definitely land in more complex examples!

examples/TestEm3/TestEm3.cu
```diff
@@ -193,10 +195,17 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double
   COPCORE_CUDA_CHECK(cudaMemcpy(MCIndex_dev, MCIndex_host, sizeof(int) * numVolumes, cudaMemcpyHostToDevice));
   COPCORE_CUDA_CHECK(cudaMemcpyToSymbol(MCIndex, &MCIndex_dev, sizeof(int *)));
 
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, 0);
+
   // Capacity of the different containers aka the maximum number of particles.
-  constexpr int Capacity = 1024 * 1024;
+  // Use 2/7 of GPU memory for each of e+/e-/gammas, leaving 1/7 for the rest.
+  int Capacity = deviceProp.totalGlobalMem / sizeof(Track) * (2.0 / 7.0);
```
bernhardmgruber, line 203: So I realized that the choice of `Capacity` has a gigantic impact on the runtime, because it directly determines the number of batches that will be iterated on later. We should definitely separate this change from the rest, so we can benchmark the impact of the kernel separation alone.

PR #207, opened 2022-05-17
Use double buffering for storing tracks in TestEm3

bernhardmgruber:
The main difference of this version of TestEm3 is that it uses two storage arrays per track type. The kernels then always read the full track and write surviving tracks back to the other array. This way, the tracks are always stored dense without gaps in-between.

First tests on my workstation showed that this approach is actually a bit slower than the base TestEm3. However, with dense track arrays we can now easily switch to SoA and other memory layouts and profit from better memory access patterns again. Anyway, more investigation is pending.

Please merge first: 

- [x] #209


bernhardmgruber:
> The commit is missing many files I think, in particular the `CMakeLists.txt`, `electrons.cu` and `gammas.cu`.

Thx for spotting! I rebased before opening the PR and seemed to have messed up.

> Also can you please start your commit messages with a capital letter (as is usually commended) and provide some details in the commit message itself, not only the PR summary?

Done, thx!

bernhardmgruber:
Here are some measurements of `TestEm3 --batch 1000` on @hageboeck's V100:
```
* master
	Run time: 1.23467
	Run time: 1.23561
	Run time: 1.23438
	Run time: 1.2378
	Run time: 1.23653
* double buffer:
	Run time: 1.38669
	Run time: 1.38506
	Run time: 1.38961
	Run time: 1.38551
	Run time: 1.38619
* master SoA
	Run time: 1.19734
	Run time: 1.19712
	Run time: 1.19742
	Run time: 1.19608
	Run time: 1.19669
* double buffer SoA
	Run time: 1.15108
	Run time: 1.15122
	Run time: 1.14912
	Run time: 1.14987
	Run time: 1.14978
	Run time: 1.14793
```
The double buffer version is slower than master, but if we switch it to a SoA layout later (not proposed in this PR), it beats master and master switched to a SoA layout.

agheata (approved):

hahnjo:
Please note that TestEm3 is mostly relevant for physics validation; it was originally thought of as the "best" workflow with the setup hard-coded for validation studies. It was never the plan to have many "variants" of it, although I realize we already had some before this PR. What we're however missing here is a separate example capable of loading arbitrary GDML files for studying more realistic performance characteristics - it doesn't help making a trivial example such as TestEm3 faster if you cannot demonstrate better performance for the more complete cases.

bernhardmgruber:
Thank you for pointing that out! I am aware that TestEm3 is not the real, final thing. However, it already contains the entire physics workload, which still makes it a great benchmark. I agree, there is more work necessary to handle the geometry faster for more realistic scenarios, but that is already someone else's work. So for me to progress on my memory layout studies with LLAMA, I intend to stay with TestEm3 and avoid the geometry problem for now.

bernhardmgruber:
That is for sure a valid argument! I think it is still important that we also play with the memory layout. And the double buffering was an easy way to achieve a dense buffer, which allows me to more easily explore further memory layout aspects.

hahnjo:
> What we're however missing here is a separate example capable of loading arbitrary GDML files for studying more realistic performance characteristics - it doesn't help making a trivial example such as TestEm3 faster if you cannot demonstrate better performance for the more complete cases.

Ping, can you please work on providing a complete example here?

PR #208, opened 2022-05-24
fixing hepmc reader for MT

WitekPokorski:
A new version of HepMC reader which should work fine for MT. File is read once into a vector of events and each thread accesses it by the event number. If event number is larger than the size of the vector, the event number is taken modulo the number of events read from the file.

### DIFF
--- a/examples/Example14/include/HepMC3G4AsciiReader.hh
+++ b/examples/Example14/include/HepMC3G4AsciiReader.hh
@@ -21,10 +21,12 @@ protected:
   G4String filename;
   HepMC3::ReaderAscii* asciiInput;
 
+  static std::vector<HepMC3::GenEvent*>* fEvents;
+
   G4int verbose;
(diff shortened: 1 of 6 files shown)


examples/Example14/include/HepMC3G4AsciiReader.hh
```diff
@@ -19,12 +19,16 @@ class HepMC3G4AsciiReaderMessenger;
 class HepMC3G4AsciiReader : public HepMC3G4Interface {
 protected:
   G4String filename;
+  int firstEventNumber = 0;
+  int maxNumberOfEvents = __INT_MAX__;
```
agheata, line 23: Maybe safer to use `INT_MAX` after including `<climits>`? Not sure where `__INT_MAX__` is defined and if it is portable.

agheata (approved):

Issue #210, opened 2022-06-16
HepMC3 primary generation crashes if re-using the same event


hahnjo:
https://github.com/apt-sim/AdePT/pull/208 added the possibility to simulate more events with Geant4 than there are in the HepMC3 files, by wrapping around when reaching the last event. However, actually trying this results in crashes because `HepMC3G4Interface` expects that `GenerateHepMCEvent` returns an owning pointer and consequently `delete`s it in the next event.

PR #211, opened 2022-06-17
removing unnecessary delete

WitekPokorski:
fixing https://github.com/apt-sim/AdePT/issues/210 
(removing unnecessary delete)

### DIFF
--- a/examples/Example14/include/HepMC3G4Interface.hh
+++ b/examples/Example14/include/HepMC3G4Interface.hh
@@ -35,8 +35,7 @@ protected:
   void HepMC2G4(const HepMC3::GenEvent* hepmcevt, G4Event* g4event);
 
   // Implement this method in his/her own concrete class.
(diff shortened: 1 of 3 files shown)


examples/Example14/include/HepMC3G4Interface.hh
```diff
@@ -35,8 +35,7 @@ protected:
   void HepMC2G4(const HepMC3::GenEvent* hepmcevt, G4Event* g4event);
 
   // Implement this method in his/her own concrete class.
-  // An empty event will be created in default.
-  virtual HepMC3::GenEvent* GenerateHepMCEvent(int);
+  virtual HepMC3::GenEvent* GenerateHepMCEvent(int) = 0;
```
hahnjo, line 38: If you make this function pure-virtual, you should delete the implementation in `HepMC3G4Interface.cc`.

PR #212, opened 2022-06-17
Add new example with faster field propagation

hahnjo:
This is based on Example 13, but restricts the step length of charged
particles in case of magnetic field for better performance. The idea
is to make shorter steps where propagation finishes after a low-ish
number of chord steps to reduce divergence. The step limit has to
happen *before* MSC to avoid approximations when just aborting field
propagation (as was already possible with the max_iterations argument).

Performance numbers on RTX 2070 SUPER, with `-particles 1000`:
```
  Bz = 1T
    example13: 24.74s, example18: 23.10s, -6.7%
  Bz = 3.8T
    example13: 96.04s, example18: 79.46s, -17.3%
```


examples/Example18/electrons.cu
```diff
@@ -0,0 +1,386 @@
...
+        physicalStepLength           = limit;
+        restrictedPhysicalStepLength = true;
+        elTrack.SetPStepLength(physicalStepLength);
+        // Note: We are limiting the true step length, which is converted to
+        // a shorter geometry step length in HowFarToMSC. In that sense, the
+        // limit is an over-approximation, but that is fine for our purpose.
+      }
+    }
+
+    G4HepEmElectronManager::HowFarToMSC(&g4HepEmData, &g4HepEmPars, &elTrack, &rnge);
```
hahnjo, line 118: This is the interesting part that changes compared to Example 13

hahnjo (commented):
@jonapost @mnovak42 this implements the strategy that we discussed some time ago during a coffee at the community meeting. It's worth something, but you can see that 3.8T is still very slow. It's probably possible to tweak the `limit` heuristic a bit, but I wouldn't expect huge factors out of that...

hahnjo:
ping @agheata @jonapost

agheata (approved):

PR #213, opened 2022-06-24
Keep some Track members in registers

bernhardmgruber:
It is beneficial to keep some of the Track members in registers for the duration of the kernel.

Just running `TestEm3` without arguments on my RTX 2060, 5 time mean:
before: 2.67394s
after: 2.61814s

`TestEm3DoubleBuffer`:
before: 2.93838s
after: 2.6724s

Examples with are correct:

- [ ] Example13
- [ ] Example14
- [ ] Example16
- [ ] Example17
- [ ] Example18
- [x] Example19
- [ ] TestEm3
- [ ] TestEm3DoubleBuffer

### DIFF
--- a/examples/TestEm3/electrons.cu
+++ b/examples/TestEm3/electrons.cu
@@ -1,4 +1,4 @@
-// SPDX-FileCopyrightText: 2021 CERN
+// SPDX-FileCopyrightText: 2022 CERN
 // SPDX-License-Identifier: Apache-2.0
 
 #include "TestEm3.cuh"
@@ -42,19 +42,27 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < activeSize; i += blockDim.x * gridDim.x) {
     const int slot       = (*active)[i];
     Track &currentTrack  = electrons[slot];
-    const int volumeID   = currentTrack.navState.Top()->id();
+    auto energy          = currentTrack.energy;
+    auto pos             = currentTrack.pos;
+    auto dir             = currentTrack.dir;
+    auto navState        = currentTrack.navState;
+    const int volumeID   = navState.Top()->id();
     const int theMCIndex = MCIndex[volumeID];
 
     auto survive = [&] {
+      currentTrack.energy   = energy;
+      currentTrack.pos      = pos;
+      currentTrack.dir      = dir;
+      currentTrack.navState = navState;
       activeQueue->push_back(slot);
     };
 
     // Init a track with the needed data to call into G4HepEm.
     G4HepEmElectronTrack elTrack;
     G4HepEmTrack *theTrack = elTrack.GetTrack();
-    theTrack->SetEKin(currentTrack.energy);
+    theTrack->SetEKin(energy);
     theTrack->SetMCIndex(theMCIndex);
-    theTrack->SetOnBoundary(currentTrack.navState.IsOnBoundary());
+    theTrack->SetOnBoundary(navState.IsOnBoundary());
     theTrack->SetCharge(Charge);
     G4HepEmMSCTrackData *mscData = elTrack.GetMSCTrackData();
     mscData->fIsFirstStep        = currentTrack.initialRange < 0;
@@ -69,8 +77,8 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
 
     // Compute safety, needed for MSC step limit.
     double safety = 0;
-    if (!currentTrack.navState.IsOnBoundary()) {
-      safety = BVHNavigator::ComputeSafety(currentTrack.pos, currentTrack.navState);
+    if (!navState.IsOnBoundary()) {
+      safety = BVHNavigator::ComputeSafety(pos, navState);
     }
     theTrack->SetSafety(safety);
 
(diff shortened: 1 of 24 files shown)


examples/TestEm3/electrons.cu
```diff
@@ -80,8 +88,7 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
     for (int ip = 0; ip < 3; ++ip) {
       double numIALeft = currentTrack.numIALeft[ip];
       if (numIALeft <= 0) {
-        numIALeft                  = -std::log(currentTrack.Uniform());
-        currentTrack.numIALeft[ip] = numIALeft;
+        numIALeft = -std::log(currentTrack.Uniform());
```
bernhardmgruber, line 91: The write to `currentTrack.numIALeft[ip]` should not be necessary, because it is overwritten later

bernhardmgruber:
Ping.

bernhardmgruber:
Should we apply these changes to other examples as well? Which ones?

examples/TestEm3/electrons.cu
hahnjo: Correct.

hahnjo:
> Should we apply these changes to other examples as well? Which ones?

I would say at least to the recent ones with full physics including MSC, so starting with `example13`.

bernhardmgruber:
I updated my Geant4 installation and could also build examples 14 and 17 now.

bernhardmgruber:
And all tests passed locally :)

hahnjo:
Seems to work and give identical results for `example13` and `example16`, with tiny improvements in performance. For `example14` though, execution hangs for larger runs and the few events that finished before I aborted the run were not identical...


[example14_validation.10GeV.48x1000.GPU.t12.mac](https://github.com/apt-sim/AdePT/files/9044896/example14_validation.10GeV.48x1000.GPU.t12.mac.txt)

bernhardmgruber:
I fixed an issue in example14 and example 17. Could you please try again? Thank you for the thorough testing!

bernhardmgruber:
I rebased on master and applied the changes to example18 as well.

bernhardmgruber:
> This still leads to hangs in `example14` on my machine with the same macro I posted yesterday...

I have troubles running the macro on my workstation. It always goes out of memory, even if I set `int AdeptIntegration::kCapacity = 128;` and also if a try with reduced capacity on `master`.

hahnjo:
> > This still leads to hangs in `example14` on my machine with the same macro I posted yesterday...
> 
> I have troubles running the macro on my workstation. It always goes out of memory, even if I set `int AdeptIntegration::kCapacity = 128;` and also if a try with reduced capacity on `master`.

`kCapacity` is adjusted based on the UI command `/example14/adept/milliontrackslots 8`. If that doesn't fit on your GPU, you have to lower that number, and possibly also adjust `/example14/adept/threshold`.

bernhardmgruber:
> `kCapacity` is adjusted based on the UI command `/example14/adept/milliontrackslots 8`. If that doesn't fit on your GPU, you have to lower that number, and possibly also adjust `/example14/adept/threshold`.

Thanks, I am trying it with half the values for `milliontrackslots` and `threshold` and can reproduce the hang.

bernhardmgruber:
@hahnjo your macro now runs fine (with halfed `milliontrackslots` and `threshold`) on my workstation (RTX 2060).

hahnjo:
Thanks, this finishes now also on my side. However, I see changed results in the integration `example14` which surprises me because all other examples were clean - are changes expected? And it's not only rounding errors in the summed energy deposition, but also significant changes in the number of particles which likely means that something in the RNG sequence in Geant4 changes:
```
G4WT5 > EndOfEventAction 31: electrons 8640225
G4WT5 > EndOfEventAction 31: positrons 543250
G4WT5 > EndOfEventAction 31: gammas    4889163
G4WT5 > EndOfEventAction 31: killed    0
```
and with the change merged:
```
G4WT5 > EndOfEventAction 31: electrons 8647634
G4WT5 > EndOfEventAction 31: positrons 543651
G4WT5 > EndOfEventAction 31: gammas    4892197
G4WT5 > EndOfEventAction 31: killed    0
```

bernhardmgruber:
> However, I see changed results in the integration example14 which surprises me because all other examples were clean - are changes expected?

Not at all, there should be no changes! And example17 ran fine?

hahnjo:
> And example17 ran fine?

I didn't test because I know that it has problems with the compaction; in particular it's not reproducible, which is always bad for a baseline of a comparison...

hahnjo:
I can still reproduce changes in `example14` (after applying https://github.com/apt-sim/AdePT/pull/230)

bernhardmgruber:
I reviewed the diff again and indeed found two instances where I did not replace `currentTrack.navState` by just `navState` in Example14 and Example17. I fixed those two now.

hahnjo:
> I reviewed the diff again and indeed found two instances where I did not replace `currentTrack.navState` by just `navState` in Example14 and Example17. I fixed those two now.

Good, many events are now identical :+1: however, a few still show different histories whereas I understand they should not. Please take a look.

Also, testing `example13` / `example18` with magnetic field yields different results. `TestEm3` with magnetic field seems to give identical results though, not sure what this means...

For `example19`, do you want to fold https://github.com/apt-sim/AdePT/pull/225 into this PR or keep that one separate. AFAICT, it's basically the same modification...

bernhardmgruber:
> For example19, do you want to fold https://github.com/apt-sim/AdePT/pull/225 into this PR or keep that one separate. AFAICT, it's basically the same modification...

Answered here: https://github.com/apt-sim/AdePT/pull/225#issuecomment-1305800680. Please merge #225 sooner if possible, I will try to get this PR in later then!

bernhardmgruber:
I included the changes from #225 in this PR now.

hahnjo (approved):
I verified that Example13, Example14, Example18, Example19, TestEm3, and TestEm3DoubleBuffer give identical results when compiled with no `fmad`.

PR #214, opened 2022-06-24
Automatically adjust buffer capacity in TestEm3

bernhardmgruber:
This change automatically determines the buffer capacity for the electron, positron and gamma buffers based on the total available global memory. This change is a derivation of the first commit in #197.

The choice of using 1/4th of the global memory size as Capacity for each of the 3 buffers is rather arbitrary and smaller to what @amadio proposed before. However, 2/7th did not work on my workstation as I seem to have more VRAM occupied for my desktop environment. It is still important though that most of the available VRAM is also available to the example being run and not just a hard-coded constant that people tweak depending on their GPU memory size.

### DIFF
--- a/examples/TestEm3/TestEm3.cu
+++ b/examples/TestEm3/TestEm3.cu
@@ -193,8 +193,12 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double 
   COPCORE_CUDA_CHECK(cudaMemcpy(MCIndex_dev, MCIndex_host, sizeof(int) * numVolumes, cudaMemcpyHostToDevice));
   COPCORE_CUDA_CHECK(cudaMemcpyToSymbol(MCIndex, &MCIndex_dev, sizeof(int *)));
 
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, 0);
+
   // Capacity of the different containers aka the maximum number of particles.
-  constexpr int Capacity = 1024 * 1024;
+  // Use 1/4 of GPU memory for each of e+/e-/gammas, leaving 1/4 for the rest.
+  const size_t Capacity = (deviceProp.totalGlobalMem / sizeof(Track)) / 4;
 
   std::cout << "INFO: capacity of containers set to " << Capacity << std::endl;
   if (batch == -1) {
@@ -215,7 +219,7 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double 
   //  * objects to manage slots inside the memory,
   //  * queues of slots to remember active particle and those needing relocation,
   //  * a stream and an event for synchronization of kernels.
-  constexpr size_t TracksSize  = sizeof(Track) * Capacity;
+  const size_t TracksSize      = sizeof(Track) * Capacity;
   constexpr size_t ManagerSize = sizeof(SlotManager);
   const size_t QueueSize       = adept::MParray::SizeOfInstance(Capacity);
 
@@ -314,7 +318,7 @@ void TestEm3(const vecgeom::cxx::VPlacedVolume *world, int numParticles, double 
     int transportBlocks;
 
     int inFlight;
-    int loopingNo = 0;
+    int loopingNo         = 0;
     int previousElectrons = -1, previousPositrons = -1;
 
     do {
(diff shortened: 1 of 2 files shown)


hahnjo:
> It is still important though that most of the available VRAM is also available to the example being run and not just a hard-coded constant that people tweak depending on their GPU memory size.

Why not just use the default one and be done with it? We had many discussions on variations of this patch already, I don't see what changes to past proposals of this. It's just not realistic to use the maximum memory and then expect the simulation to provide enough particles *in one event* to fill that.

bernhardmgruber:
> It's just not realistic to use the maximum memory and then expect the simulation to provide enough particles in one event to fill that.

I explained that yesterday during the talk. We are doing R&D work and while, apparently (because I have only heard about this), Geant4 cannot easily provide particles from more than 1 event now, it might be able to do so a few years down the road. So it is worthwhile to consider relaxed requirements for doing R&D work. And we need the bigger batch sizes and thus capacities to explore how to utilize GPUs efficiently.

hahnjo:
> I explained that yesterday during the talk. We are doing R&D work and while, apparently (because I have only heard about this), Geant4 cannot easily provide particles from more than 1 event now, it might be able to do so a few years down the road. So it is worthwhile to consider relaxed requirements for doing R&D work. And we need the bigger batch sizes and thus capacities to explore how to utilize GPUs efficiently.

I must have missed that part and can't find it in the slides. And we do have a solution for the utilization problem *today*: use multiple threads on the CPU to submit work to the GPU(s). That is realistic because Geant4 can handle one event per each thread, and their EM parts together can keep the GPU busy. That doesn't need any of this unrealistic increase of capacity / memory / batch size.

bernhardmgruber:
> I must have missed that part and can't find it in the slides.

I think I mentioned it around slide 20 when I discussed the separate interactions kernel implementation.

> And we do have a solution for the utilization problem today: use multiple threads on the CPU to submit work to the GPU(s). That is realistic because Geant4 can handle one event per each thread, and their EM parts together can keep the GPU busy. That doesn't need any of this unrealistic increase of capacity / memory / batch size.

There are of course multiple solutions to get the GPU busy. I am just trying a different approach, where AdePT can just buffer up tracks from many events and thus provide bigger batches to the GPU. I know that that does not fit Geant4's current workflow, but then maybe Geant4 also needs to adapt at some point if this route proofs to be successful. We are doing R&D after all and we should explore different strategies.

> Another argument against this kind of dynamic sizing is that it will lead to slightly different simulations on different machines with different amounts of memory when magnetic field propagation is enabled and we potentially kill particles at the tails...

That may be and I don't have the amount of in-depth knowledge on magnetic field propagation to know whether there is a solution to that. In my mental model, I can process tracks independently from each other so it does not matter for the correctness or determinism whether I put 100 or 10'000 of them on the GPU at the same time. But it matters for the throughput.

Anyway, I will move my effort to the new example created by @hageboeck now in #216.

hahnjo:
> > And we do have a solution for the utilization problem today: use multiple threads on the CPU to submit work to the GPU(s). That is realistic because Geant4 can handle one event per each thread, and their EM parts together can keep the GPU busy. That doesn't need any of this unrealistic increase of capacity / memory / batch size.
> 
> There are of course multiple solutions to get the GPU busy. I am just trying a different approach, where AdePT can just buffer up tracks from many events and thus provide bigger batches to the GPU. I know that that does not fit Geant4's current workflow, but then maybe Geant4 also needs to adapt at some point if this route proofs to be successful. We are doing R&D after all and we should explore different strategies.

Sure, but this topic was *again* discussed last week during the Celeritas visit, and right now all data clearly shows that multiple threads with reasonably sized batches of particles (within expectations of LHC experiments) is actually *more* efficient. I'm pasting some numbers for TestEm3MT on my 2070 SUPER that I also showed last week:
```
-particles 10000 -batch 1000 -threads 1
  16.5937

-particles 10000 -batch 500 -threads 2
  15.9731

-particles 10000 -batch 250 -threads 4
  15.7286


-particles 10000 -batch 400 -threads 1
  16.6332

-particles 10000 -batch 200 -threads 2
  15.9354

-particles 10000 -batch 100 -threads 4
  15.7649


-particles 10000 -batch 200 -threads 1
  17.1891

-particles 10000 -batch 100 -threads 2
  16.3435

-particles 10000 -batch 50 -threads 4
  16.2854


-particles 10000 -batch 100 -threads 1
  18.5353

-particles 10000 -batch 50 -threads 2
  17.3869

-particles 10000 -batch 25 -threads 4
  17.8254

================================================================================

-particles 10000 -batch 100 -threads 1
  18.5353

-particles 10000 -batch 100 -threads 2
  16.3435

-particles 10000 -batch 100 -threads 4
  15.7649

-particles 10000 -batch 100 -threads 8
  15.6614
```

I'd like to point out that this is without *any* hadronic interactions, which we know scales linearly with more threads in Geant4.

bernhardmgruber:
> Sure, but this topic was again discussed last week during the Celeritas visit, and right now all data clearly shows that multiple threads with reasonably sized batches of particles (within expectations of LHC experiments) is actually more efficient.

That is a good insight to have! But the difference between the various thread counts is not that big and this may shift due to subsequent optimizations of the GPU kernels. Nonetheless, I think we should also have support for multiple threads in the `SplitKernels` example eventually.

PR #215, opened 2022-07-07
Allow inlining FlatWrapper()/FlatArrayWrapper(), but not Advance()

bernhardmgruber:
Runtime (5x mean):

TestEm3:
master: 2.6339
PR: 2.40765

example18:
master: 1.23815
PR: 1.21754

### DIFF
--- a/base/inc/CopCore/include/CopCore/Ranluxpp.h
+++ b/base/inc/CopCore/include/CopCore/Ranluxpp.h
@@ -53,7 +53,8 @@ public:
   RanluxppEngineImpl() = default;
 
   /// Produce next block of random bits
-  __host__ __device__ void Advance()
+  // bgruber: this call is expensive and requires a lot of registers. Avoiding inlining here makes the code faster.
+  __host__ __device__ __attribute__((noinline)) void Advance()
   {
     uint64_t lcg[9];
     to_lcg(fState, fCarry, lcg);
--- a/examples/Example13/example13.cuh
+++ b/examples/Example13/example13.cuh
@@ -57,11 +57,11 @@ extern __constant__ __device__ int Zero;
 
 class RanluxppDoubleEngine : public G4HepEmRandomEngine {
   // Wrapper functions to call into RanluxppDouble.
-  static __host__ __device__ __attribute__((noinline)) double FlatWrapper(void *object)
+  static __host__ __device__ double FlatWrapper(void *object)
   {
     return ((RanluxppDouble *)object)->Rndm();
   }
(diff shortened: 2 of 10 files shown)


hahnjo:
This may be faster, but it's not correct. Please see https://github.com/apt-sim/AdePT/pull/115 for the rationale and talk to me, I can oftentimes provide more details of *why* things are the way they are...

hahnjo:
I again deeply looked into the topic, and I can still reprodue the original crash that led to #115, meaning the compiler issue is still there and just waits to hit us again. Especially with `example19` from https://github.com/apt-sim/AdePT/pull/216, we now have kernels that use random numbers but are not clamped to the maximum register usage (because the interaction kernels don't call geometry stuff). At the same time, it appears we have accumulated a number of other changes that currently prevent the bug from kicking in. However, I would still disagree that the changeset is legit and am against merging it.

Regarding performance, first of all I can "only" observe 2% improvement for `TestEm3` from `master`. Second, most of that seems to come from *adding* `__attribute__((noinline))` to the `Advance()` method, which we can do and still keep the workaround from #115. Would that make sense?

bernhardmgruber:
I added the change to example19 as well:

`example19 -gdml_file ../examples/Example14/macros/testEm3.gdml -particles 1000` (5 run mean):
master: 2.38086s
PR: 2.28701s

bernhardmgruber:
> [...] However, I would still disagree that the changeset is legit and am against merging it.

Ok, then let's keep it in for now. I cannot prove the bug is gone for sure, so if you can reproduce it, then that is solid evidence.

> Regarding performance, first of all I can "only" observe 2% improvement for TestEm3 from master.

The biggest change is actually the size of the generated code, since `Advance` was inlined into a lot of places. This PR  changes the lines of disassembly of TransportElectrons from 17896 to 11176. That is my main motivation for this change.

> Second, most of that seems to come from adding __attribute__((noinline)) to the Advance() method, which we can do and still keep the workaround from https://github.com/apt-sim/AdePT/pull/115. Would that make sense?

Do you suggest then to not remove `__attribute__((noinline))` anywhere? Or can you think of a way to make the hack work with the currently proposed changes?

hahnjo:
> Do you suggest then to not remove `__attribute__((noinline))` anywhere? Or can you think of a way to make the hack work with the currently proposed changes?

Yes, we have to keep the `noinline` on all `FlatWrapper` and `FlatArrayWrapper` because we have to pretend to call *exactly* the same function that is behind the function pointer.

hahnjo:
Superseded by https://github.com/apt-sim/AdePT/pull/218

PR #216, opened 2022-07-07
Add a SplitKernels example

hageboeck:
- New example based on example 18
- Discrete interactions moved to dedicated kernels for tests, profiling and R&D

@bernhardmgruber


examples/SplitKernels/gammas.cu
```diff
@@ -110,122 +113,149 @@ __global__ void TransportGammas(Track *gammas, const adept::MParray *active, Sec
...
+  } else if constexpr (ProcessIndex == 2) {
+    // Invoke photoelectric process.
+    const double theLowEnergyThreshold = 1 * copcore::units::eV;
 
-      // Check the new gamma energy and deposit if below threshold.
-      if (newEnergyGamma > LowEnergyThreshold) {
-        currentTrack.energy = newEnergyGamma;
-        currentTrack.dir    = newDirGamma;
+    const double bindingEnergy = G4HepEmGammaInteractionPhotoelectric::SelectElementBindingEnergy(
+        &g4HepEmData, theMCIndex, soaData.gamma_PEmxSec[soaSlot], energy, &rnge);
```
hahnjo, line 216: `soaData.gamma_PEmxSec` is never set, so very likely this is reading uninitialized memory...

examples/SplitKernels/main.cu
```diff
@@ -314,43 +323,93 @@ void runGPU(int numParticles, double energy, int batch, const int *MCIndex_host,
...
+
+        IonizationPos<<<32, ThreadsPerBlock, 0, interactionStreams[1]>>>(
+            positrons.tracks, positrons.queues.currentlyActive, secondaries, positrons.queues.nextActive, globalScoring,
+            scoringPerVolume, positrons.soaData);
+        BremsstrahlungPos<<<128, ThreadsPerBlock, 0, positrons.stream>>>(
+            positrons.tracks, positrons.queues.currentlyActive, secondaries, positrons.queues.nextActive, globalScoring,
+            scoringPerVolume, positrons.soaData);
+        AnnihilationPos<<<8, ThreadsPerBlock, 0, interactionStreams[2]>>>(
+            positrons.tracks, positrons.queues.currentlyActive, secondaries, positrons.queues.nextActive, globalScoring,
+            scoringPerVolume, positrons.soaData);
```
hahnjo, line 376: same

examples/SplitKernels/main.cu
```diff
@@ -314,43 +323,93 @@ void runGPU(int numParticles, double energy, int batch, const int *MCIndex_host,
...
+                                                                secondaries, gammas.queues.nextActive, globalScoring,
+                                                                scoringPerVolume, gammas.soaData);
+        // About 10% of all gammas:
+        ComptonScattering<<<64, ThreadsPerBlock, 0, interactionStreams[1]>>>(
+            gammas.tracks, gammas.queues.currentlyActive, secondaries, gammas.queues.nextActive, globalScoring,
+            scoringPerVolume, gammas.soaData);
+        // About 15% of all gammas:
+        PhotoelectricEffect<<<64, ThreadsPerBlock, 0, interactionStreams[2]>>>(
+            gammas.tracks, gammas.queues.currentlyActive, secondaries, gammas.queues.nextActive, globalScoring,
+            scoringPerVolume, gammas.soaData);
```
hahnjo, line 411: same

examples/SplitKernels/main.cu
```diff
@@ -175,8 +175,13 @@ void runGPU(int numParticles, double energy, int batch, const int *MCIndex_host,
...
   COPCORE_CUDA_CHECK(cudaMemcpyToSymbol(MCIndex, &MCIndex_dev, sizeof(int *)));
 
+  cudaDeviceProp deviceProp;
+  cudaGetDeviceProperties(&deviceProp, 0);
+
   // Capacity of the different containers aka the maximum number of particles.
-  constexpr int Capacity = 1024 * 1024;
+  // Use 1/5 of GPU memory for each of e+/e-/gammas, leaving 2/5 for the rest.
+  const size_t Capacity = (deviceProp.totalGlobalMem / sizeof(Track) / 5);
+  const int MaxBlocks = deviceProp.maxGridSize[0];
```
hahnjo, line 184: For all the reasons that we discussed many times by now and I keep repeating (last time in https://github.com/apt-sim/AdePT/pull/214), I'm against doing this.

hahnjo (changes requested):
Can we call this `example19`, please, following the existing naming scheme?

examples/SplitKernels/gammas.cu
hageboeck: Ah, thanks! I was already aware that there was a mistake when applying the patch, but I didn't have the time yesterday to review the diff! You saved me some work here!

examples/SplitKernels/main.cu
```diff
@@ -314,43 +323,93 @@ void runGPU(int numParticles, double energy, int batch, const int *MCIndex_host,
...
         COPCORE_CUDA_CHECK(cudaEventRecord(electrons.event, electrons.stream));
-        COPCORE_CUDA_CHECK(cudaStreamWaitEvent(stream, electrons.event, 0));
+        COPCORE_CUDA_CHECK(cudaStreamWaitEvent(interactionStreams[0], electrons.event, 0));
+
+        IonizationEl<<<32, ThreadsPerBlock, 0, interactionStreams[0]>>>(
+            electrons.tracks, electrons.queues.currentlyActive, secondaries, electrons.queues.nextActive, globalScoring,
+            scoringPerVolume, electrons.soaData);
+            BremsstrahlungEl<<<128, ThreadsPerBlock, 0, electrons.stream>>>(
+            electrons.tracks, electrons.queues.currentlyActive, secondaries, electrons.queues.nextActive, globalScoring,
+            scoringPerVolume, electrons.soaData);
```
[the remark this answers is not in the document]
hageboeck: They are kind of arbitrary, indeed. To be tuned later. It's just some numbers that will allow Bernhard to run some tests.

examples/SplitKernels/main.cu
hageboeck: same

examples/SplitKernels/main.cu
hageboeck: same

examples/SplitKernels/main.cu
hageboeck: I disagree. I think it's reasonable to support as many tracks as the GPU can handle. Like this, you can test various batch sizes via the command line without having to recompile, and you push the out-of-memory crash as far out as possible.
It doesn't mean that you will always run with the maximum batch size.

hageboeck: If that's the issue, we can just pin the default size to 100, or another number that you can choose here.

examples/SplitKernels/main.cu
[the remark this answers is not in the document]
hahnjo: But already right now, it's not a general approach for a "demonstrator"...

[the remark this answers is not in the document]
hageboeck: I unsure what you're trying to say. I'm trying to get some code ready for Bernhard, so he gets unblocked. There should be no issue to change these numbers without affecting his work in a later commit.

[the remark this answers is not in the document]
bernhardmgruber: I don't mind any of these numbers for now. They can become tuning parameters and probably should be tuned. From my experience a low number of threads and a very high number of threads are generally inferior to a moderate number of threads when using grid strided loops. But this has to be figured out later and is probably even GPU specific.

examples/SplitKernels/main.cu
bernhardmgruber: I don't think these batch sizes are unrealistic in the future. They may require changes on the Geant4 workflow though, but that should not matter now. This new example is a playground for R&D on we are trying to get the most out of the GPU, knowing that our changes may not fit the current Geant4 workflow.

bernhardmgruber:
> Can we call this example19, please, following the existing naming scheme?

We purposely chose something different and there is precedence for it with `ECS`, `FisherPrice_*`, `Raytracer_Benchmark`, `TestEm3`, etc. It is also different compared to the other examples, which seem rather finished, in that we want to continuously apply patches and improvements to `SplitKernels`. We want a playground for fast evolution.

And generally, the name `example19` does not carry any semantic meaning. It is already confusing enough that many other examples also carry no names indicating what they do, and they also not necessarily share a chronology. I would even propose to rename the existing examples as well so they include names like `Geant4_integration` or `standalone`, etc.

hahnjo:
`example19` is the naming scheme for the next example, and I'd like to ask that you follow that. What would you do if you have another idea, to implement on top of split kernels? Creating a new `example20` or whatever the number will be is easy and cheap.

bernhardmgruber:
> example19 is the naming scheme for the next example, and I'd like to ask that you follow that.

As I pointed out, it is not the naming scheme for all examples. And IMO it is also not a great naming scheme.

> What would you do if you have another idea, to implement on top of split kernels?

Easy, if the change is strictly better we change `SplitKernels`. If it is a variant that is not better, like a compacted or double buffered version, we can name it e.g. `SplitKernels_compact`. We also have precedence for that with `TestEm3` and `TestEm3Compact`, which are definitely better names than `exampleX`.

hahnjo:
> As I pointed out, it is not the naming scheme for all examples.

Yes, and as I pointed out in the past, this was not supposed to happen.

examples/SplitKernels/main.cu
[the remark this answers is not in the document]
hahnjo: The point I'm trying to make here is that the numbers are *completely* independent of the number of particles currently in flight. But even making a guess based on those is going to be hard because, well, the processes are stochastic. For that reason, I'd propose to use `transportBlocks` here. This overestimates the number of blocks actually needed, but anything else is just tuned to a very specific scenario.

examples/SplitKernels/main.cu
hahnjo: Let me once again explain what I'm after: I want a realistic scenario be bested by default, and in a way that is comparable to existing examples.

examples/SplitKernels/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+1. Obtain safety unless the track is currently on a boundary.
+2. Determine physics step limit, including conversion to geometric step length according to MSC.
+3. Query geometry (or optionally magnetic field) to get geometry step length.
+4. Convert geometry to true step length according to MSC, apply net direction change and discplacement.
+5. Apply continuous effects; kill track if stopped.
+6. If the particle reaches a boundary, perform relocation.
+7. If not, and if there is a discrete process:
+ 1. Sample the final state.
+ 2. Update the primary and produce secondaries.
```
hahnjo, line 39: This needs to be updated...

examples/SplitKernels/electrons.cu
```diff
@@ -0,0 +1,435 @@
...
+    // Check if a delta interaction happens instead of the real discrete process.
+    if (G4HepEmElectronManager::CheckDelta(&g4HepEmData, theTrack, currentTrack.Uniform())) {
+      // A delta interaction happened, move on.
+      activeQueue->push_back(globalSlot);
+      continue;
+    }
+
+    // Also advance the current RNG state to provide a fresh round of random
+    // numbers after MSC used up a fair share for sampling the displacement.
+    currentTrack.rngState.Advance();
```
hahnjo, line 287: As I've tried to explain in the previous PR, this call to `Advance()` is not needed because it will be done by `Branch` in the interaction kernels.

bernhardmgruber:
> Can we call this example19, please, following the existing naming scheme?

By now I don't care anymore whether this is called SplitKernels or example19. I would just really like to see it merged.

@hahnjo, @hageboeck and me would like to continue to work on that example and merge multiple PRs on top, including very experimental changes. So it is important for us to have freedom in that example.

hahnjo:
I stand by my previous comments:
1. This should be called `example19`, following the established naming scheme. If there are smaller tweaks to the example maintaining the workflow, they can be done in this example in follow-up PRs. For bigger changes, just add a new example.
2. IMO at least the default batch size should be adjusted to provide comparable results to the existing examples (on second thought, I don't care so much about the capacity, but right now these two are linked).
3. The arbitrary choices in the launch configurations for the split kernels should take the number of particles currently in flight into account. For an easy start, I suggested just defaulting to `transportBlocks`.

hageboeck:
> I stand by my previous comments:
> 
> 1. This should be called `example19`, following the established naming scheme. If there are smaller tweaks to the example maintaining the workflow, they can be done in this example in follow-up PRs. For bigger changes, just add a new example.
> 2. IMO at least the default batch size should be adjusted to provide comparable results to the existing examples (on second thought, I don't care so much about the capacity, but right now these two are linked).
> 3. The arbitrary choices in the launch configurations for the split kernels should take the number of particles currently in flight into account. For an easy start, I suggested just defaulting to `transportBlocks`.

I believe all of the above are addressed now. @bernhardmgruber @hahnjo 
For the batch sizes I went for 52 in the end, as that's what example18 would do by default.

examples/Example19/SplitKernels.cuh
```diff
@@ -0,0 +1,259 @@
+// SPDX-FileCopyrightText: 2022 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+#ifndef SPLITKERNELS_CUH
+#define SPLITKERNELS_CUH
+
+#include "SplitKernels.h"
```
hahnjo, line 7: Please rename to `example19.{cuh,h}` (or `example.{cuh,h}` if you want something more generic)

hahnjo (approved):
Looks good from a code point of view; some minor last comments inline.

FWIW here are the results for `-particles 10000` (10k) on my RTX 2070 SUPER:
| | `example18` | `example19` |
| --- | --- | --- |
| `-batch 52` (default) | 102s | 117s |
| `-batch 209` | 54.6s | 63.8s |
| `-batch 419` | 45.5s | 49.5s |

examples/Example19/SplitKernels.cuh
hageboeck: Generic is better for (future) diffs, thanks!

hahnjo:
From https://github.com/apt-sim/AdePT/pull/203#discussion_r861498661
> Another approach (that I think we had in the past? not sure) is having queues for the discrete processes. That would even save us from determining which Track need to run in which of the (split) kernels.

I gave this a try, and it seems to be equivalent in terms of performance but we don't need that awfully templated `InteractionLoop`. Let me know if you want me to submit this as a PR...

PR #217, opened 2022-08-23
Update G4HepEm.

hageboeck:
This fixes a few uninintialised GPU pointers to suppress cuda warnings.


hahnjo (approved):

hahnjo:
For future reference: This update of G4HepEm changes results of the examples involving multiple scattering, as expected.

hahnjo:
So the MSC changes were, in fact, not fully correct; will be fixed by https://github.com/apt-sim/AdePT/pull/223

PR #218, opened 2022-08-25
Update G4HepEm and adapt RNG interface

hahnjo:
Instead of function pointers, provide inline implementations of the `G4HepEmRandomEngine` methods. This allows to remove the compiler hack from commit https://github.com/apt-sim/AdePT/commit/cd1b441c830211548bdabd66672add0ed330fd66 ("Work around underestimated register count") and enables more inlining of the RNG methods.

### DIFF
--- a/examples/Example10/example10.cuh
+++ b/examples/Example10/example10.cuh
@@ -53,42 +53,20 @@ struct Track {
   }
 };
 
-// Defined in example10.cu
-extern __constant__ __device__ int Zero;
+// Define inline implementations of the RNG methods for the device.
+// (nvcc ignores the __device__ attribute in definitions, so this is only to
+// communicate the intent.)
+inline __device__ double G4HepEmRandomEngine::flat()
+{
+  return ((RanluxppDouble *)fObject)->Rndm();
+}
 
-class RanluxppDoubleEngine : public G4HepEmRandomEngine {
-  // Wrapper functions to call into RanluxppDouble.
-  static __host__ __device__ __attribute__((noinline))
-  double FlatWrapper(void *object)
-  {
-    return ((RanluxppDouble *)object)->Rndm();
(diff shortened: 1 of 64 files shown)


examples/Example10/example10.cuh
```diff
@@ -53,42 +53,18 @@ struct Track {
...
-  // Wrapper functions to call into RanluxppDouble.
-  static __host__ __device__ __attribute__((noinline))
-  double FlatWrapper(void *object)
-  {
-    return ((RanluxppDouble *)object)->Rndm();
+// Define inline implementations of the RNG methods.
+inline double G4HepEmRandomEngine::flat()
+{
+  return ((RanluxppDouble *)fObject)->Rndm();
+}
```
bernhardmgruber, line 60: So IIUC, the trick is now that g4HepEm provides only the interface for `G4HepEmRandomEngine` and an implementation in a separate cpp file that is linked into g4HepEmRun. Since g4HepEm is compiled normally (not with CUDA), the implementation of `G4HepEmRandomEngine` will become host code.

Now AdePT defines the same member functions a second time, but as device code, because in the header for `G4HepEmRandomEngine` the functions are marked with `G4HepEmHostDevice` and when AdePT uses that header, the macro `G4HepEmHostDevice` includes `__device__` and AdePT only calls the functions from device code.

So we would end up with two implementations in the binary, the host one provided by g4HepEm (via the g4hepEmRun library) and the device one as provided by the AdePT example. I guess this is fine.


 would prefer to mark the functions __device__ to communicate the intent that this is really supposed to be device code (since the host version comes from g4hepEmRun)

bernhardmgruber (approved):

hahnjo:
CI failure is due to the Alpaka toy example, which I propose to remove in https://github.com/apt-sim/AdePT/pull/219

bernhardmgruber (approved):

hageboeck (approved):
Like the idea! 👍

PR #206, opened 2022-05-17
Update alpaka API usage

bernhardmgruber:
The alpaka API has changed some time ago. This PR makes AdePT's alpaka examples compile again.

PR #219, opened 2022-08-26
Delete toy examples

hahnjo:
These are very old and not useful anymore, plus the Alpaka build seems again broken in Jenkins.


agheata (commented):
For the Alpaka one, maybe keep its folder for a while more (for the sake of the Alpaka integration demonstration), but remove the folder from the compilation?

hahnjo:
We can always go back to the `git` history for that, no?

bernhardmgruber:
So the alpaka example failed to compiled on my machine for a while, since I had alpaka 0.9.0 installed. So I fixed it in: #206. Quite recently I noticed then that the alpaka version on LCG was old (0.4.0 IIRC), so I requested that to be updated to 0.9.0 as well. I know the update was released on the LCG nightlies, but I don't know yet whether it is part of an LCG release yet. Maybe that causes the problem.

hahnjo:
I can reproduce the build failure with the tag 0.9.0 *unless* I add `-Dalpaka_ACC_GPU_CUDA_ENABLE=ON` to the AdePT build. I don't know if that's needed now...

agheata:
OK, it seems that struggling to keep this compiling does not worth it. However, the code is not big so if we just keep it un-compiled would make it easier to revive it at some point when we'll eventually be able to use Alpaka in more examples, without having to dig it out from history. So I vote for that.

bernhardmgruber:
> I can reproduce the build failure with the tag 0.9.0 unless I add -Dalpaka_ACC_GPU_CUDA_ENABLE=ON to the AdePT build. I don't know if that's needed now...

Oh yes, that needs to be enabled. Alpaka changed enabling backends by default at some time and now backends must be explicitly enabled by the user. Since the alpaka fisherprice example makes explicit use of types from the CUDA backend, e.g. `alpaka::rand::engine::uniform_cuda_hip::Xor`, the backend must be turned on to successfully compile the example.

I am sorry I did not notice this extra step earlier, I am already so used to this behavior of alpaka.

I would suggest to enable the backend explicitly in the `CMakeLists.txt` of the alpaka fisherprice example.

bernhardmgruber:
Let's give it a try: #221

PR #224, opened 2022-09-30
Remove unnecessary store and unused variable in example19

bernhardmgruber:


examples/Example19/electrons.cu
```diff
@@ -87,8 +87,7 @@ static __device__ __forceinline__ void TransportElectrons(Track *electrons, cons
     for (int ip = 0; ip < 3; ++ip) {
       double numIALeft = currentTrack.numIALeft[ip];
       if (numIALeft <= 0) {
-        numIALeft                  = -std::log(currentTrack.Uniform());
-        currentTrack.numIALeft[ip] = numIALeft;
```
bernhardmgruber, line 91: This store will be overwritten again later.

examples/Example19/gammas.cu
```diff
@@ -51,8 +51,7 @@ __global__ void TransportGammas(Track *gammas, const adept::MParray *active, Sec
     for (int ip = 0; ip < 3; ++ip) {
       double numIALeft = currentTrack.numIALeft[ip];
       if (numIALeft <= 0) {
-        numIALeft                  = -std::log(currentTrack.Uniform());
-        currentTrack.numIALeft[ip] = numIALeft;
```
bernhardmgruber, line 55: Same here.

hahnjo:
True, the first store of `numIALeft` is not needed - maybe the second was only executed conditionally at some point? Not sure...

But this improvement applies to all examples, no?

hahnjo:
> > But this improvement applies to all examples, no?
> 
> It probably does. Do you want me to apply this change set to other examples as well? Which ones?

The most recent examples probably, that we run comparisons against: `example13` and `example18` (in addition to `example19`) for the standalone cases, and `example14` for the integration example (maybe also `example17`, but we know that this one is broken so it should not be used anyways...)

bernhardmgruber:
I applied the change to the other suggested examples as well.

hahnjo (approved):

PR #227, opened 2022-10-04
[ex19] Allow for placing the particle gun from the host.

hageboeck:
For TestEm3, the gun needs to be placed at negative x. When "estEm3" is encountered in the gdml name, the gun will shift itself.

@bernhardmgruber 
```
Run time: 0.153657

Mean energy deposit          9.87835 GeV <--
Mean number of gamma         5199.97
```
:+1:

### DIFF
--- a/examples/Example19/main.cpp
+++ b/examples/Example19/main.cpp
@@ -194,7 +194,15 @@ int main(int argc, char *argv[])
   OPTION_DOUBLE(energy, 10); // entered in GeV
   energy *= copcore::units::GeV;
   OPTION_INT(batch, 52);
+  OPTION_STRING(gunpos, "0,0,0");
+  OPTION_STRING(gundir, "1,0,0");
   OPTION_BOOL(rotatingParticleGun, false);
+
+  // parse gun config
+  GunConfig gunConfig{};
+  std::sscanf(gunpos.c_str(), "%f,%f,%f", &gunConfig.position[0], &gunConfig.position[1], &gunConfig.position[2]);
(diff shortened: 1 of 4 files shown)


examples/Example19/main.cpp
```diff
@@ -254,9 +254,18 @@ int main(int argc, char *argv[])
   scoringPerVolume.energyDeposit      = energyDeposit;
   GlobalScoring globalScoring;
 
+  const bool testEm3 = gdml_file.find("estEm3") != std::string::npos;
```
hahnjo, line 257: Matching for a specific name is very, very ugly. Can we make this more general, please?

bernhardmgruber: Like I said on mattermost, I would just pass the values via command line arguments, like: `example19 -gunpos -220,0,0 -gundir 1,0,0`. This way we don't need to hardcode anything.

bernhardmgruber (approved):

examples/Example19/main.cpp
hageboeck: Better to put it in the readme. The defaults are correct for CMS.

bernhardmgruber: Added to readme.

bernhardmgruber:
Hmm, it seems I messed up the squashing of the git commit message :S

hahnjo:
It's a weird behavior to push commits to a PR and then accept and merge it yourself...

bernhardmgruber:
> It's a weird behavior to push commits to a PR and then accept and merge it yourself...

I agree, that's why I had @hageboeck have a look as well!

PR #228, opened 2022-10-07
Mark a track's survival more explicit

bernhardmgruber:
Adds a lambda function `survive()` which pushes the track back into the active queue.

Running `example19 -particles 10000 -batch 5000 -gdml_file ../examples/Example14/macros/testEm3.gdml` 5x on V100:

before:
Mean: 4.00011
Stddev: 0.00030375

after:
Mean: 4.01492
Stddev: 0.000460478

Output is identical. The slight slowdown is surprising, but it is by far compensated by the subsequent optimizations based on this PR, See #225 and #229.


bernhardmgruber:
[One of the CI runs](https://lcgapp-services.cern.ch/spi-jenkins/job/AdePT-CI/345/console) is failing with:
```
+ /usr/bin/nvidia-smi
Failed to initialize NVML: Driver/library version mismatch
```
I don't have experience with the CI. Can someone please have a look and fix it? Thx!

hahnjo (approved):

PR #229, opened 2022-10-07
Put RNG in shared memory where beneficial

bernhardmgruber:
The shared memory hardware can handle the access pattern on the RNG better than the local memory (in case of spilling) or global memory. For most kernels, it is thus beneficial to move the RNG into shared memory at the beginning of the kernel and store it back at the end in case the track survives.


hahnjo:
> The benefit in the interaction kernels is significantly smaller. Furthermore, contention for the shared memory is increased, since SM is also needed to compact active tracks at the beginning of the interaction kernels.

As mentioned in https://github.com/apt-sim/AdePT/pull/216#issuecomment-1235260059, we can get rid of this super complex "compaction in shared memory" code and just use queues... Let me know if you want me to push my WIP code somewhere.

bernhardmgruber:
> > The benefit in the interaction kernels is significantly smaller. Furthermore, contention for the shared memory is increased, since SM is also needed to compact active tracks at the beginning of the interaction kernels.
> 
> As mentioned in [#216 (comment)](https://github.com/apt-sim/AdePT/pull/216#issuecomment-1235260059), we can get rid of this super complex "compaction in shared memory" code and just use queues... Let me know if you want me to push my WIP code somewhere.

Sure, please push your changes somewhere and share them with us!

bernhardmgruber:
@agheata it was totally worth it for example19 IIRC. It may also be useful for other examples as well. I don't have the time to look into this anymore now that I am in the middle of my PhD writeup. So I leave this PR as a suggestions to you. Continue with it as you please.

PR #234, opened 2022-11-01
Add new Example20 with interaction queues instead of SOAData

hahnjo:
This revives an old concept from the toy examples 1, 2, and 4 (deleted in commit 9230c4a6c5) to have one queue per discrete interaction. This appears to be equally fast as the quite complex scheduling kernel based on SOAData.

@bernhardmgruber see here for not using `__shared__` memory in Example19 as mentioned here: https://github.com/apt-sim/AdePT/pull/229#issuecomment-1273021178


hageboeck:
Great to see that this works, thanks. Have you considered creating an example20 out of this, and document the difference to ex19 as was done for the other examples?

hahnjo:
> Great to see that this works, thanks. Have you considered creating an example20 out of this, and document the difference to ex19 as was done for the other examples?

Yes, and I can do so if you want. However, there is no difference in the executed workflow, it's just implemented differently. In the past, we did these "technical improvements" within existing examples...

hageboeck:
> Yes, and I can do so if you want. However, there is no difference in the executed workflow, it's just implemented differently. In the past, we did these "technical improvements" within existing examples...

Yes, please do so. Thanks!

hahnjo:
> > Yes, and I can do so if you want. However, there is no difference in the executed workflow, it's just implemented differently. In the past, we did these "technical improvements" within existing examples...
> 
> Yes, please do so. Thanks!

Done. Would be interesting to know what you have in mind with `example19` using the `SOAData`...

agheata (approved):

PR #235, opened 2022-11-02
Recompute macroscopic cross-section in example19

hahnjo:
Instead of storing to memory, just redo the computation in case a photoelectric interaction was selected. No performance difference, but less memory usage.


agheata (approved):
Makes sense to me

hageboeck (approved):
Sure, let's try this.

PR #240, opened 2023-03-24
Added surface navigator and made TestEm3 use it on demand

agheata:
This adds `SurfNavigator.h` allowing to dispatch the AdePT navigation interfaces to the VecGeom surface model. The surface model is currently available in the `surface_model` VecGeom branch, exposing  `VecGeom_SURF_FOUND` as cmake option. This is detected by AdePT cmake and only used if the user option `ADEPT_USE_SURF` is ON. AdePT surface model integration is optional and protected at build time.

In case `ADEPT_USE_SURF=ON` and the surface support is detected in VecGeom, the TestEm3 example will automatically use the SurfNavigator. I have validated statistically SurfNavigator against BVH navigator. The results are not numerically identical due to steps relocated by the solid models in between layers, generating subsequent zero steps but changing the random sequence. The surface model does not see 'holes' between layers, and does not need pushes for relocation.

Here are the preliminary performance measurements, comparing different navigators:
100K electrons, no field
- BVH: 152 sec
- Loop: 162 sec
- Surf: 156 sec

100K electrons, Bz = 1 Tesla
- BVH: 194 sec, 35 killed particles
- Loop: 1818 sec, 85 killed particles
- Surf: 184 sec, 27 killed particles

NB: that the surface model contains no BVH optimizations at this point

EDEP relative difference TestEm3 100K electrons surface model vs  BVH (no field)

EDEP relative difference TestEm3 100K electrons surface model vs  BVH (Bz = 1 Tesla)


base/inc/AdePT/SurfNavigator.h
```diff
@@ -0,0 +1,94 @@
+// SPDX-FileCopyrightText: 2020 CERN
```
hahnjo, line 1:
```suggestion
// SPDX-FileCopyrightText: 2023 CERN
```

hahnjo (approved):
LG, modulo some minor comments.

The results indeed seem to be almost identical because a rounding difference in one particle only "infects" its children, but all other particles continue to be propagated unchanged.

PR #242, opened 2023-04-14
Fixed raytrace benchmark to work with current VecGeom master

agheata:
VecGeom commit [8359663a](https://gitlab.cern.ch/VecGeom/VecGeom/-/commit/8359663a625c3d6e9fdcd62cbcede1a3b1ca7d68) removed the user hooks in LogicalVolume, which were used in the ray-tracer benchmark for storing custom optical properties. This commit replaces the hooks with an array using volume indices to retrieve the same information.

### DIFF
--- a/examples/Raytracer_Benchmark/Material.h
+++ b/examples/Raytracer_Benchmark/Material.h
@@ -22,7 +22,7 @@ void SetMaterialStruct(MyMediumProp *volume_container, std::vector<vecgeom::Logi
   int i = 0;
 
   for (auto lvol : logicalvolumes) {
-    // lvol->Print();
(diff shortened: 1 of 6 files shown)


examples/Raytracer_Benchmark/Material.h
```diff
@@ -22,7 +22,7 @@ void SetMaterialStruct(MyMediumProp *volume_container, std::vector<vecgeom::Logi
   int i = 0;
```
WitekPokorski, line 22: Is this line still needed?

agheata: No, leftover...

PR #244, opened 2023-04-18
Exa17 fixed and numerically compatible with Exa14.

agheata:
Added all missing fixes compared to Exa14, pdg bug and sorting by energy.
The compacting operation on the track container invalidated in the previous implementation the queues of leaked track references. Now leaked tracks are extracted and copied by the transport kernels.

This example uses much less memory that Exa14, and gives a large margin for playing with the GPU buffer size (holding the tracks coming from Geant4 to be showered on the GPU). The plot below shows the dependence of the single-threaded CPU time for cms2018, shooting an event with 10K electrons at 10 GeV.  This is unrealistic, but shows the impact of the event size on the performance in the current approach (small events produce smaller buffers).

Example17  10K electrons @ 10 GeV

Interesting enough, the time per primary does not depend on the event size as I expected:
10K elec, 1K elec and 100 elec


agheata:
@hahnjo thanks for finding the synchronization bug, this is now ready for review

examples/Example17/src/EMShowerModel.cc
```diff
@@ -27,14 +27,16 @@
 EMShowerModel::EMShowerModel(G4String aModelName, G4Region *aEnvelope)
     : G4VFastSimulationModel(aModelName, aEnvelope), fMessenger(new EMShowerMessenger(this))
 {
-  fRegion = aEnvelope;
+  auto tid = G4Threading::G4GetThreadId();
```
hahnjo, line 30: why do we need this?

examples/Example17/src/EMShowerModel.cc
```diff
@@ -27,14 +27,16 @@
...
+  fRegion  = aEnvelope;
 }
 
 //....oooOO0OOooo........oooOO0OOooo........oooOO0OOooo........oooOO0OOooo......
 
 EMShowerModel::EMShowerModel(G4String aModelName)
     : G4VFastSimulationModel(aModelName), fMessenger(new EMShowerMessenger(this))
 {
+  auto tid = G4Threading::G4GetThreadId();
```
hahnjo, line 39: same

hahnjo:
This seems to be missing the changes from https://github.com/apt-sim/AdePT/pull/208, could you please add them for testing with realistic events in MT mode?

agheata:
> This seems to be missing the changes from #208 (edit: and of course also the follow-up in #211), could you please add them for testing with realistic events in MT mode?

Good that you noticed :+1: , added now

hahnjo (approved):

PR #245, opened 2023-06-06
Benchmarking scripts

JuanGonzalezCaminero:
* BenchmarkManager class

* Example13 benchmark

* Example17 benchmark

### DIFF
--- a/base/inc/AdePT/BenchmarkManager.h
+++ b/base/inc/AdePT/BenchmarkManager.h
@@ -1,0 +1,202 @@
+// SPDX-FileCopyrightText: 2020 CERN
+// SPDX-License-Identifier: Apache-2.0
+
+/**
+ * @file BenchmarkManager.h
+ * @brief Benchmarking utilities: Timers, accumulators and formatted output.
+ */
+
+#ifndef BENCHMARK_H
+#define BENCHMARK_H
+
+#if defined BENCHMARK
+
+// Entry management
+#include <map>
+#include <string>
+// Timing
+#include <chrono>
+// Output
+#include <iostream>
+#include <fstream>
+#include <iomanip>
+// Output files and directories
+#include <filesystem>
+
+// The clock to use for measurements
+#define CLOCK std::chrono::system_clock
+// The precission with which we will provide the results
+#define PRECISSION std::chrono::microseconds
+
+/**
+ * @brief A type containing a timestamp, accumulated time, and whether the timer is running
+ * @details The timestamp marks the start of the current measuring period, the accumulated duration
+ * represents the total time measured under the same tag. It can also be set and added to manually
+ * to function as an accumulator.
+ */
+struct TimeInfo {
+  std::chrono::time_point<CLOCK> start;                                 ///< Start timestamp
+  std::chrono::duration<CLOCK::rep, CLOCK::period> accumulatedDuration; ///< Total time measured under the same tag
+  bool counting;                                                        ///< Whether the timer is running
+};
+
+template <class TTag>
+class BenchmarkManager {
+
+private:
+  std::map<TTag, TimeInfo> fTimers;     ///< Maps a tag to a TimeInfo struct
+  std::map<TTag, double> fAccumulators; ///< Maps a tag to a double accumulator
+  std::string fOutputDir;               ///< Output directory
+  std::string fOutputFilename;          ///< Output filename
+
+public:
+  BenchmarkManager(){};
+
+  ~BenchmarkManager(){};
+
+  /** @brief Sets a timestamp associated to this timer tag */
+  void timerStart(TTag tag)
+  {
+    auto &tagtimer    = fTimers[tag]; // Avoid counting the overhead from map access
+    tagtimer.counting = true;
+    tagtimer.start    = CLOCK::now();
+  }
+
+  /** @brief Compares the current time with the start of the timer and adds to the accumulated duration */
+  void timerStop(TTag tag)
+  {
+    if (fTimers[tag].counting) {
+      auto stopTimestamp = CLOCK::now(); // Avoid counting the overhead from map access
+      auto &tagtimer     = fTimers[tag];
+      tagtimer.counting  = false;
+      tagtimer.accumulatedDuration += stopTimestamp - tagtimer.start;
+    }
+  }
+
+  /** @brief Returns the accumulated duration for the specified tag in seconds */
+  double getDurationSeconds(TTag tag)
+  {
+    return (double)(std::chrono::duration_cast<PRECISSION>(fTimers[tag].accumulatedDuration)).count() /
+           PRECISSION::period::den;
+  }
+
+  /** @brief Sets the accumulator value for the specified tag */
+  void setAccumulator(TTag tag, double value) { fAccumulators[tag] = value; }
+
+  /** @brief Returns the accumulator value for the specified tag */
+  double getAccumulator(TTag tag) { return fAccumulators[tag]; }
+
+  /** @brief Adds to the accumulator for the specified tag */
+  void addToAccumulator(TTag tag, double value)
+  {
+    if (fAccumulators.find(tag) == fAccumulators.end()) {
+      // Initialize if the accumulator didn't exist
+      fAccumulators[tag] = value;
+    } else {
+      fAccumulators[tag] += value;
+    }
+  }
+
+  /** @brief Empties the timer map */
+  void reset() { fTimers.clear(); }
+
+  /** @brief Removes a timer from the map */
+  void removeTimer(TTag tag) { fTimers.erase(tag); }
+
+  /** @brief Checks if an timer is in the map */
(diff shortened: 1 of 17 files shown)


base/inc/AdePT/BenchmarkManager.h
```diff
@@ -0,0 +1,172 @@
...
+		void timerStart(TTag tag){}
+		void timerEnd(TTag tag){}
+    double getDurationSeconds(TTag tag){return 0;}
+    void addDurationSeconds(TTag tag, double duration){}
+    void reset(){}
+    void removeTimer(TTag tag){}
+    bool hasTimer(TTag tag){return false;}
+		void exportCSV(TTag filename){}
+    static bool isActive(){return false;}
+};
```
agheata, line 168: Why do you need a second implementation if BENCHMARK is not on? IMO the utility should always be the same and the protection should be put in a macro BENCHMARK_START/END(tag) that would compile to empty if BENCHMARK is not on. You can then protect per example if you want calling benchmarking on conditions.

examples/Example13/example13.cpp
```diff
@@ -196,30 +199,56 @@ int main(int argc, char *argv[])
...
+  while ((opt = getopt(argc, argv, "b:")) != -1) {
+    switch (opt) {
+      //-b [benchmark output filename]
+      case 'b':
+        benchmark_file = optarg;
+        break;
+      default:
+        ;
+    }
+  }
```
agheata, line 214: You should rather use OPTION_STRING from ArgParser.h

examples/Example13/example13.cpp
```diff
@@ -196,30 +199,56 @@ int main(int argc, char *argv[])
...
+  std::string path(argv[0]);
+  if(benchmark_file.empty()) benchmark_file = path.substr(path.find_last_of("/\\") + 1);
+
   vecgeom::Stopwatch timer;
   timer.Start();
 
   NVTXTracer tracer("InitG4");
+  BenchmarkManager<std::string> benchmark_manager;
+  benchmark_manager.timerStart("Total time");
+  benchmark_manager.timerStart("InitG4");
```
agheata, line 225: This can become:
```
BENCHMARK_INIT(); // macro creating the benchmarker if BENCHMARK is set
BENCHMARK_START("Total time");
BENCHMARK_START("InitG4");

File: examples/Example17/src/Run.cc [hunk 4]
```diff
@@ -0,0 +1,102 @@
...
+// ********************************************************************
+//
+#include "G4RunManager.hh"
+#include "G4Run.hh"
+#include "Run.hh"
+
+#include "BenchmarkManager.h"
+#include <cmath>
+
+#define STDEV(N, MEAN, SUM_SQUARES) N>1 ? sqrt((SUM_SQUARES-N*pow(MEAN,2))/N) : 0
```
agheata, line 35: MEAN*MEAN instead of pow(MEAN,2), cheaper

agheata (changes requested):
A first batch of comments, I may have more. Make sure you `clang-format -style=file` the modified files, add licenses and add comments on data members and class descriptions.

base/inc/AdePT/BenchmarkManager.h
```diff
@@ -0,0 +1,172 @@
...
+  {
+  	epoch = CLOCK::now();
+  };
+
+  ~BenchmarkManager(){};
+
+  //Sets a timestamp associated to this timer tag
+  void timerStart(TTag tag)
+  {
+    timers[tag].start = CLOCK::now();
```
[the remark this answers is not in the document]
JuanGonzalezCaminero: If we do it in this way we are including the overhead in the timing, I think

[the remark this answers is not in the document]
agheata: Right, to move the overhead outside the time measurement:
```
auto &tagtimer = timers[tag]; // this is overhead
tagtimer.start = CLOCK::now();
```

base/inc/AdePT/BenchmarkManager.h
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  //Compares the current time with the start of the timer and adds to the accumulated duration
+  void timerEnd(TTag tag)
+  {
+    timers[tag].accumulated_duration += (CLOCK::now() - timers[tag].start);
+  }
+
+  //Returns the accumulated duration for the specified tag in seconds, with millisecond precission
+  double getDurationSeconds(TTag tag)
```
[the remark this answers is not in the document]
JuanGonzalezCaminero: Apparently the map [] operator is not const, so I cannot declare the function as const either

examples/Example17/src/EventAction.cc
```diff
@@ -65,9 +77,33 @@ void EventAction::BeginOfEventAction(const G4Event *)
...
+    auto aBenchmarkManager = currentRun->getBenchmarkManager();
+    aBenchmarkManager->timerStop(Run::timers::EVENT);
+
+    //Accumulate the timings
+    double eventTime = aBenchmarkManager->getDurationSeconds(Run::timers::EVENT);
+    double nonEMTime = aBenchmarkManager->getDurationSeconds(Run::timers::NONEM_EVT);
+    double ecalTime = eventTime - nonEMTime;
+    
+    aBenchmarkManager->addDurationSeconds(Run::timers::EVENT_SUM, eventTime);
+    aBenchmarkManager->addDurationSeconds(Run::timers::EVENT_SQ, pow(eventTime, 2));
```
agheata, line 96: Avoid calling the power function for squared values, here it does not matter much, but in performance-critical code it may. It is not granted that the compilers will do the optimization for you [see here](https://stackoverflow.com/questions/2940367/what-is-more-efficient-using-pow-to-square-or-just-multiply-it-with-itself)

examples/Example17/include/Run.hh
```diff
@@ -0,0 +1,85 @@
...
+    // Acummulator for the sum and squared sum of the timings across all events
+    EVENT_SUM,
+    EVENT_SQ,
+
+    // Global execution timer
+    TOTAL
+  };
+
+private:
+  BenchmarkManager<TAG_TYPE> *fBenchmarkManager;
```
agheata, line 82: Is there a scenario where you need more than one benchmark manager? Because if you have only one, you can make it a singleton and get a static instance pointer from wherever you need it.

examples/Example17/src/SteppingAction.cc
```diff
@@ -0,0 +1,98 @@
...
+SteppingAction::~SteppingAction() {}
+
+void SteppingAction::UserSteppingAction(const G4Step *theStep)
+{
+
+  // Check if we moved to a new volume
+  G4VPhysicalVolume *previousVolume = theStep->GetPreStepPoint()->GetTouchableHandle()->GetVolume();
+  G4VPhysicalVolume *currentVolume  = theStep->GetPostStepPoint()->GetTouchableHandle()->GetVolume();
+
+  if (previousVolume != currentVolume) {
```
agheata, line 58: I'm not sure this works for all boundary crossings, there can be cases where the physical volume being crossed into is the same as the pre-step one, although in a different touchable. To be checked, but it is probably safer to use: `(theStep->GetPostStepPoint()->GetStepStatus() == G4StepStatus::fGeomBoundary). Not sure however if you need to get this from the post step point.

examples/Example17/include/Run.hh
agheata: We just discussed, you need this working in MT mode, so a singleton won't do

examples/Example17/src/SteppingAction.cc
```diff
@@ -0,0 +1,96 @@
...
+// * use.  Please see the license in the file  LICENSE  and URL above *
+// * for the full disclaimer and the limitation of liability.         *
+// *                                                                  *
+// * This  code  implementation is the result of  the  scientific and *
+// * technical work of the GEANT4 collaboration.                      *
+// * By using,  copying,  modifying or  distributing the software (or *
+// * any work based  on the software)  you  agree  to acknowledge its *
+// * use  in  resulting  scientific  publications,  and indicate your *
+// * acceptance of all terms of the Geant4 Software license.          *
+// ********************************************************************
```
agheata, line 26: Not needed

examples/Example17/include/Run.hh
```diff
@@ -0,0 +1,86 @@
...
+// * use.  Please see the license in the file  LICENSE  and URL above *
+// * for the full disclaimer and the limitation of liability.         *
+// *                                                                  *
+// * This  code  implementation is the result of  the  scientific and *
+// * technical work of the GEANT4 collaboration.                      *
+// * By using,  copying,  modifying or  distributing the software (or *
+// * any work based  on the software)  you  agree  to acknowledge its *
+// * use  in  resulting  scientific  publications,  and indicate your *
+// * acceptance of all terms of the Geant4 Software license.          *
+// ********************************************************************
```
agheata, line 26: Not needed

agheata (approved):

PR #251, opened 2023-06-21
Investigation on AdePT/G4 Discrepancy

JuanGonzalezCaminero:


agheata:
I think this is not needed anymore at this point, since the geometry bug was found/fixed

PR #253, opened 2023-06-28
Example17 Randomized particle gun

JuanGonzalezCaminero:
* Adds the option to shoot particles uniformly in the chosen Phi/Theta ranges
* Adds the option to define a list of particles to be shot, and choose one at random each time
* Includes a necessary extra check in the SteppingAction for particles going out of the world


examples/Example17/src/ParticleGun.cc
```diff
@@ -0,0 +1,48 @@
...
+#include "G4PrimaryParticle.hh"
+#include "G4Event.hh"
+#include "Randomize.hh"
+#include "G4SystemOfUnits.hh"
+
+void ParticleGun::GenerateRandomPrimaryVertex(G4Event *aEvent, G4double aMinPhi, G4double aMaxPhi, G4double aMinTheta,
+                                              G4double aMaxTheta, std::vector<G4ParticleDefinition *> *aParticleList)
+{
+  // Choose a particle from the list
+  G4ParticleDefinition *aParticleDefinition = (*aParticleList)[(int)(G4UniformRand() * aParticleList->size())];
```
agheata, line 16: Here you will need to take into account the fraction for each particle type. The typical algorithm does:
```
  auto rand = G4UniformRand();
  size_t ipart     = 0;
  float weight = fractionsList[ipart];
  while (rand > weight && ipart < aParticleList->size())
    weight +=  fractionsList[++ipart];
  assert(ipart < aParticleList->size());
```

examples/Example17/include/PrimaryGeneratorAction.hh
```diff
@@ -54,12 +55,18 @@ public:
   void SetDefaultKinematic();
   void SetRndmBeam(G4double val) { fRndmBeam = val; }
```
agheata, line 56: You may need to set a seed for the random number generator, and a function to initialize the generator from this seed. We need to be able to produce reproducible but different sequences, and to replay the same sequence for G3ant4/AdePT

examples/Example17/src/ParticleGun.cc
```diff
@@ -0,0 +1,48 @@
...
+  auto *vertex = new G4PrimaryVertex(particle_position, particle_time);
+
+  // Create new primaries and set them to the vertex
+  //
+  G4double mass = particle_definition->GetPDGMass();
+  for (G4int i = 0; i < NumberOfParticlesToBeGenerated; ++i) {
+    auto *particle = new G4PrimaryParticle(particle_definition);
+
+    G4double phi                = (aMaxPhi - aMinPhi) * G4UniformRand() + aMinPhi;
+    G4double theta              = (aMaxTheta - aMinTheta) * G4UniformRand() + aMinTheta;
```
agheata, line 30: Here you generate uniform in theta, but for a isotropic distribution (same probability per solid angle) you need to generate uniform in cos(theta):
```
  auto theta = std::acos(std::cos(aMinTheta) + G4UniformRand() * (std::cos(aMaxTheta) - std::cos(aMinTheta)));
```

examples/Example17/src/PrimaryGeneratorAction.cc
```diff
@@ -47,45 +50,49 @@ void PrimaryGeneratorAction::GeneratePrimaries(G4Event *aEvent)
...
+    // randomize the beam, if requested.
+    if (fRndmBeam > 0.) {
+      G4ThreeVector oldPosition = fParticleGun->GetParticlePosition();
+      G4double rbeam            = 0.5 * fRndmBeam;
+      G4double x0               = oldPosition.x();
+      G4double y0               = oldPosition.y() + (2 * G4UniformRand() - 1.) * rbeam;
+      G4double z0               = oldPosition.z() + (2 * G4UniformRand() - 1.) * rbeam;
+      fParticleGun->SetParticlePosition(G4ThreeVector(x0, y0, z0));
+      fParticleGun->GeneratePrimaryVertex(aEvent);
+      fParticleGun->SetParticlePosition(oldPosition);
```
agheata, line 87: This seems to override the randomized gun position. So we need to make sure that we randomize the position for all types of gun in case ` fRndmBeam > 0`

agheata (changes requested):

PR #255, opened 2023-07-12
Use G4endl with G4cout

hahnjo:
This gives the proper `G4WT` prefixes.

### DIFF
--- a/examples/Example14/src/PrimaryGeneratorAction.cc
+++ b/examples/Example14/src/PrimaryGeneratorAction.cc
@@ -108,5 +108,5 @@ void PrimaryGeneratorAction::Print() const
 {
   G4cout << "=== Gun shooting " << fParticleGun->GetParticleDefinition()->GetParticleName() << " with energy "
          << fParticleGun->GetParticleEnergy() / GeV << "[GeV] from: " << fParticleGun->GetParticlePosition() / mm
-         << " [mm] along direction: " << fParticleGun->GetParticleMomentumDirection() << "\n";
+         << " [mm] along direction: " << fParticleGun->GetParticleMomentumDirection() << G4endl;
 }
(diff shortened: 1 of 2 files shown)


agheata (approved):

agheata:
Actually the build failed due to:
```
CMake Error at examples/Example3/CMakeLists.txt:39 (add_executable):
  CMake doesn't support IPO for current compiler
```
This is due to this line in example3 cmake:
`set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)`
It looks like cmake changed the policy to generate an error if the compiler does not support it, as per: https://cmake.org/cmake/help/latest/policy/CMP0069.html

@hahnjo do you mind looking into this?

hahnjo:
> Actually the build failed due to:
> 
> ```
> CMake Error at examples/Example3/CMakeLists.txt:39 (add_executable):
>   CMake doesn't support IPO for current compiler
> ```
> 
> This is due to this line in example3 cmake: `set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)` It looks like cmake changed the policy to generate an error if the compiler does not support it, as per: https://cmake.org/cmake/help/latest/policy/CMP0069.html

Sure, but this is not related to this pull request, is it?

agheata:
> Sure, but this is not related to this pull request, is it?

No, this PR triggered it.

hahnjo:
> > Sure, but this is not related to this pull request, is it?
> 
> No, this PR triggered it.

Not really, it was already there yesterday: http://cdash.cern.ch/index.php?project=AdePT&date=2023-07-12 (it also wouldn't make sense for this change to trigger a CMake problem...)

agheata:
I meant it revealed it, it was triggered probably by an upgrade of cmake on the test machines. Did not notice it was there in the nightlies. The question is, we fix it by removing the cmake line in the example, or applying IPO only where available? It is easy just to remove, but maybe we will need it in future for other examples? What is your take?

hahnjo:
I don't know, I don't think we need it right now. But `example3` is by Guilherme, so I don't really know why it is there in the first place.

agheata:
@amadio do we need this CMAKE_INTERPROCEDURAL_OPTIMIZATION option in example3?

PR #257, opened 2023-07-19
[Example3] disabled IPO due to CMake change of policy.

agheata:
CI is currently broken due to this


agheata:
We need this to unbreak the CI, can someone have a look?

hahnjo (approved):
Not fully sure why it breaks when going from CMake 3.23.2 to 3.26.2, CMP0069 was already there before and the documentation states that `CMAKE_INTERPROCEDURAL_OPTIMIZATION` may fail since CMake 3.9...

I also think the code could be completely removed; as far as I understand, it only worked with Intel compilers?

PR #260, opened 2023-08-10
Fix compilation issue in example17

JuanGonzalezCaminero:


hahnjo:
@JuanGonzalezCaminero did you see https://github.com/apt-sim/AdePT/pull/259?

PR #262, opened 2023-09-12
BVHNavigator fix, passing incorrect last volume to BVH

JuanGonzalezCaminero:


hahnjo:
This change doesn't look good for 1 million 10 GeV electrons in TestEm3:

fig_CaloEDep

agheata:
We know that the fix does not work in this way, closing this PR

PR #261, opened 2023-09-11
Draft BVH Refactoring

JuanGonzalezCaminero:
* Adapting the BVHNavigator to work with the modified BVH in VecGeom
* Validation done:
  * Example17, 1 event, 1000 e-, Gun 10-170 in Theta, 0-360 in Phi: Numerically identical to master
  * TestEM3: Numerically identical to master
* Performance comparison:
  * Example17, 64 events, 1000 e-, Gun 10-170 in Theta, 0-360 in Phi.
  * Average Event time ratio new/master : 1.00081
  * Average ECAL time ratio new/master : 1.00423
  * Average Non EM time ratio new/master: 0.99945

PR #264, opened 2023-10-13
Navigation refactoring moving navigators in VecGeom

agheata:
This combines the VecGeom version change and the draft BVH refactoring, to have a coherent migration in one go:

* Moving the requested version of VecGeom to 2.0.0-dev.3, which contains the navigators moved from AdePT, and needed by #261. We need to use VecGeom_VERSION_STRING to select dev versions of VecGeom until the 2.0.0 release

* Adapting the BVHNavigator to work with the modified BVH in VecGeom (by @JuanGonzalezCaminero)
* Validation done:
  * Example17, 1 event, 1000 e-, Gun 10-170 in Theta, 0-360 in Phi: Numerically identical to master
  * TestEM3: Numerically identical to master
* Performance comparison:
  * Example17, 64 events, 1000 e-, Gun 10-170 in Theta, 0-360 in Phi.
  * Average Event time ratio new/master : 1.00081
  * Average ECAL time ratio new/master : 1.00423
  * Average Non EM time ratio new/master: 0.99945


base/inc/AdePT/BVHNavigator.h
```diff
@@ -9,352 +9,12 @@
...
-      }
-      assert(!state.Top()->GetLogicalVolume()->GetUnplacedVolume()->IsAssembly());
-    }
-  }
-};
+using BVHNavigator = vecgeom::BVHNavigator;
 
 } // End namespace COPCORE_IMPL
-#endif // RT_LOOP_NAVIGATOR_H_
+#endif // RT_LOOP_NAVIGATOR_H_
```
JuanGonzalezCaminero, line 20: RT_NAVIGATOR_H_

JuanGonzalezCaminero (approved):

base/inc/AdePT/BVHNavigator.h
agheata: OK we fix this in a separate PR not to re-trigger the CI now

agheata:
The crashes in examples when using cms_2018 are likely due to not increasing the stack limit, after a commit in VecGeom that was setting it to 8KB by default. Many examples are already protected, none is crashing on my machine or on the test machine, but we probably need to extend the protection. About the reproducibility and performance measurements, @JuanGonzalezCaminero can comment.

JuanGonzalezCaminero:
Hi Jonas, indeed the results are no longer numerically identical, the comparison was done before the commit moving the navigators to VecGeom. 
However we can't compare the results from VecGeom commit `804dcfad` to the current master because commit `0693dd8e` does produce differences in the results. The comparison with commit `44ec0798` does produce identical results, however we can no longer run it with AdePT since this PR doesn't include [This](https://github.com/apt-sim/AdePT/pull/261/commits/bafc400442ad17f7a5b91e00825aa1f084da027e) commit with the partial changes (Modified navigator but still in AdePT)

agheata:
> However we can't compare the results from VecGeom commit `804dcfad` to the current master because commit `0693dd8e` does produce differences in the results.

This is the fix for extruded solids non-symmetrical about XoY

agheata:
> However we can't compare the results from VecGeom commit `804dcfad` to the current master because commit `0693dd8e` does produce differences in the results.

This is the fix for extruded solids non-symmetrical about XoY

hahnjo:
> indeed the results are no longer numerically identical, the comparison was done before the commit moving the navigators to VecGeom.

Ok, but there are two sources of differences that should be understood individually:
1. Moving from `v1-patches` to `master` before the BVH refactoring - where does this come from?
2. Refactoring the BVH and moving the navigators from AdePT to VecGeom - if the code is moved 1:1, I would expect it to produce identical results. Why is that not the case?

> because commit `0693dd8e` does produce differences in the results

> This is the fix for extruded solids non-symmetrical about XoY

This commit touches PolygonalShell which is only used by simple extruded solids. There are no extruded solids in `cms2018` (they are used in the HGCal for Phase2), so I'm not sure I understand how this can be responsible for changing results?

> The comparison with commit 44ec0798 does produce identical results, however we can no longer run it with AdePT since this PR doesn't include [This](https://github.com/apt-sim/AdePT/pull/261/commits/bafc400442ad17f7a5b91e00825aa1f084da027e) commit with the partial changes (Modified navigator but still in AdePT)

Ok, I picked only that one commit and the BVH refactoring alone seems to give numerical results while staying with VecGeom commit https://gitlab.cern.ch/VecGeom/VecGeom/-/commit/44ec0798fcc7d689246fb214620e520382d6c294 Can run more tests next week, if needed.

agheata:
The only source of differences is **actually** coming from this merge request: https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/1004 (commit e607cbc72ac323c31747c0181c4e1ead886965ea) introducing operator*= for Transformation3D and applying it for cached transformations. This is one commit after the one fixing the simple extruded. The rationale was twofold: having numerical matching between transformations as computed in Geant4, and doing less copies while getting the top transformation in the non-cached mode. Basically, we replace the weird operation MultiplyFromRight with a normal matrix multiplication like in G4. However, due to the different order of operations, truncation is done differently so the final matrices and not numerically equal. There is a unit test that was implemented at the same time, and replacing the `ApproxEqual` with `==` makes it fail.

The move to master before the refactoring is probably an effect of my starting from Juan's branch, and adding the version change on top of it, I was rushing because the CI was broken by a premature VecGeom tag deployment.

hahnjo:
> The only source of differences is **actually** coming from this merge request: https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/1004 (commit e607cbc72ac323c31747c0181c4e1ead886965ea)

Alright, this explains the differences seen together with the refactoring of the BVH navigator and the move from AdePT to VecGeom. The difference between `v1-patches` to VecGeom commit `804dcfadd0253ba81a2875da99cf40a2007b3f94` is still a mystery to me...

hahnjo:
> The difference between `v1-patches` to VecGeom commit `804dcfadd0253ba81a2875da99cf40a2007b3f94` is still a mystery to me...

It's https://gitlab.cern.ch/VecGeom/VecGeom/-/commit/011010a2 which makes sense in retrospective, I guess...

agheata:
> > The difference between `v1-patches` to VecGeom commit `804dcfadd0253ba81a2875da99cf40a2007b3f94` is still a mystery to me...
> 
> It's https://gitlab.cern.ch/VecGeom/VecGeom/-/commit/011010a2 which makes sense in retrospective, I guess...

Ah, OK, yes, it is understandable now. Can we run the testEM3 and example17 validation vs. Geant4 to make sure that all these transformation changes (tested to alter the values of transformation elements at ~1E-15 level) do not affect physics validation?

PR #269, opened 2023-12-11
Use `PROJECT_{SOURCE,BINARY}_DIR` in place of `CMAKE_` equivalent

drbenmorgan:
When included as a subproject as `add_subdirectory` or similar, the `CMAKE_{SOURCE,BINARY}_DIR` variables will point to those of the highest level project. This will not be the same as the root of the AdePT source code. Include/etc paths will not then be correct, leading to compile time errors. Additionally, generated files may appear in the wrong place, or clash with higher level projects. This _may_ be one of the methods experiments use to integrate the code, so good to support it, and will also help co-working with Celeritas.

Use `PROJECT_{SOURCE,BINARY}_DIR` in place of the `CMAKE_` versions to guarantee pointing to the AdePT root source/build directories.

### DIFF
--- a/examples/Example12/CMakeLists.txt
+++ b/examples/Example12/CMakeLists.txt
@@ -39,4 +39,4 @@ set_target_properties(example12 PROPERTIES CUDA_SEPARABLE_COMPILATION ON CUDA_RE
 
 # Tests
 add_test(NAME example12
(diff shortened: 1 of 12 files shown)


agheata (approved):
great, thanks @drbenmorgan !

PR #271, opened 2024-01-19
Re-organising AdePT as a library

JuanGonzalezCaminero:
* Common headers are now in the AdePT/ directory
* Elements independent from AdePT but re-used across examples are now in /examples/common
* Adding new Example1, which links against the AdePT library, and contains only user-defined code which AdePT doesn't depend on
  * Example1 is based on the custom Tracking Manager integration strategy previously used in the legacy example 22
  * The custom Tracking Manager now allows to do region-based GPU tracking
* Removed most legacy examples
* Exporting three new libraries, AdePT, AdePT_cuda and AdePT_G4_integration
* CopCore is no longer an internal package and the library is also exported as AdePT::CopCore
* The AdePT core library doesn't depend on Geant4, only AdePT_G4_integration does
* G4HepEM is now an external dependency
* Adding time information in the AdePT Tracks, to be validated
* General cleanup


agheata (approved):
Great work @JuanGonzalezCaminero LGTM !

PR #282, opened 2024-03-08
Making AdePT a header-only library

JuanGonzalezCaminero:
This PR removes the circular dependency between AdePT and AdePT_G4_Integration by making AdePT a header-only library.
- AdePTTransport is now templated on the Integration layer used. The only library that is now compiled is AdePT_G4_integration.so
- The Scoring now consists on free functions templated on a "Scoring struct" containing the necessary information for the specific scoring implementation. New scoring approaches should be implemented through partial template specialization of the methods defined in AdePTScoringTemplate.cuh.
- No performance impact observed
- Small change in the total edep as a result of slight differences in the compiler optimizations
- TestEM3 has been removed as it was no longer interesting to keep it in it's previous form. It will be re-introduced in the future using the new way of integrating with AdePT
- Removed the option `--source-in-ptx` in Debug mode, as it was causing a compilation error in some configurations. It has been replaced with `-G`, which should still allow us to place breakpoints in the kernels


CMakeLists.txt
```diff
@@ -124,7 +124,7 @@ add_compile_options("$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:RelWithDebInfo>>:
 # - For Debug, generate full debug information - this completely disables optimizations!
 add_compile_options("$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<CONFIG:Debug>>:--device-debug>")
 # - For both, interleave the source in PTX to enhance the debugging experience.
-add_compile_options("$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<OR:$<CONFIG:RelWithDebInfo>,$<CONFIG:Debug>>>:--source-in-ptx>")
+# add_compile_options("$<$<AND:$<COMPILE_LANGUAGE:CUDA>,$<OR:$<CONFIG:RelWithDebInfo>,$<CONFIG:Debug>>>:--source-in-ptx>")
```
agheata, line 127: Maybe mention in a comment above that this option is broken for some configurations.

include/AdePT/core/AdePTTransport.cuh
```diff
@@ -1,100 +1,458 @@
...
+{
+  all.leakedElectrons->clear();
+  all.leakedPositrons->clear();
+  all.leakedGammas->clear();
+}
+
+bool InitializeField(double bz)
+{
+  // Try 16384 if debug mode is crashing
+  COPCORE_CUDA_CHECK(vecgeom::cxx::CudaDeviceSetStackLimit(8192 * 2));
```
agheata, line 192: Not sure this belongs here, but OK for now. This should be a configuration parameter in future.

agheata (approved):

include/AdePT/core/AdePTTransport.cuh
JuanGonzalezCaminero: Yes, this one has been moved around quite a bit, we need to figure out the best place

PR #284, opened 2024-03-18
Host scoring optimization

JuanGonzalezCaminero:
Small optimization when scoring in the host, we only initialize the objects needed for scoring once, rather than once per group of hits


include/AdePT/integration/AdePTGeant4Integration.hh
```diff
@@ -69,6 +69,16 @@ private:
...
+
+  bool fScoringObjectsInitialized{false};
+  G4NavigationHistory *fPreG4NavigationHistory;
+  G4NavigationHistory *fPostG4NavigationHistory;
+  G4Step *fG4Step;
+  G4TouchableHandle fPreG4TouchableHistoryHandle;
+  G4TouchableHandle fPostG4TouchableHistoryHandle;
+  G4Track *fElectronTrack;
+  G4Track *fPositronTrack;
+  G4Track *fGammaTrack;
```
agheata, line 81: I don't see who deletes this data, is it going to leak at the end? Why not implement a proper destructor?

agheata (changes requested):
Cleanup of data members is needed

agheata (approved):

PR #285, opened 2024-03-22
Use VecGeom recommendations for device linking

drbenmorgan:
As outlined in VecGeom README and build scripts, we have to be _very_ careful about linking to its CUDA component when we ourselves use VecGeom's CUDA functions. This is a small update to AdePT to use these recommendations:

1. Linking to `vecgeomcuda` privately, adding `vecgeomcuda_static` to the device link step.
2. Use `CUDA_RUNTIME_LIBRARY` as "Shared" to link to same runtime as used by `vecgeomcuda`

The public link to `vecgeom` is retained currently due to `AdePT_G4_integration` still exposing some of VecGeom in its interfaces, which is fine for now.

This change is required to allow use of G4VG for in memory use of Geant4-VecGeom conversions, but I've submitted just the linking change in isolation first to cross-check if this exposes `as-needed` (it doesn't locally at least) or performance issues.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -163,19 +163,23 @@ target_link_libraries(AdePT_G4_integration
   PUBLIC 
     CopCore
     VecGeom::vecgeom
-    VecGeom::vecgeomcuda_static
-    VecGeom::vgdml
     ${Geant4_LIBRARIES}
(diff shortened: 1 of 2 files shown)


drbenmorgan:
Closing as superseded by #289

PR #290, opened 2024-05-03
CI: add possibility to switch operating system

andresailer:
In view of the EoL of centos7, this allows one to move to alma9. 

* The GPU machine on alma9 has the Tesla4-el9 label. (The other one has both TeslaT4 and TeslaT4-centos7, so one isn't accidentaly using the wrong machine anymore)
* There is now `/cvmfs/sft.cern.ch/lcg/views/devAdePT/latest/x86_64-el9-gcc11-opt/` with the same package version configuration as the centos7 stack.
* By itself this change here would already change the OS to default to el9, and it would probably take several runs in jenkins to update the operating system parameter. I am not sure how to chose the OS for the PR pipeline otherwise at the moment.


JuanGonzalezCaminero (approved):

agheata (approved):
Thanks @andresailer !

PR #296, opened 2024-08-02
Keeping track of original particle ID

JuanGonzalezCaminero:
AdePT tracks were previously not storing any track ID. With this change we will propagate the ID of particles coming into AdePT to their secondaries. This ID is added to the hit information returned to Geant4.

In addition to the hits, the ID is also set as the mother for leaked particles.


agheata (approved):
I checked that the branch produces the same results as the master, so I would say it is fine to merge.

PR #299, opened 2024-08-09
Original particle ID used as ParentID for the hits

JuanGonzalezCaminero:
This changes PR [296](https://github.com/apt-sim/AdePT/pull/296) to pass the propagated primary ID as parentID for the hits generated by AdePT.

It also fixes an issue during track reconstruction on the CPU side where some vectors wouldn't be properly initialized


WitekPokorski (approved):
working fine!

PR #300, opened 2024-08-13
Surf model

agheata:
This re-enables the surface model if `VECGEOM_USE_SURF` is detected in VecGeom, and `ADEPT_USE_SURF` is ON (default OFF)

This triggers `using AdePTNavigator = SurfNavigator` which gets used for the geometry calls.
Behind `SurfNavigator` we use for now `vgbrep::Navigator` loop navigator for the `Locate` and `Safety` calls, and `vgbrep::BVHSurfNavigator` for `ComputeStep`.

Using `ADEPT_USE_SURF_SINGLE` additionally triggers the surface model with a single precision. When available, this flag will be cross-checked against a corresponding flag in VecGeom.


JuanGonzalezCaminero (approved):

agheata:
We merge this because it can process all geometries without overlaps, but we know that there are still fixes to be done to be able to validate the model.

PR #301, opened 2024-08-14
Earlier no GPU region check

JuanGonzalezCaminero:
In case there are no GPU regions defined, we return tracks to Geant4. This MR moves the check for an empty list of GPU regions to the `HandoverOneTrack` method, to avoid an issue where we would call the `PreUserTrackingAction` twice for tracks that were returned to Geant4.


WitekPokorski (approved):
looks OK

PR #303, opened 2024-08-27
 FillG4NavigationHistory fix

WitekPokorski:
This fixes the wrong logic in FillG4NavigationHistory and adds the missing MaterialCutsCouple pointer in the 'fake' step reconstructed from AdePT information and used to call the G4 sensitive detectors.

### DIFF
--- a/src/AdePTGeant4Integration.cpp
+++ b/src/AdePTGeant4Integration.cpp
@@ -364,7 +364,7 @@ void AdePTGeant4Integration::FillG4NavigationHistory(unsigned int aNavIndex, G4N
     }
   }
   // Once finished, remove the extra levels if the current state is shallower than the previous history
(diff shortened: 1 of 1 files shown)


agheata (approved):

agheata:
Looks like example1 is failing with these changes

WitekPokorski:
OK, this is super weird. The reason for the crash is that sometimes (very rarely) the volume that is reconstructed as sensitive is ECAL, which is the mother volume of the sensitive detectors (and it's not sensitive on its own). The thing is that all other sensitive volumes are reconstructed OK and all the physics results are correct. I don't know what's happening, but one possibility is that with the fix for the navigation history, we are now hitting some rare bug in VecGeom which gives the mother volume instead of the daughter volume in some rare cases. But I could be wrong.... I will keep on investigating.

WitekPokorski:
OK, I found one additional problem which I overlooked in the previous fix (visible only in specific case which we hit with example1). Now, it should be all fine.

PR #322, opened 2024-11-29
Better use of safety.

agheata:
Implementing two safety features:

1. Safety caching: the tracks hold the point where safety is computed, and the safety value in that point. When safety is needed in a different point, the method track.GetSafety(new_point) will subtract from the cached value and return the positive remainder or zero otherwise. The GetSafety method can take an accurate limit value, and when the remainder is smaller than this limit the returned safety is zero. It is the user responsibility to call track.SetSafety(pos, safety) whenever the safety is fully recomputed in a new position.

2. Safety calculation uses a new surface model feature only to compute safety accurately when close to boundaries, and only use the aligned bounding box safety when far away. The distance limit is passed as a third parameter to `AdePTNavigator::ComputeSafety`, and must be typically equal to the discrete interaction step.

### DIFF
--- a/include/AdePT/core/Track.cuh
+++ b/include/AdePT/core/Track.cuh
@@ -21,19 +21,21 @@ struct Track {
   using Precision = vecgeom::Precision;
 
   RanluxppDouble rngState;
-  double eKin;
-  double numIALeft[4];
-  double initialRange;
-  double dynamicRangeFactor;
-  double tlimitMin;
+  double eKin{0.};
+  double numIALeft[4]{0., 0., 0., 0.};
+  double initialRange{0.};
+  double dynamicRangeFactor{0.};
+  double tlimitMin{0.};
 
-  double globalTime{0};
-  double localTime{0};
-  double properTime{0};
+  double globalTime{0.};
+  double localTime{0.};
+  double properTime{0.};
 
-  vecgeom::Vector3D<Precision> pos;
-  vecgeom::Vector3D<Precision> dir;
-  vecgeom::NavigationState navState;
+  vecgeom::Vector3D<Precision> pos;   ///< track position
+  vecgeom::Vector3D<Precision> dir;   ///< track direction
+  vecgeom::Vector3D<float> safetyPos; ///< last position where the safety was computed
+  float safety{0.f};                  ///< last computed safety value
+  vecgeom::NavigationState navState;  ///< current navigation state
 
 #ifdef USE_SPLIT_KERNELS
   // Variables used to store track info needed for scoring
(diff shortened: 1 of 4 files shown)


include/AdePT/core/Track.cuh
```diff
@@ -31,9 +31,36 @@ struct Track {
   double localTime{0};
   double properTime{0};
 
-  vecgeom::Vector3D<Precision> pos;
-  vecgeom::Vector3D<Precision> dir;
-  vecgeom::NavigationState navState;
+  vecgeom::Vector3D<Precision> pos;   ///< track position
+  vecgeom::Vector3D<Precision> dir;   ///< track direction
+  vecgeom::Vector3D<float> safetyPos; ///< last position where the safety was computed
+  float safety{0};                    ///< last computed safety value
```
SeverinDiederichs, line 37:
```suggestion
  float safety{0.f};                    ///< last computed safety value
```

include/AdePT/core/Track.cuh
```diff
@@ -31,9 +31,36 @@ struct Track {
...
+  vecgeom::NavigationState navState;  ///< current navigation state
+  
+  /// @brief Get recomputed cached safety ay a given track position
+  /// @param new_pos Track position
+  /// @param accurate_limit Only return non-zero if the recomputed safety if larger than the accurate_limit
+  /// @return Recomputed safety.
+  __host__ __device__ VECGEOM_FORCE_INLINE 
+  float GetSafety(vecgeom::Vector3D<Precision> const &new_pos, float accurate_limit = 0) const
```
SeverinDiederichs, line 45:
```suggestion
  float GetSafety(vecgeom::Vector3D<Precision> const &new_pos, float accurate_limit = 0.f) const
```

include/AdePT/core/Track.cuh
```diff
@@ -31,9 +31,36 @@ struct Track {
...
+  {
+    float dsafe = safety - accurate_limit;
+    if (dsafe <= 0) return 0;
```
SeverinDiederichs, line 48:
```suggestion
    if (dsafe <= 0.f) return 0.f;
```

include/AdePT/core/Track.cuh
```diff
@@ -31,9 +31,36 @@ struct Track {
...
+    float distSq = (vecgeom::Vector3D<float>(new_pos) - safetyPos).Mag2();
+    if (dsafe * dsafe < distSq) return 0.;
```
SeverinDiederichs, line 50:
```suggestion
    if (dsafe * dsafe < distSq) return 0.f;
```

SeverinDiederichs (changes requested):
This PR breaks this mac:
[example1_ttbar.txt](https://github.com/user-attachments/files/18298388/example1_ttbar.txt)

It simply stalls and does not proceed, while on master it runs through.

Additionally, could the `DEBUG_NO_SECONDARY 1` things be removed?

PR #324, opened 2024-12-06
upgrade to latest G4hepem

SeverinDiederichs:
This PR introduces the required changes needed to use the latest version of G4HepEm. In G4HepEm the representation for the cross sections changed in this [PR](https://github.com/mnovak42/g4hepem/pull/115). Additionally, the e-+/gamma-nuclear reactions were added in G4HepEm. 

This PR
1.  uses the new function `G4HepEmGammaManager::SampleInteraction` and fixes the numIALeft.
**_Note_**: that since we don't register the gamma/lepton nuclear processes in the physics list they are currently ignored by AdePT and G4HepEm.
2. adopts the TrackingManager approach by using a derived class of the G4HepEmTrackingManager. This supersedes #316. The G4HepEmTrackingManagerSpecialized is instantiated as a member of the AdePTTrackingmanager and it takes care of all CPU propagation and cleans up the propagation loop.
3. adds a direct conversion of the Geant4 NavigationHistory to a VecGeom NavigationState. This cleans up many problems at boundaries that can become costly for edges of GPU regions, where particles are frequently moved between CPU and GPU, or very tangent particles could get stuck. This reduces the run time for 1e6 10 GeV electrons being shot at testEm3 where all layers are GPU regions (but the calorimeter is not) from
Before: Run time: 1582.26
Now: Run time: 1281.32
4. cleans up the AdePT physics. Now `FTFP_BERT_AdePT` derives from `G4EmStandardPhysics`.

Currently, the gamma/lepton nuclear processes are not registered in the PhysicsList. However, if they are registered, G4HepEm treats them correctly, while AdePT just ignores them. For electrons, they are suppressed by an infinite interaction length, for gamma (since they are handled internally in G4HepEm), we simply do a empty step. There is a implementation commented out that re-draws in case a gamma-nuclear reaction is drawn. It re-draws up to 3 times and in the (unlikely, but possible) case that we still end up with gamma-nuclear, we do a redundant step limited by gamma-nuclear but not invoking any process. However, using this changes the physics by ~ 1 % at the tail, leading to worse agreement, therefore we for now accept the performance penalty of empty steps.

**Testing:**
For the testing, different variations with testEM3 where tested:
1. making everything a GPU region via `/adept/setTrackInAllRegions true`.
2. making all layers a GPU region (but not the encompassing `Calorimeter` region.
3. using 3 different regions that include various layers, without specific order
4. using no regions at all.

These settings were tested against running with the `--noadept` option, which uses the `FTFP_BERT_HepEm` physics list. All 4 configurations were tested with this PR and with the `master` branch.

**Outcome:**
This PR shows excellent agreement between AdePT and G4 with G4HepEm: (the plot is for making everything a GPU region)
Screenshot 2025-01-01 at 16 58 52

It is slightly improved to the `master` branch:
Screenshot 2025-01-01 at 17 00 27

No major differences are observed using 2. all layers being regions, 3. using only certain regions, and 4. using no regions. 
2. All layers GPU:
Screenshot 2025-01-01 at 18 53 07
3. Various GPU regions:
Screenshot 2025-01-01 at 18 53 48
4. No GPU regions:
Screenshot 2025-01-01 at 18 54 21

In the `master`, 3. gets stuck and never finishes. It could still be off in the plot above, to be investigated with higher statistics. Since it didn't work at all in the `master` this is still a significant improvement.
The overall run time for G4 + HepEm in our example is improved using the tracking manager approach: From `Run time: 2842.99 s` (master) to `Run time: 2255.76 s` using this PR (16 threads, 1e6 10 GeV electrons being shot at testEm3)

Additionally the energy deposition of CMS was checked and found to be at good agreement. The run time for CMS is similar, maybe slightly improved using AdePT everywhere.

To Do:

- [x] Proper validation after #326 is  merged.
- [x] Validate run with CMS.
- [ ] After [this PR](https://github.com/mnovak42/g4hepem/pull/117) is merged in G4HepEm, tag that version of G4HepEm and update CI
- [x] implement changes also in split kernels
- [x] validate with split kernels (validate kernels give roughly the same result but not to machine precision anymore, however, before this PR it was not compiling anymore).

NOTE: These are breaking changes, as older versions of G4HepEm cannot be used anymore.

### DIFF
--- a/include/AdePT/kernels/gammas.cuh
+++ b/include/AdePT/kernels/gammas.cuh
@@ -85,17 +85,21 @@ __global__ void TransportGammas(adept::TrackManager<Track> *gammas, Secondaries 
     // Call G4HepEm to compute the physics step limit.
     G4HepEmGammaManager::HowFar(&g4HepEmData, &g4HepEmPars, &gammaTrack);
     G4HepEmGammaManager::SampleInteraction(&g4HepEmData, &gammaTrack, currentTrack.Uniform());
-
-    // Get result into variables.
-    double geometricalStepLengthFromPhysics = theTrack->GetGStepLength();
-    int winnerProcessIndex                  = theTrack->GetWinnerProcessIndex();
+    int winnerProcessIndex = theTrack->GetWinnerProcessIndex();
 
     // disable photo-nuclear reaction that would need to be handled by G4 itself
     if (winnerProcessIndex == 3) {
-      winnerProcessIndex = -1;
-      assert(0); // currently, the gamma-nuclear processes are not registered in the AdePTPhysicsList, so they should
-                 // never be hit.
+      // since we do an redundant step if gamma-nuclear is drawn, we redraw up to 3 times
+      int trials_left = 3;
+      do {
+        G4HepEmGammaManager::SampleInteraction(&g4HepEmData, &gammaTrack, currentTrack.Uniform());
+        trials_left--;
+        if (theTrack->GetWinnerProcessIndex() == 3 && trials_left == 0) winnerProcessIndex = -1;
+      } while (theTrack->GetWinnerProcessIndex() == 3 && trials_left > 0);
     }
+
+    // Get result into variables.
+    double geometricalStepLengthFromPhysics = theTrack->GetGStepLength();
 
     // Leave the range and MFP inside the G4HepEmTrack. If we split kernels, we
     // also need to carry them over!
--- a/include/AdePT/kernels/gammas.cuh
+++ b/include/AdePT/kernels/gammas.cuh
@@ -90,18 +90,9 @@ __global__ void TransportGammas(adept::TrackManager<Track> *gammas, Secondaries 
     // avoid photo-nuclear reaction that would need to be handled by G4 itself
     if (winnerProcessIndex == 3) {
       winnerProcessIndex = -1;
-      // NOTE: re-drawing was found to change the physics slightly, thus, it is commented out for now
-      // // since we do an redundant step if gamma-nuclear is drawn, we redraw up to 3 times
-      // int trials_left = 3;
-      // do {
-      //   // reset `number-of-interaction-left` since it is always consumed by the previous call to SampleInteraction()
-      //   theTrack->SetNumIALeft(-std::log(currentTrack.Uniform()), 0);
-      //   G4HepEmGammaManager::SampleInteraction(&g4HepEmData, &gammaTrack, currentTrack.Uniform());
-      //   trials_left--;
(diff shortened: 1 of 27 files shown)


include/AdePT/kernels/gammas.cuh
```diff
@@ -77,20 +77,36 @@ __global__ void TransportGammas(adept::TrackManager<Track> *gammas, Secondaries
...
+
+    // avoid photo-nuclear reaction that would need to be handled by G4 itself
+    if (winnerProcessIndex == 3) {
+      winnerProcessIndex = -1;
+      // NOTE: re-drawing was found to change the physics slightly, thus, it is commented out for now
+      // // since we do an redundant step if gamma-nuclear is drawn, we redraw up to 3 times
+      // int trials_left = 3;
+      // do {
+      //   // reset `number-of-interaction-left` since it is always consumed by the previous call to SampleInteraction()
+      //   theTrack->SetNumIALeft(-std::log(currentTrack.Uniform()), 0);
```
hahnjo, line 98: Shouldn't this rather be
```suggestion
      //   theTrack->SetNumIALeft(-std::log(currentTrack.Uniform()), 3);
```
?

SeverinDiederichs: Thanks for checking! I think for gammas now only the 0 index should be used, which was introduced in [this PR](https://github.com/mnovak42/g4hepem/pull/115). Easier to find in [these lines](https://github.com/mnovak42/g4hepem/blob/master/G4HepEm/G4HepEmRun/include/G4HepEmGammaManager.icc#L93-L98).

hahnjo: Ah ok. But then I think it doesn't need to be set at all because `SampleInteraction` just resets it...

Also, is this approach correct? It's resampling an interaction from a point that was reached with a mfp including the gamma-nuclear interactions, but without actually moving. I guess this is why you are seeing slight changes in physics...

SeverinDiederichs: Just for documentation:

After an offline discussion, we agreed that no simple re-drawing is possible. The `HowFar` method returns now smaller steps due to the gamma-nuclear reactions in comparison to without gamma-nuclear reactions. If we then just re-draw we get shorter steps which is exactly what was observed in form of a lower energy deposition in the tail region for testEm3 and a higher in the front. Thus, we must do an empty step without a reaction to compensate for the smaller step size returned by `HowFar`.

Additional note: the `SampleInteractions` only needs to be called after the geometry step limit is known, since we don't need to sample the interaction if we are limited by geometry anyway.

agheata (approved):

PR #325, opened 2024-12-09
Move the asynchronous mode components into AdePT

JuanGonzalezCaminero:
This PR integrates the asynchronous mode components from `AsyncExample` into AdePT as a new backend that can be activated via a compile-time option.

In order to compile either backend into the final library `AdePTTrackingManager.hh` includes either `AdePTTransport.h` or `AsyncAdePTTransport.hh`, which will pull in the gcc-compiled parts of the code. `AdePTTrackingManager.cu` includes `AdePTTransport.cuh` or `AsyncAdePTTransport.cuh` for the nvcc-compiled side. This is done via the cmake option `-DASYNC_MODE=[ON|OFF]`

There are no changes to the current synchronous backend, and as expected validation shows no difference in performance or results compared to the current `master`. For the asynchronous backend, there is no performance difference with the current standalone example, and the validation shows good agreement with Geant4.


include/AdePT/core/CommonStruct.h
```diff
@@ -65,4 +78,106 @@ struct TrackBuffer {
...
+    unsigned int maxTracks;
+    std::atomic_uint nTrack;
+    mutable std::shared_mutex mutex;
+  };
+  std::array<ToDeviceBuffer, 2> toDeviceBuffer;
+  std::atomic_short toDeviceIndex{0};
+
+  unsigned int fNumToDevice{0};   ///< number of slots in the toDevice buffer
+  unsigned int fNumFromDevice{0}; ///< number of slots in the fromDevice buffer
+  unique_ptr_cuda<TrackDataWithIDs, CudaHostDeleter<TrackDataWithIDs>> toDevice_host;                              ///< Tracks to be transported to the device
```
SeverinDiederichs, line 122: Could you make sure this is clang-formatted? This line is not formatted

include/AdePT/core/AsyncAdePTTransport.hh
```diff
@@ -25,45 +28,38 @@
...
-  std::unique_ptr<GPUstate> fGPUstate;               ///< CUDA state placeholder
-  std::vector<PerEventScoring> fScoring;             ///< User scoring objects per G4 worker
+  int fCUDAStackLimit{0};                              ///< CUDA device stack limit
+  std::vector<IntegrationLayer> fIntegrationLayerObjects;
+  // We can't use a unique_ptr because we can't have the definition of GPUState
+  // in a file compiled by gcc
+  GPUstate *fGPUstate;               ///< CUDA state placeholder
+  // We need to use pointers to the scoring objects because we can't have the complete definition 
+  // in a file compiled by gcc
+  std::vector<AdePTScoring*> fScoring;               ///< User scoring objects per G4 worker
```
SeverinDiederichs, line 62: This shows why we should refactor the header only approach.. Can we add a FIXME in the comment here, so we can easily find it again? I am not sure the pointer is deleted in the destructor

include/AdePT/core/HostScoringImpl.cuh
```diff
@@ -239,7 +239,6 @@ inline void EndOfTransport(HostScoring &hostScoring, HostScoring *hostScoring_de
 {
   // Transfer back scoring.
   CopyHitsToHost(hostScoring, hostScoring_dev, stream);
-  ResetBufferStatsGPU<<<1, 1, 0, stream>>>(hostScoring_dev);
```
SeverinDiederichs, line 242: Is this some accident in picking the wrong line in the re-basing?

This line fixed a very hard to trace bug that happened in very large events, see [this PR](https://github.com/apt-sim/AdePT/pull/329)

include/AdePT/core/PerEventScoringImpl.cuh
```diff
@@ -0,0 +1,347 @@
...
+//   /// @brief Clear hits on device to reuse for next event
+//   void ClearGPU(cudaStream_t cudaStream = 0)
+//   {
+//     COPCORE_CUDA_CHECK(cudaMemsetAsync(fScoring_dev, 0, sizeof(GlobalCounters), cudaStream));
+//     COPCORE_CUDA_CHECK(cudaStreamSynchronize(cudaStream));
+//   }
+
+//   /// @brief Print scoring info
+//   void Print() { fGlobalCounters.Print(); };
+// };
```
SeverinDiederichs, line 266: Could you add a comment on this? What is the state, why is it commented out but kept?

include/AdePT/core/PerEventScoringImpl.cuh
```diff
@@ -0,0 +1,347 @@
...
+
+//   // TODO: this needs to be done by the caller, or the net energy array moved to the scoring data
+//   // fGPUNetEnergy[threadId] = 0.;
+
+//   // TODO: This isn't critical, if needed we need to add a debug level parameter to the interface
+//   // if (fDebugLevel >= 2) {
+//   //   G4cout << "\n\tScoring for event " << eventId << G4endl;
+//   //   scoring.Print();
+//   // }
+// }
```
SeverinDiederichs, line 343: Could you please add a comment on the state of this part?

include/AdePT/integration/AdePTTrackingManager.hh
```diff
@@ -60,7 +64,12 @@ private:
   static inline int fNumThreads{0};
   std::set<G4Region const *> fGPURegions{};
   int fVerbosity{0};
+  // std::shared_ptr<AdePTTransportInterface> fAdeptTransport;
+  #ifndef ASYNC_MODE
+  std::unique_ptr<AdePTTransportInterface> fAdeptTransport;
+  #else
   std::shared_ptr<AdePTTransportInterface> fAdeptTransport;
+  #endif
```
SeverinDiederichs, line 72: I guess this part is irrelevant for performance. Wouldn't it be easier to just use `std::shared_ptr` in both cases to simplify the code and make it more readable?

include/AdePT/kernels/electrons.cuh
```diff
@@ -1,6 +1,8 @@
 // SPDX-FileCopyrightText: 2022 CERN
 // SPDX-License-Identifier: Apache-2.0
 
+#pragma message( "Compiling " __FILE__ )
+
```
SeverinDiederichs, line 5:
```suggestion
#pragma message( "Compiling synchronous kernels")
```

Is the idea to have some compiler output in case one doesn't look at the cmake? Then I would probably make the output a bit clearer

include/AdePT/magneticfield/fieldPropagatorConstBz.h
```diff
@@ -13,6 +13,8 @@
 
 #include <AdePT/magneticfield/ConstBzFieldStepper.h>
 
+// #include <cuda/std/cmath>
+
```
SeverinDiederichs, line 17:
```suggestion
```

Just some cleaning

include/AdePT/navigation/AdePTNavigator.h
```diff
@@ -9,13 +9,13 @@
 #ifndef ADEPT_NAVIGATOR_H_
 #define ADEPT_NAVIGATOR_H_
 
-#include <AdePT/navigation/LoopNavigator.h>
+// #include <AdePT/navigation/LoopNavigator.h>
```
SeverinDiederichs, line 12:
```suggestion
```

Just some cleaning

test/CMakeLists.txt
```diff
@@ -38,7 +38,7 @@ file(DOWNLOAD https://gitlab.cern.ch/VecGeom/VecGeom/raw/v1.2.0/persistency/gdml
 #----------------------------------------------------------------------------#
 # Detailed tests
 #----------------------------------------------------------------------------#
-add_subdirectory(testField)
+# add_subdirectory(testField)
```
SeverinDiederichs, line 41: Does this not compile or work anymore? Or can this be activated again? What is the status of this?

src/AdePTTrackingManager.cc
```diff
@@ -17,6 +17,14 @@
...
 #include <algorithm>
 
+#ifdef ASYNC_MODE
+std::shared_ptr<AdePTTransportInterface> InstantiateAdePT(AdePTConfiguration &conf)
+{
+  static std::shared_ptr<AsyncAdePT::AsyncAdePTTransport<AdePTGeant4Integration>> adePT{new AsyncAdePT::AsyncAdePTTransport<AdePTGeant4Integration>(conf)};
+  return adePT;
+}
+#endif
+
```
SeverinDiederichs, line 27: This looks very similar to the AdePTTransportFactory that we had before, just that the transport factory didn't require the #ifdefs. Just for the future, but I think that should be cleaned.

src/AdePTTrackingManager.cu
```diff
@@ -1,11 +1,22 @@
 // SPDX-FileCopyrightText: 2023 CERN
 // SPDX-License-Identifier: Apache-2.0
 
+// #include <cub/device/device_merge_sort.cuh>
```
SeverinDiederichs, line 4:
```suggestion
```
Just some cleaning

SeverinDiederichs (changes requested):
Thanks for this huge efforts! There are a couple of things that need to be addressed i.e., removing the testField or removing a line that fixed a critical bug. Could you have a look please?

src/AdePTTrackingManager.cc
JuanGonzalezCaminero: The #ifdefs are required since this part of the code is common to both backends. When using the sync mode `AsyncAdePTTransport` is not defined.
About the name, I'm not sure if it's correct to refer to a free function as a factory, and the purpose of it is more to make it work like a singleton. We should review whether this is the best way to instantiate the async backend

include/AdePT/core/HostScoringImpl.cuh
JuanGonzalezCaminero: Probably a rebase accident, yes, I'll add the line back

include/AdePT/core/AsyncAdePTTransport.hh
JuanGonzalezCaminero: Now using a `unique_ptr` again, the scoring objects are no longer stored as pointers either. This was a leftover from the first stages of the migration, thanks for pointing out

include/AdePT/integration/AdePTTrackingManager.hh
JuanGonzalezCaminero: I don't see an advantage to using a more complex solution where `unique_ptr` is enough. Could also be more error-prone

include/AdePT/kernels/electrons.cuh
JuanGonzalezCaminero: Was for debug, removed.

src/AdePTTrackingManager.cc
SeverinDiederichs: Sorry, my point was not phrased well:
I think the #ifdefs in the AdePTTrackingManager could be hidden inside an InstantiateAdePT function, i.e., it gets called for both and then internally does something different according to the ifdefs.

SeverinDiederichs (approved):
No more comments from my side, thanks for these efforts!

src/AdePTTrackingManager.cc
agheata: Let's say this can be done as a cleanup later, for now the priority is to have it merged

PR #326, opened 2024-12-13
fix memory corruption in scoring

SeverinDiederichs:
This PR fixes a major memory corruption problem in the scoring:

Previously, the + 1 here:
```
      ::new (toucheableHandleStorage) G4TouchableHandle{::new (toucheableHistoryStorage + 1) G4TouchableHistory};
```

Was a too small offset. Thus, the `fPostG4TouchableHistoryHandle` overlapped with the `fPreG4TouchableHistoryHandle` and overwrote its values. This is now fixed. The error broke the physics validation (note: the physics validation was done using the latest version of G4HepEm which will be made available via #324):

Before this PR:

Screenshot from 2024-12-11 16-55-26
(the absolute scale here is off due to a change in the number of particles that was not incorporated into the plot, the relative error is correct)
After this PR:
Screenshot from 2024-12-13 15-47-10

Additionally, some quality of life improvements are added:

1. The way of accessing the correct ID was reduced from:
```
hitID = fScoringMap[aStep->GetPreStepPoint()->GetTouchable()->GetHistory()->GetTopVolume()->GetInstanceID()];
```
to
```
std::size_t hitID = fScoringMap[aStep->GetPreStepPoint()->GetPhysicalVolume()->GetInstanceID()];
```
2. some unused functions were removed
3. some doubled includes were removed
4. a new option of `--accumulated_events` was added to the integrationBenchmark example. Using that option, the edep of all events is written to a single line in the output.csv.

### DIFF
--- a/examples/IntegrationBenchmark/src/EventAction.cc
+++ b/examples/IntegrationBenchmark/src/EventAction.cc
@@ -143,14 +143,24 @@ void EventAction::EndOfEventAction(const G4Event *aEvent)
       auto id = hit->GetPhysicalVolumeId() + Run::accumulators::NUM_ACCUMULATORS;
       // Reset the accumulator from the last event
       // aTestManager->setAccumulator(id, 0);
-      // Set the accumulator
-      aTestManager->setAccumulator(id, hit->GetEdep());
-      // aTestManager->addToAccumulator(id, hit->GetEdep());
+
+      if (!fRunAction->GetDoAccumulatedEvents()) {
+        // Set the accumulator
+        aTestManager->setAccumulator(id, hit->GetEdep());
+      } else {
+        // write all events to one accumulator:
+        aTestManager->addToAccumulator(id, hit->GetEdep());
+      }
     }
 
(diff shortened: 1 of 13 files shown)


agheata (approved):
Good finding @SeverinDiederichs

examples/IntegrationBenchmark/src/EventAction.cc
```diff
@@ -143,14 +143,24 @@ void EventAction::EndOfEventAction(const G4Event *aEvent)
...
+      }
     }
 
-    // Write data to output file. Validation data can take a lot of memory, and we don't need to aggregate
-    // the results of multiple events at runtime, so for better performance it's easier to write the output here
-    aTestManager->exportCSV(false);
+    if (!fRunAction->GetDoAccumulatedEvents()) {
+      // Write data to output file. Validation data can take a lot of memory, and we don't need to aggregate
+      // the results of multiple events at runtime, so for better performance it's easier to write the output here
+      // aTestManager->exportCSV(false);
```
SeverinDiederichs, line 159:
```suggestion
      aTestManager->exportCSV(false);
```

PR #329, opened 2025-01-06
reset buffer stats in HostScoring

SeverinDiederichs:
This PR fixes a critical bug in the HostScoring implementation:

The 3 variables required for the next hit slot
```
  hostScoring_dev->fBufferStart        = 0;
  *(hostScoring_dev->fNextFreeHit_dev) = 0;
  *(hostScoring_dev->fUsedSlots_dev)   = 0;
```
were not reset between events. Although they were ever increased via `  unsigned int next = hostScoring_dev->fNextFreeHit_dev->fetch_add(1);` (and the `fBufferStart` was set to be the `fNextHitFree`), it worked due to the circular nature of the buffer `return next % hostScoring_dev->fBufferCapacity;`. However, as soon as the maximum int was exceeded, the simulation would simply stall and not do anything else anymore. This PR fixes that by resetting the three variables between events using a new function `ResetBufferStatsGPU`.

Note that there are a huge amount of changes due to the missing clang formatting in the file. The only other change was to reduce some input parameters for e.g., `CopyHitsToHost` and just use the hostscoring itself to get the stats.

### DIFF
--- a/include/AdePT/core/HostScoringImpl.cuh
+++ b/include/AdePT/core/HostScoringImpl.cuh
@@ -37,13 +37,7 @@ __device__ __forceinline__ unsigned int GetNextFreeHitIndex(HostScoring *hostSco
   // Atomic addition, each GPU thread accessing concurrently gets a different slot
   unsigned int next = hostScoring_dev->fNextFreeHit_dev->fetch_add(1);
   assert(next >= hostScoring_dev->fBufferStart);
-  if (next < hostScoring_dev->fBufferStart) {
-    // if (next %1000 == 0) {
-    printf("Oh no, YOU SHOULD NEVER BE HERE!: next %u fBufferStart %u \n", next, hostScoring_dev->fBufferStart);
-  }
   if ((next - hostScoring_dev->fBufferStart) >= hostScoring_dev->fBufferCapacity) {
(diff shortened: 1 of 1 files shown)


include/AdePT/core/HostScoringImpl.cuh
```diff
@@ -98,145 +106,147 @@ __device__ __forceinline__ void refresh_stats(HostScoring *hostScoring_dev)
...
 
-
-
+template <typename IntegrationLayer>
+inline void EndOfTransport(HostScoring &hostScoring, HostScoring *hostScoring_dev, cudaStream_t &stream,
+                           IntegrationLayer &integration)
+{
+  // Transfer back scoring.
+  CopyHitsToHost(hostScoring, hostScoring_dev, stream);
+  ResetBufferStatsGPU<<<1, 1, 0, stream>>>(hostScoring_dev);
```
agheata, line 242: Why launch a new kernel to reset this and not call a Reset function inside say `UpdateBufferStartGPU` where we already pass the hostScoring_dev pointer?

agheata (commented):
Good for me, except resetting the counters in a new kernel launch, see comment inline. I wonder why the CI doesn't catch missing clang-format, we need to investigate.

agheata (approved):

PR #331, opened 2025-01-10
fix minor things for strict compiler settings

SeverinDiederichs:
In the hunt for some ugly non-reproducibility bug, I used strict compiler settings with `add_compile_options(-Wall -Wextra -Werror -Wuninitialized)`

This PR cleans unused variables and incorrect types that don't have any effect but prevent the usage of strict compiler settings.


agheata (approved):
Good! Maybe we should enable those options by default, to catch problems earlier? I'm OK with adding them in this PR.

agheata:
And we should add a clang-format test in the CI

PR #332, opened 2025-01-13
Fix validation, add CI

SeverinDiederichs:
This PR fixes an issue with the `--accumulated_events` option, which was not correct when run in multi-threading. In fact, previously shown physics validation plots used 16x number of events too little. Thus, the error is smaller than it appeared.

Then, the PR fixes an error in initializing the track on GPU. An index for the navigation state was missing. It was also added for the global times to improve memory access pattern. The error was responsible for a higher difference in the validation for using only certain regions in testEM3.

Finally, this PR adds a validation CI test. It shoots 1e5 10 GeV electrons on testEm3 and compares against a high-statistics run (1e7 10 GeV electrons) and compares the energy deposition. Relative errors errors must be below 1% to pass the test. 
Note that this could be more restricted (i.e., less than 0.5% or even lower) but that would require a higher number of electrons and could take longer.

Additionally, a reproducibility CI test was added. However, it is currently commented out and not used, since the code is currently not reproducible and the test can only be activated once the error is fixed.


agheata:
You have to add a file `examples/IntegrationBenchmark/ci_tests/benchmark_files/g4_em3_10e7_e-.csv.license`  containing:
```
SPDX-FileCopyrightText: 2025 CERN
SPDX-License-Identifier: CC-BY-ND-4.0
```
to fix the REUSE Compliance Check

agheata (approved):

PR #289, opened 2024-05-01
Minimal implementation of in-memory Geant4->VecGeom conversion with G4VG

drbenmorgan:
This implements the minimal code/link to use [G4VG](https://github.com/celeritas-project/g4vg) for converting a Geant4 `G4VPhysicalVolume` to `vecgeom::VPlacedVolume` in-memory without an intermediary GDML file. Because of the link to CUDA-enabled VecGeom, the link structure originally trialled in #279 is also implemented, and now works properly thanks to the work done removing the old circular dependency described in that PR. Note that it's certain we'll need the `--no-as-needed` trick for linkers that use `as-needed` by default, but first things first. The major changes are thus:

1. Import the `CudaRdcUtils` and `FindVecGeom` modules from Celeritas to support coherent linking/device linking of VecGeom into Cuda clients.
2. Use and export the functions to compile and link the AdePT libraries/tests
3. Move import of G4VG to top level and bump to latest upstream commit which better supports use and install
4. Implement `AdePTGeant4Integration::CreateVecGeomWorld(const G4VPhysicalVolume *physvol)` to take the input Geant4 volume and convert to VecGeom using G4VG's interface.

Because of the linking issues seen before I haven't gone further yet in migrating the code to use `CreateVecGeomWorld`. I'd like to get a round of CI in first just to be sure there's no other issue. We then should confirm/decide how to solve the `as-needed` issue. Whilst it's a bit of a blunt instrument, we probably just link `AdePT_G4_integration` with `-Wl,--no-as-needed`, and perhaps enforce downstream clients to use that as well.

Supersedes #279 and #285.

PR #340, opened 2025-01-31
Integrate G4VG for use of in-memory conversion of Geant4->VecGeom geometries

drbenmorgan:
This supersedes #289 to integrate the now standalone celeritas-project/G4VG library into AdePT for in-memory geometry conversions without an intermediate GDML file. The as-submitted implementation is:

- G4VG is pulled in via CMake's `FetchContent` mechanism to build it as part of AdePT.
- It is linked in to the AdePT integration library
- `AdePTGeant4Integration::CreateVecGeomWorld(G4VPhysicalVolume const *physvol)` is implemented to use G4VG to do the in-memory conversion.
- `AdePTTrackingManager::InitializeAdePT()` is updated to use that function _if_ `AdePTConfiguration::fVecGeomGDML` is empty.
- An extra test of example1 is provided that is identical to the existing test, except for not setting `AdePTConfiguration::fVecGeomGDML`.
  - I've confirmed locally that this results in the in memory conversion being called, but there's no reproducibility check against the GDML-using version.

~~I'm marking as draft as there's a final merge to do upstream in G4VG to get a proper commit to use, however, I wanted to start things off to get some initial CI checks done and for initial comments.~~ That's done, so fully ready for review!

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -152,6 +152,7 @@ if(Geant4_FOUND)
       g4vg
       GIT_REPOSITORY https://github.com/sethrj/g4vg 
       GIT_TAG 7e9f3224c57b97d665d885f3a6bccb8fee1eaf54)
+    set(BUILD_SHARED_LIBS ON)
     FetchContent_MakeAvailable(g4vg)
   endif()
 endif()
@@ -216,6 +217,7 @@ target_link_libraries(AdePT_G4_integration
     G4HepEm::g4HepEmData
     G4HepEm::g4HepEmInit
     G4HepEm::g4HepEmRun
+    G4VG::g4vg
     CUDA::cudart
 )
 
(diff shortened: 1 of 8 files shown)


drbenmorgan:
Latest commit uses the current production tag of G4VG v1.0.1, so fully ready assuming CI passes!

agheata:
> Latest commit uses the current production tag of G4VG v1.0.1, so fully ready assuming CI passes!

@JuanGonzalezCaminero is currently testing it by activating the conversion in example1 and debugging some crashes related to using the placed volume map filled during the conversion. We will merge as soon as this is clarified/fixed

JuanGonzalezCaminero:
PR looks good, thanks Ben!

While testing the G4VG generated PlacedVolume ID to G4 Physical volume map on AdePT we have noticed a small number of wrongly mapped IDs as compared to a map generated by a geometry visitor. It would be great if @sethrj could comment on this

However we are merging this PR now as the integration is unrelated to this issue, and we can work around it by generating the volume map on AdePT side for the time being.

PR #341, opened 2025-01-31
Unify common part of RecordHit

JuanGonzalezCaminero:
The sync and async modes differ only on how the GPUHit slot is retrieved, this PR unifies the rest of the code in the `FillHit` call

Also, improve the RecordHit interface by passing references instead of pointers.


SeverinDiederichs (commented):
Could you please clang-format the files? Then it looks good to me

SeverinDiederichs (approved):
Thanks for the unification!

PR #349, opened 2025-02-18
Using G4VG's PlacedVolume map

JuanGonzalezCaminero:
Enables the new G4VG option to avoid using VecGeom's reflection factory (https://github.com/celeritas-project/g4vg/pull/22), which solves the wrong mapping of the daughters of reflected volumes in G4VG's PlacedVolume ID to G4 physical volume map.

Note that since we don't build the map on AdePT side anymore this change is incompatible with any previous G4VG commits.


SeverinDiederichs (approved):
Thanks for taking care of this fix!

PR #351, opened 2025-02-26
Handle converted G4VPVParameterised/G4PVReplica volumes from G4VG in geometry checks

drbenmorgan:
The latest updates to G4VG allow Geant4 parametrised/replica volumes to be converted to a VecGeom representation by direct placements. The need for this comes from the Delta Assessment "B5" example that uses these structures and for which there is no direct VecGeom equivalent except by making a direct placement for each copy. That's supported as of [G4VG 1.0.3](https://github.com/celeritas-project/g4vg/releases/tag/v1.0.3), but using this in AdePT and trying out B5 still leads to errors from `AdePTG4Integration::visitGeometry`.

That's sort of expected as the translation between Geant4 and VecGeom means that there will be N VecGeom placements for each replicated volume, N corresponding to the total number of replicas. Consequently, the checks in `visitGeometry` fail as:

1. The number of daughters in a given LV may be larger
2. Volume transformations may not match up as the original "parent" volume in Geant4 is compared against some copy of the replicated volume in VecGeom
    
The changes here address this, at least as far as the B5 geometry is concerned:

1. The number of LV daughters on the Geant4 side are counted as the sum of daughter multiplicities
   - For ordinary volumes this should be identical
   - It should also work for a mix of daughters (some standard, some parametrised), but this is untested. 
2. The transformation problem is a bit hacky, though pretty much copies what is done on the G4VG side:
   - If the Geant4 PV provides a `G4VPVParametrisation`, use that to transform it using the copy number of the compared VecGeom PV before extracting the transform from the Geant4 PV
   - If the Geant4 PV is a `G4PVReplica` use a `G4ReplicaNavigation` to do the same as above.
   - The hacky bit here is that use of this method requires a `const_cast` on the Geant4 PV as the transform step is non-const...
   - This does modify the underlying PV, but I think that's o.k (?) as it's only relevant for the navigation stage.

Testing with a local build against B5 shows this to now run to completion without error, but full validation has not been done yet (that's probably best addressed separately).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -157,7 +157,8 @@ if(Geant4_FOUND)
     FetchContent_Declare(
       g4vg
       GIT_REPOSITORY https://github.com/celeritas-project/g4vg
-      GIT_TAG 638c273b744c7ac0fa865ddb15c161b853a4029b)
+      GIT_TAG f4308d392a502b7bd7468938f8d9a63198d3d866 # v1.0.3
+    )
     # G4VG builds static by default, so change this to shared to match current
     # way AdePT is built.
     # could also configure for PIC mode static.
@@ -165,7 +166,7 @@ if(Geant4_FOUND)
     FetchContent_MakeAvailable(g4vg)
     message(STATUS "Using FetchContent to build G4VG as part of AdePT")
   else()
-    find_package(G4VG 1.0.1 REQUIRED)
+    find_package(G4VG 1.0.3 REQUIRED)
     message(STATUS "Found G4VG: ${G4VG_DIR}")
   endif()
 endif()
--- a/src/AdePTGeant4Integration.cpp
+++ b/src/AdePTGeant4Integration.cpp
@@ -17,6 +17,8 @@
 #include <G4FieldManager.hh>
 #include <G4RegionStore.hh>
 #include <G4TouchableHandle.hh>
+#include <G4PVReplica.hh>
+#include <G4ReplicaNavigation.hh>
 
 #include <G4HepEmState.hh>
 #include <G4HepEmData.hh>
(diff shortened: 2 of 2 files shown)


drbenmorgan:
Argh, reproducibility in CMS is failing, though not at high percentages - will need to do some more checks..

SeverinDiederichs:
> Argh, reproducibility in CMS is failing, though not at high percentages - will need to do some more checks..

Hi Ben, thanks for the PR!
I looked at the errors in the CI and this looks to me like a really significant reproducibility error and not just rounding.
Let me know if you need any help with debugging!

agheata (commented):
Thanks @drbenmorgan ! It is pretty weird that the CMS reproducibility test fails, because the actual changes done by Ben should not be hit, since cms_2018 does not have replicas/parameterisations. Does this mean that  actually  the change from G4VG 1.0.1 to 1.0.3 broke it?

drbenmorgan:
Taking a look at this locally, I'm not able to reproduce the failure with:

- Geant4 11.2.2
- VecGeom master (as of when I submitted) and I don't see anything obviously significant since)
- All CUDA builds with GCC13+CUDA 12.5 (compute capability 8.6)

The CI is using GCC11/CUDA 11.8 plus geant4 11.0.0 and VecGeom 2.0.0-dev3, compute capability 75. I'm trying to run that environment on lxplus-gpu, the build is fine, just the GPU is clogged with other jobs so I can't run the CMS test without running out of memory.

I'll try locally with those geant4/vecgeom versions in case this is causing the problem.

drbenmorgan:
@phsft-bot build

drbenmorgan:
Huh, now the CI is passing! I _guess_ as long as no one else is seeing a problem, then maybe a glitch/clash on the Jenkins side perhaps?

SeverinDiederichs:
> Huh, now the CI is passing! I _guess_ as long as no one else is seeing a problem, then maybe a glitch/clash on the Jenkins side perhaps?

I hope so! I also cannot reproduce the error locally, for me it passes as well.

SeverinDiederichs (approved):
Great! Thank you very much Ben! I think we can go ahead then

PR #352, opened 2025-02-27
Storing vertex information

JuanGonzalezCaminero:
As part of the modifications required to run AdePT in ATLAS Athena, with this PR we keep track of vertex information needed by some user codes in Athena, mirroring this change in the G4HepEM tracking manager: https://github.com/mnovak42/g4hepem/pull/108/commits/b3dc506679424e0814c67b75612863bd25e44aba

As in the previous Athena-related PR https://github.com/apt-sim/AdePT/pull/347, we don't pass this information for scoring yet, but this can be done if eventually needed.

Performance test for these additions:

`CMS2018, 32 TTbar, 8 Threads, Transport on ECAL+HCAL`

|        | Sync    | Async   |
|--------|---------|---------|
| Master | 305.363 | 284.428  |
| New    | 308.255 | 283.359 |

`TestEm3, 128 events, 100x10GeV e-`

|        | Sync    | Async   |
|--------|---------|---------|
| Master | 26.2131 | 17.5006  |
| New    | 26.0619 | 17.2154 |

Sync: 4M Track slots, 10M Hit slots
Async: 4M Track slots, 25M Hit slots

### DIFF
--- a/include/AdePT/core/AdePTTransport.h
+++ b/include/AdePT/core/AdePTTransport.h
@@ -41,9 +41,11 @@ public:
   int GetNfromDevice() const { return fBuffer.fromDevice.size(); }
 
   /// @brief Adds a track to the buffer
-  void AddTrack(int pdg, int parentId, double energy, double x, double y, double z, double dirx, double diry,
-                double dirz, double globalTime, double localTime, double properTime, int threadId, unsigned int eventId,
-                unsigned int trackIndex, vecgeom::NavigationState &&state, vecgeom::NavigationState &&originState);
+  void AddTrack(int pdg, int parentId, double energy, double vertexEnergy, double x, double y, double z, double dirx,
+                double diry, double dirz, double vertexX, double vertexY, double vertexZ, double vertexDirx,
+                double vertexDiry, double vertexDirz, double globalTime, double localTime, double properTime,
+                int threadId, unsigned int eventId, unsigned int trackIndex, vecgeom::NavigationState &&state,
+                vecgeom::NavigationState &&originState);
 
   void SetTrackCapacity(size_t capacity) { fCapacity = capacity; }
   /// @brief Get the track capacity on GPU
--- a/include/AdePT/core/CommonStruct.h
+++ b/include/AdePT/core/CommonStruct.h
@@ -104,19 +104,27 @@ struct TrackDataWithIDs : public adeptint::TrackData {
   unsigned int trackId{0};
   short threadId{-1};
 
-  TrackDataWithIDs(int pdg_id, int parentId, double ene, double x, double y, double z, double dirx, double diry,
-                   double dirz, double gTime, double lTime, double pTime, vecgeom::NavigationState &&state,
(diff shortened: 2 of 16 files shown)


include/AdePT/core/AdePTTransport.h
```diff
@@ -41,9 +41,11 @@ class AdePTTransport : public AdePTTransportInterface {
   int GetNfromDevice() const { return fBuffer.fromDevice.size(); }
 
   /// @brief Adds a track to the buffer
-  void AddTrack(int pdg, int parentId, double energy, double x, double y, double z, double dirx, double diry,
-                double dirz, double globalTime, double localTime, double properTime, int threadId, unsigned int eventId,
-                unsigned int trackIndex, vecgeom::NavigationState &&state, vecgeom::NavigationState &&originState);
+  void AddTrack(int pdg, int parentId, double energy, double vertexEnergy, double x, double y, double z, double dirx,
```
agheata, line 44: In future we may want to simplify this interface and at least pass Vector3D rather than components

agheata:
The CI fails

JuanGonzalezCaminero:
> The CI fails

Passing now by reducing the number of track slots on GPU, which was too large for the sync mode. It seems that sync/async mode are allocating different amount of memory for the same buffer sizes

include/AdePT/core/CommonStruct.h
```diff
@@ -104,19 +104,27 @@ struct TrackDataWithIDs : public adeptint::TrackData {
...
-                   short threadId = -1)
+  TrackDataWithIDs(int pdg_id, int parentId, double ene, double vertexEne, double x, double y, double z, double dirx,
+                   double diry, double dirz, double vertexX, double vertexY, double vertexZ, double vertexDirx,
+                   double vertexDiry, double vertexDirz, double gTime, double lTime, double pTime,
+                   vecgeom::NavigationState &&state, vecgeom::NavigationState &&originState, unsigned int eventId = 0,
+                   unsigned int trackId = 0, short threadId = -1)
       : TrackData{pdg_id,
                   parentId,
                   ene,
+                  vertexEne,
```
SeverinDiederichs, line 115: I noticed that various names are used: `vertexEkin`, `vertexEne`, and `vertexEnergy`. Could you please unify them?

SeverinDiederichs (commented):
Apart from the minor comment, looks good to me. Thank you for the addition!

include/AdePT/core/CommonStruct.h
JuanGonzalezCaminero: In this case I'm trying to be consistent with the naming of things within the class (Which is inconsistent between classes, for example it's called `ene` in `TrackData` and `Ekin` in `Track`). I think that we can refactor all fields in a separate commit

SeverinDiederichs: Ok, would be good to do that soon and not forget about it. I agree, that can be done in a separate PR.

Then we can go ahead with this one. Thanks!

PR #357, opened 2025-03-09
Adding external B field via covfie

SeverinDiederichs:
This PR adds an external B field via the covfie library. This PR is based on #356 which must be merged first.

The B field implementation still requires major cleaning:

- Currently if AdePT is compiled with covfie support, no constant B field is available anymore, just the 3D fieldmap from file. Both options should be available, just the RK propagator should be compiled and not the Helical Stepper.
- The 3D B field view is passed as a global, just like the constant B field. Both should be cleaned.
- The implementation still has many things that need to be cleaned and improved, e.g., the safety calculated in the RK stepper should be re-used and use the setSafety functionality.

Given that the review relies on this branch, we should merge it and do the necessary developments for the review and do the cleaning and other improvements later.


agheata (approved):
Thanks @SeverinDiederichs, LGTM !

PR #358, opened 2025-03-12
Fix leaks in asynchronous mode

JuanGonzalezCaminero:
This fixes the extraction of leaked tracks in Async mode. The previous extraction of leaks didn't work in most cases for two main reasons:
- The size of the buffer used for extracting the leaked tracks is limited, so if there are too many, the extraction needs to be done in multiple steps. However, there was no mechanism to do this, tracks would get postponed until the next extraction step, but the corresponding events were notified that all leaks had been extracted. This means that whenever the number of leaks exceeded the buffer size the overflow was lost.
- There wasn't a way to guarantee that all leaks corresponding to the event requesting a flush had been extracted. The tracks are not extracted in order of insertion, and the transport kernels were allowed to insert new tracks into the leak queues in-between transfers of tracks from device to host. This means that we couldn't know when all leaks corresponding to an event had actually been extracted.

With these changes:
- The leak queues are frozen whenever an event requests a flush of their leaks. We keep two leak queues per particle type, we swap them at the start of the leak extraction, and don't swap them back until the extraction ends and another event requests a flush. This ensures that when we finish transferring the frozen queue, all leaks associated to that event have been extracted.
- We use a new state machine that doesn't return to the idle state until all leaks have been transferred. The functionality is similar to that of the scoring thread, but instead of running separately from the transport loop, the work is done by the AdePT thread.
- Returning all leaks uncovers an issue with the MPMC queues used to store leaked tracks during transport. These queues have a fixed size and need to be emptied before they can be used again, and it is left to the user to check whether the queue is full. The queues were inadvertedly overflowing in specific use cases (Such as the testEm3 geometry when running large events), this has been solved by stopping transport in situations where the queues are close to overflowing but a leak extraction is already in progress. Since this is not an issue in more realistic use cases, this mechanism ensures that the execution will be error-free in all situations without affecting runtime in realistic runs.
- In specific use cases like running large events in testEm3 the results were sometimes non-reproducible. This was likely due to some interference between the leak extraction, hit extraction, and particle injection, which ran on the same stream. The issue is solved by running these three operations on different streams, but should be revisited in the future.

The new `ExtractStates` are the following:
- **Idle**: No extraction is in progress
-  **ExtractionRequested**: An event requested a flush, waiting for transport to finish
    - We move from `Idle` to `ExtractionRequested` when an event has requested a flush, and their `EventState` is `HitsFlushed`. The number of particles in flight for that event is guaranteed to be 0 at this point
    - We wait for the transport kernels to finish running as they are writing to the same leak queues that we need to transfer
-  **TracksNeedTransfer**: An event requested a flush, leak buffer on device has tracks to transfer
    - Transport has finished, the leak queues are frozen and we can start transferring tracks
-  **PreparingTracks**: Tracks are being copied to the staging buffer
    - Tracks are copied from the leak queues to the transfer buffer, the numbers of transferred and remaining tracks are copied to the host.
-  **TracksReadyToCopy**: Staging buffer is ready to be copied to host
-  **CopyingTracks**: Tracks are being copied to host
-  **TracksOnHost**: Some or all the tracks have been transferred from device to host and are waiting in the copy buffer
-  **SavingTracks**: Tracks are being copied to per-event queues
    - Tracks are copied from the transfer buffer, which can contain tracks for all events, into per-event queues
-  **TracksSaved**: Tracks have been moved from the copy buffer to their respective per-event queues
    - All tracks are stored in the per-event queues, so the transfer buffer can be re-used. If there are any remaining leaks on GPU the state goes back to `TracksNeedTransfer`, otherwise the state moves to `Idle`

extract_state_machine

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,9 +42,6 @@ set(CMAKE_CUDA_STANDARD_REQUIRED ${CMAKE_CXX_STANDARD_REQUIRED})
 set(CMAKE_CUDA_EXTENSIONS OFF)
 set(CMAKE_INCLUDE_DIRECTORIES_PROJECT_BEFORE ON)
 
-# set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xcompiler=-fno-sanitize=address") 
-# set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address") 
-
 #----------------------------------------------------------------------------#
 # User options
 #----------------------------------------------------------------------------#
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,6 +42,9 @@ set(CMAKE_CUDA_STANDARD_REQUIRED ${CMAKE_CXX_STANDARD_REQUIRED})
 set(CMAKE_CUDA_EXTENSIONS OFF)
 set(CMAKE_INCLUDE_DIRECTORIES_PROJECT_BEFORE ON)
 
+# set(CMAKE_CUDA_FLAGS "${CMAKE_CUDA_FLAGS} -Xcompiler=-fno-sanitize=address") 
+# set(CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} -fsanitize=address") 
+
 #----------------------------------------------------------------------------#
 # User options
 #----------------------------------------------------------------------------#
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -327,6 +327,10 @@ __global__ void ZeroEventCounters(Stats *stats)
     stats->perEventInFlight[i] = 0;
     stats->perEventLeaked[i]   = 0;
   }
+  for (unsigned int i = threadIdx.x; i < ParticleType::NumParticleTypes; i += blockDim.x) {
+    stats->nLeakedCurrent[i] = 0;
+    stats->nLeakedNext[i]    = 0;
+  }
 }
 
 /**
@@ -369,16 +373,30 @@ __global__ void CountCurrentPopulation(AllParticleQueues all, Stats *stats, Trac
 __global__ void CountLeakedTracks(AllParticleQueues all, Stats *stats, TracksAndSlots tracksAndSlots)
 {
   constexpr auto nQueue = 2 * ParticleType::NumParticleTypes;
+  // One block processes each queue
   for (unsigned int queueIndex = blockIdx.x; queueIndex < nQueue; queueIndex += gridDim.x) {
-    const auto particleType   = queueIndex / 2;
+    const auto particleType =
+        queueIndex < ParticleType::NumParticleTypes ? queueIndex : queueIndex - ParticleType::NumParticleTypes;
     Track const *const tracks = tracksAndSlots.tracks[particleType];
     auto const queue = queueIndex < ParticleType::NumParticleTypes ? all.queues[particleType].leakedTracksCurrent
                                                                    : all.queues[particleType].leakedTracksNext;
-    const auto end   = queue->size();
-    for (unsigned int i = threadIdx.x; i < end; i += blockDim.x) {
+    const auto size  = queue->size();
+    for (unsigned int i = threadIdx.x; i < size; i += blockDim.x) {
       const auto slot     = (*queue)[i];
       const auto threadId = tracks[slot].threadId;
       atomicAdd(stats->perEventLeaked + threadId, 1u);
+    }
+
+    // Update the global usage
+    if (threadIdx.x == 0) {
+      // printf("%d, %d\n", particleType, queueIndex);
+      queueIndex < ParticleType::NumParticleTypes ? stats->nLeakedCurrent[particleType] = size
+                                                  : stats->nLeakedNext[particleType]    = size;
+      // if (queueIndex < ParticleType::NumParticleTypes) {
+      //   printf("NUM LEAKED CURR: %d, real: %lu\n", *(stats->nLeakedCurrent + particleType), size);
+      // } else {
+      //   printf("NUM LEAKED NEXT: %d, real: %lu\n", *(stats->nLeakedNext + particleType), size);
+      // }
     }
   }
 }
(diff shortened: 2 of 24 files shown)


CMakeLists.txt
```diff
@@ -42,6 +42,12 @@ set(CMAKE_CUDA_STANDARD_REQUIRED ${CMAKE_CXX_STANDARD_REQUIRED})
 set(CMAKE_CUDA_EXTENSIONS OFF)
 set(CMAKE_INCLUDE_DIRECTORIES_PROJECT_BEFORE ON)
 
+
+# DEBUG
+# set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fsanitize=address -fno-omit-frame-pointer -g")
```
andresailer, line 47: Can I interest you in our builds with no-omit-framepointer enabled in the compiler (gcc14fp)?

test/regression/CMakeLists.txt
```diff
@@ -23,8 +23,8 @@ endmacro()
 #----------------------------------------------------------------------------#
 # Common Data
 #----------------------------------------------------------------------------#
-file(COPY "${PROJECT_SOURCE_DIR}/examples/data/testEm3.gdml" DESTINATION "${PROJECT_BINARY_DIR}")
-set(TESTING_GDML "${PROJECT_BINARY_DIR}/testEm3.gdml")
+# file(COPY "${PROJECT_SOURCE_DIR}/examples/data/testEm3.gdml" DESTINATION "${PROJECT_BINARY_DIR}")
+# file(COPY "${PROJECT_SOURCE_DIR}/examples/data/testEm3_regions.gdml" DESTINATION "${PROJECT_BINARY_DIR}")
```
SeverinDiederichs, line 27: This seems to be a leftover and should be reverted. (the commented out code was fixed and the now removed code is actually correct)

test/regression/IntegrationTest/src/RunAction.cc
```diff
@@ -76,7 +76,7 @@ void RunAction::EndOfRunAction(const G4Run *)
 
   // Print timer just for the master thread since this is called when all workers are done
   if (tid < 0) {
-    std::cout << "Run time: " << time << "\n";
+    G4cout << "Run time: " << time << "\n";
```
SeverinDiederichs, line 79: I think this was changed because with G4cout it didn't print the time anymore. Can you confirm it actually prints the time although it is not using std::cout?

SeverinDiederichs (commented):
Great, thank you for this big effort! See minor comments below

test/regression/CMakeLists.txt
JuanGonzalezCaminero: We are actually not using the copied files as we give the absolute path in the scripts. I can uncomment these lines and change the path in the scripts too, both things work for me

test/regression/IntegrationTest/src/RunAction.cc
JuanGonzalezCaminero: Ah I didn't remember which one was correct, as both were printing the time. I changed it because I thought it hadn't been updated, will revert

SeverinDiederichs: If both work, then no preference from my side

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -960,70 +993,222 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
+          }
+        }
+
+        // If not extracting tracks from a previous event, freeze the current leak queues and swap the next ones in
+        if (gpuState.extractState.load(std::memory_order_acquire) == ExtractState::Idle && leakExtractionNeeded) {
+
+          // if (gpuState.extractState.load(std::memory_order_acquire) == ExtractState::Idle &&
+          //   std::any_of(eventStates.begin(), eventStates.end(), [](const auto &eventState) {
+          //     return eventState.load(std::memory_order_acquire) == EventState::HitsFlushed;
+          //   })) {
```
SeverinDiederichs, line 1058:
```suggestion
```

Looks like some leftover?

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -960,70 +993,222 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
+          // Update the state so that the staging buffer will not be modified again until tracks have been copied
+          // gpuState.extractState = ExtractState::PreparingTracks;
+          AdvanceExtractState(ExtractState::TracksNeedTransfer, ExtractState::PreparingTracks, gpuState.extractState);
+
+          // // Populate the staging buffer and copy to host
+          constexpr unsigned int block_size = 128;
+          const unsigned int grid_size      = (trackBuffer.fNumFromDevice + block_size - 1) / block_size;
+          FillFromDeviceBuffer<<<grid_size, block_size, 0, extractStream>>>(
+              allLeaked, trackBuffer.fromDevice_dev.get(), trackBuffer.fNumFromDevice,
+              // printtotal, allLeaked, trackBuffer.fromDevice_dev.get(), trackBuffer.fNumFromDevice,
```
SeverinDiederichs, line 1118:
```suggestion
```

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -804,61 +852,45 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
-          }
-          // If eventState leaves Inject unexpectedly, reset iteration tracking:
-          else if (injectIteration[i] != -1 && state != EventState::Inject) {
-            injectIteration[i] = -1;
+            eventState = EventState::InjectionCompleted;
           }
         }
 
-        if (auto &toDevice = trackBuffer.getActiveBuffer(); toDevice.nTrack.load(std::memory_order_acquire) > 0) {
+        if (auto &toDevice = trackBuffer.getActiveBuffer(); toDevice.nTrack > 0) {
```
SeverinDiederichs, line 863: Why was the atomic read removed here?

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -804,61 +852,45 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
           gpuState.injectState = InjectState::CreatingSlots;
 
           trackBuffer.swapToDeviceBuffers();
           std::scoped_lock lock{toDevice.mutex};
-          const auto nInject = std::min(toDevice.nTrack.load(std::memory_order_acquire), toDevice.maxTracks);
+          const auto nInject = std::min(toDevice.nTrack.load(), toDevice.maxTracks);
```
SeverinDiederichs, line 868: Here I understand that we use the lock now

SeverinDiederichs (commented):
See minor comments

SeverinDiederichs (approved):
Great, thank you for this big effort!

PR #359, opened 2025-03-12
remove error-prone name check and G4VG linking test

SeverinDiederichs:
A minor cleaning PR that removes the error-prone name check (which is triggered in FullSimLight) and the G4VG linking test, as we now build it directly.


SeverinDiederichs:
CI fails due to the upgrade to the required upgrade to GCC 13 in the CI. Should work after the nightlies build the correct version.

agheata (approved):

PR #360, opened 2025-03-16
Enable event finishing

SeverinDiederichs:
This PR introduces a new feature for the async mode: `/adept/FinishLastNParticlesOnCPU`.

It is not complete yet, as they are currently just killed and not pushed to the CPU, as this will require #358 to work and be merged.

Still, killing the last N particles can currently be used to assess some performance bottlenecks with the magnetic field.

Also, a timer is added to the printout of the async printouts.

This allows for the plotting the numbers of flight vs time and not just only vs iterations, as this severely skews it towards the fast iterations at the tail:
Screenshot 2025-03-16 at 10 45 45

### DIFF
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -827,6 +827,7 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
 
       AllowFinishOffEventArray allowFinishOffEvent;
       for (int i = 0; i < kMaxThreads; ++i) {
+        // if waiting for transport to finish, the last N particles may be finished on CPU
         if (eventStates[i].load(std::memory_order_acquire) == EventState::WaitingForTransportToFinish) {
           allowFinishOffEvent.flags[i] = lastNParticlesOnCPU;
         } else {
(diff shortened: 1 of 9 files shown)


include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -820,12 +825,23 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
+        }
+      }
+
       // *** ELECTRONS ***
       {
         const auto [threads, blocks] = computeThreadsAndBlocks(particlesInFlight[ParticleType::Electron]);
         TransportElectrons<PerEventScoring><<<blocks, threads, 0, electrons.stream>>>(
             electrons.tracks, electrons.queues.currentlyActive, secondaries, electrons.queues.nextActive,
-            electrons.queues.leakedTracksCurrent, gpuState.fScoring_dev, returnAllSteps, returnLastStep);
+            electrons.queues.leakedTracksCurrent, gpuState.fScoring_dev, gpuState.stats_dev, allowFinishOffEvent,
```
agheata, line 843: I think we should replace, after the review, this ugly multi-parameter interface with a whiteboard object

include/AdePT/kernels/gammas.cuh
```diff
@@ -89,6 +90,16 @@ __global__ void TransportGammas(adept::TrackManager<Track> *gammas, Secondaries
       continue;
     }
 
+#ifdef ASYNC_MODE
+    if (InFlightStats->perEventInFlightPrevious[currentTrack.threadId] < allowFinishOffEvent[currentTrack.threadId] &&
+        InFlightStats->perEventInFlightPrevious[currentTrack.threadId] != 0) {
+      printf("Thread %d Killing gammas when killing the %d last particles of event %d E=%f lvol=%d after %d steps.\n",
```
agheata, line 96: This is, I guess, temporary, until we manage to send those particles back to the CPU

agheata (approved):
It will indeed be interesting to see how much of the 20% we can recover if we send the tracks back to the CPU. I made few comments, but this is goot to be merged.

agheata:
The failure is a CI configuration issue, the build&test finished OK

PR #362, opened 2025-03-18
[mini] Improve sync AdePT output

SeverinDiederichs:
Previously, sync AdePT was not passing the debug level, so no output depending on `/adept/verbose` could be generated.

This is fixed in this PR. Furthermore, the time and eventId is printed, which is useful to get plots like these:
Screenshot from 2025-03-18 10-44-16


JuanGonzalezCaminero (approved):
Looks good to me, thanks!

PR #364, opened 2025-03-18
Fix finish LastParticlesOnCPU

SeverinDiederichs:
This fixes the LastParticlesOnCPU option. Instead of killing the last n particles, they are now properly leaked to the CPU.

Finishing the last 100 particles of a ttbar event on CPU in the B field case is as fast as kiling them. Finishing the last particles accelerates the code significantly, the number of particles per event to be finished on CPU must be tuned.

Currently, the implementation works as follows:

The last n particles in flight of an event are leaked back to the CPU.
As soon as a particle is leaked back to the CPU, the leaking particle is flagged with status `fStopButAlive`. Then, the AdePTTrackingManager recognizes this, and sets a flag to keep the remainder of the event on the CPU, such that secondaries produced from one of the leaked particles also remain on the CPU.
When the first particle is handed to the AdePTTrackingManager with a different eventId (G4 started a new event), the flag to keep particles on the CPU is reset.


agheata (approved):
Thanks @SeverinDiederichs, LGTM!

PR #365, opened 2025-03-21
Setting step status when on boundary

JuanGonzalezCaminero:
Set the step status if on a boundary. 

This change should be paired in the future with returning all steps done in sensitive volumes, even if they don't deposit any energy.


SeverinDiederichs (approved):
Thanks for fixing this!

PR #367, opened 2025-03-22
hotfix for injection in Async AdePT

SeverinDiederichs:
For some corner cases, where EM only and single particle events are used, the EventState jumps from `G4RequestsFlush` past the `InjectionCompleted` although the injected particle was not yet visible on the device. That led to the event getting closed as the particles in flight were 0, and the particle was then associated with the next event, leading to energy deposition in the next event.

This HotFix is just a fast solution to this problem. The injection state can only go to InjectionCompleted after 25 (hardcoded number) iterations. This requires proper fixing in the future.

### DIFF
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -700,6 +700,10 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
     COPCORE_CUDA_CHECK(cudaStreamWaitEvent(waitingStream, cudaEvent));
   };
 
+  // needed for the HOTFIX below
+  int injectIteration[numThreads];
+  std::fill_n(injectIteration, numThreads, -1);
+
   std::unique_ptr<HitProcessingContext> hitProcessing{new HitProcessingContext{transferStream}};
   std::thread hitProcessingThread{&HitProcessingLoop,    (HitProcessingContext *)hitProcessing.get(),
                                   std::ref(gpuState),    std::ref(eventStates),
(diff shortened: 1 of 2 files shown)


agheata (approved):

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -789,20 +797,37 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
+            eventState.store(EventState::Inject, std::memory_order_release);
+            injectIteration[i] = iteration; // store iteration
           } else if (state == EventState::Inject) {
-            eventState = EventState::InjectionCompleted;
+            if (iteration - injectIteration[i] >= 25) {
+              eventState.store(EventState::InjectionCompleted, std::memory_order_release);
+              injectIteration[i] = -1;
+            } // else: wait more iterations
+          }
+          // If eventState leaves Inject unexpectedly, reset iteration tracking:
```
JuanGonzalezCaminero, line 819: This cannot happen, as there are (Or weren't) any other transitions from inject

include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -789,20 +797,37 @@ void TransportLoop(int trackCapacity, int scoringCapacity, int numThreads, Track
...
-        for (auto &eventState : eventStates) {
-          if (const auto state = eventState.load(std::memory_order_acquire); state == EventState::G4RequestsFlush) {
-            eventState = EventState::Inject;
+        for (int i = 0; i < numThreads; ++i) {
+          auto &eventState = eventStates[i];
+          const auto state = eventState.load(std::memory_order_acquire);
+
+          // HOTFIX:
+          // the current problem is that the injected particle takes too long before it shows up
+          // until then, the EventState has counted up and requested a Flush of the event, before the particle was
```
JuanGonzalezCaminero, line 806: Due to
https://github.com/apt-sim/AdePT/pull/367/files#diff-e7786b5f8e89172d219a16625b438e59af0580796e9f7c19de3a946cf65e8085R853
The CPU thread blocks until the injection is complete, so I don't think this counter has any effect (Since we only start counting **after** the copy has already finished) The delay is probably hiding the error somewhere else

PR #371, opened 2025-03-28
Propagating track weight inside AdePT

JuanGonzalezCaminero:
Part of the modifications required to run AdePT in ATLAS Athena, mirrors this change in G4HepEM: https://github.com/mnovak42/g4hepem/pull/108/commits/0d57bd8b4f6dfb9a73052689bd54338e377d6fe4

`CMS2018, 32 TTbar, 8 Threads, Transport on full detector`

|        | Sync    | Async   |
|--------|---------|---------|
| Master | 291.489 | 151.281 |
| New    | 290.848 | 151.887 |


agheata (commented):
This propagates weights OK, but we need to add  a unit test showing that it works. I mean example1 scoring should read the weights of incoming hits and multiply the edep with the weight. And we should have an input for example1 where we set all weights to like 0.5 and we make sure we get out half of the energy deposit

PR #372, opened 2025-04-01
This adapts to compacting the cuda  VecGeom library into the main one

agheata:
The MR https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/1296 makes a single VecGeom library, based on cuda_rdc in case of CUDA support. The default mode also changes from static to shared libraries. This PR adapts AdePT to this change and requires a VecGeojm shared libs build.

So this should not be merged until https://gitlab.cern.ch/VecGeom/VecGeom/-/merge_requests/1296  is merged AND the AdePT CI uses it.


agheata:
Overseded by https://github.com/apt-sim/AdePT/pull/376

PR #373, opened 2025-04-01
Cleanup of tests

JuanGonzalezCaminero:
Moving `IntegrationBenchmark` into the `test` directory as it is not meant to be an example. Moving old tests into `test/unit` and the current validation and reproducibility into `test/regression` so that everything is registered in the same place.

Changed the test scripts to output into a temporary directory to make cleanup less error-prone


SeverinDiederichs (approved):
Much cleaner now, thank you for the restructuring!

PR #374, opened 2025-04-01
Pass track weight to scoring

JuanGonzalezCaminero:
Add track weight to GPUHits and set it to the reconstructed track used for scoring


SeverinDiederichs (approved):
Thanks for fixing this!

PR #375, opened 2025-04-01
Upgrade to latest g4hepem

SeverinDiederichs:
This PR updates AdePT such that it can use the latest version of G4HepEm.
It enables to use the perRegion parameters in G4HepEm.

Furthermore, the positron cut is now used from G4HepEm. The ApplyCut value is taken directly from G4HepEm and not via a global constant. Some default parameters are set correctly. This fixes a bug, as before, if the `MinimalMSCStepLimit` was set, the results were wrong and it could even lead to stalls.

This PR will also require updating the CI to the latest tagged version of G4HepEm.

### DIFF
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -32,12 +32,15 @@
 #include <VecGeom/surfaces/cuda/BrepCudaManager.h>
 #endif
 
-#include <G4HepEmState.hh>
 #include <G4HepEmData.hh>
 #include <G4HepEmState.hh>
 #include <G4HepEmStateInit.hh>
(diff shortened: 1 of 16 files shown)


include/AdePT/core/AsyncAdePTTransport.cuh
```diff
@@ -412,17 +415,38 @@ void CopySurfaceModelToGPU()
 #endif
 }
 
-G4HepEmState *InitG4HepEm()
+G4HepEmState *InitG4HepEm(G4HepEmConfig *hepEmConfig)
```
JuanGonzalezCaminero, line 418: Since these calls are a bit more complex now we could think of eventually moving them to a common header

JuanGonzalezCaminero (approved):
Looks good to me, thank you for the changes!

PR #376, opened 2025-04-01
Adapt to CudaRdcUtils

pcanal:
This is an extension of https://github.com/apt-sim/AdePT/pull/372


agheata:
Thanks @pcanal ! I did practically the same but did not work for me. I'll try it tomorrow.

pcanal:
Thanks.  All those are necessary and the derived code *also* needs to use the `cuda_rdc_` (or be willing to jungle with sharp knifes and hope for the best :) ).  Any errors would indicates that we either missed a spot or (hopefully not) that there is still issue to be solved in `CudaRdcUtils`

agheata:
We merged in the master the rebased version of this. Thanks @pcanal !

PR #378, opened 2025-04-03
[mini] Add CPU only regions

SeverinDiederichs:
This adds the new UI command `/adept/removeGPURegion`.

If `/adept/setTrackInAllRegions true` is used, this command can be used to switch of certain regions via `/adept/removeGPURegion LastThird`. This makes it much easier in athena to turn off certain regions, instead of listing all others.


JuanGonzalezCaminero (approved):
Looks good to me, thanks for the changes!

PR #382, opened 2025-04-10
Fix memory problems in scoring

SeverinDiederichs:
This fixes a memory problem in the `ScoringObjects`:

Before, the struct had G4Tracks as members. Those G4Tracks were initialized by passing pointers to the `G4DynamicParticle`, which was allocated with `placement new`. Then, when the struct went out of scope, the destructor of G4Track was called, which tries to `delete` the G4DynamicParticle. However, since it wasn't allocated with `new`, this would cause a crash (interestingly only observed in Gauss).

This PR fixes the issue by using pointers to G4Tracks, such that when it goes out of scope, only the pointers are deleted and the destructor is not called. 

To make it more consistent, some parts were moved into the constructor of the object.

### DIFF
--- a/src/AdePTGeant4Integration.cpp
+++ b/src/AdePTGeant4Integration.cpp
@@ -66,7 +66,7 @@ struct ScoringObjects {
 
   ScoringObjects()
   {
-    // Step
+    // Assign step points in local storage and take ownership of the StepPoints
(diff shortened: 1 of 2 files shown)


agheata (approved):
Simple and clean, LGTM!

PR #384, opened 2025-04-10
Update validation reference data and use lower statistics for test

JuanGonzalezCaminero:
Thanks to https://github.com/apt-sim/AdePT/pull/381 our agreement with G4HepEM is a factor of 10 better. This means that we can afford to use lower statistics in our validation test for a faster CI.

This PR updates the reference data taken from a pure CPU G4HepEM run, and reduces the amount of events used in the validation test.

**Validation plot shooting 1e4 particles** (The new test run)
plot_1e4

**Validation plot shooting 1e5 particles**
plot


SeverinDiederichs (approved):
Thanks, very helpful!

PR #385, opened 2025-04-11
Minor B field improvements

SeverinDiederichs:
This PR adds two minor improvements to the B field:

1. If the RK integration fails below the minimal step, previously no action was taken and it would just retry this minimal step (and consequently, fail again). Now, if an integration step fails below the minimal step, we take the minimal step anyway, to advance the track. (Comparable to the QuickAdvance in G4).
2. When integrating the cords, we now re-use the last good step for the next cords. As an initial step, we choose the physics step length for the first cord. Previously, for every cord the physics step length was taken. If the physics step length was way too high for the B field integration, it would take 1-2 iterations before the correct length was found. Then, when integrating the next cord, it would again take 1-2 failed iterations before it would work. Now, for the next cord iteration, we re-use the last good step size, which typically immediately results in a good step.
For 128 ttbar events in CMS2018, this led to a 6% performance improvement. 

Additionally, minor cleaning is added in the B field.


agheata (approved):
Good!

PR #386, opened 2025-04-14
Initialize AdePT from G4 workers

SeverinDiederichs:
In the experimental frameworks, AdePT must be initialized by the G4 workers, because at the initialization time of the master thread, the number of G4 workers is not yet known.

Therefore, the initialization is postponed to the first G4 worker. Then, all other G4 workers must wait for the first worker to have initialized AdePT before they can proceed.

This fixes running with AdePT in Gauss and Athena.

### DIFF
--- a/src/AdePTTrackingManager.cc
+++ b/src/AdePTTrackingManager.cc
@@ -49,8 +49,9 @@ void AdePTTrackingManager::InitializeAdePT()
 
   auto tid = G4Threading::G4GetThreadId();
 
-  // Master thread cannot initialize AdePT as the number of G4 worker threads may not yet be known in applications
-  if (tid < 0) return;
+  // Master thread cannot initialize AdePT as the number of G4 worker threads may not yet be known in applications.
+  // In sequential mode, the tid is -2, so there we need to continue
+  if (tid == -1) return;
 
   // a condition variable and a mutex is used for the initialization:
   // The first G4 worker that reaches the initialization, needs to initialize AdePT.
@@ -93,10 +94,6 @@ void AdePTTrackingManager::InitializeAdePT()
     // Initialize common data:
     // G4HepEM, Upload VecGeom geometry to GPU, Geometry check, Create volume auxiliary data
     fAdeptTransport->Initialize(fHepEmTrackingManager->GetConfig(), true /*common_data*/);
-    if (sequential) {
-      // Initialize per-thread data (When in sequential mode)
-      fAdeptTransport->Initialize(fHepEmTrackingManager->GetConfig());
-    }
 #endif
 
     // common init done, can notify other workers to proceed their initialization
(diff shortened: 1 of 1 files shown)


agheata (approved):

src/AdePTTrackingManager.cc
```diff
@@ -86,27 +99,38 @@ void AdePTTrackingManager::InitializeAdePT()
...
+  }
+
+  // Now the fNumThreads is known and all workers can initialize
+  fAdePTConfiguration->SetNumThreads(fNumThreads);
+
 #ifdef ASYNC_MODE
-    fAdeptTransport = InstantiateAdePT(*fAdePTConfiguration, fHepEmTrackingManager->GetConfig());
+  fAdeptTransport = InstantiateAdePT(*fAdePTConfiguration, fHepEmTrackingManager->GetConfig());
 #else
+  if (!sequential) {
```
JuanGonzalezCaminero, line 120: Why does the sequential mode thread not instantiate AdePT?

SeverinDiederichs: It does, see line 98. This section is for the other workers that need to initialize their instance. The sequential worker would initialize here a second time

JuanGonzalezCaminero: At least from the previous initialization code it looks like in sequential mode there is only the master thread with `tid==-1`. If that's the case in sequential mode AdePT won't be initialized at all

SeverinDiederichs: I looked it up, in fact, in sequential mode there is only one thread with `tid == -2` (see here: https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/global/management/include/G4Threading.hh?ref_type=heads#L278). So you are correct, it would not be initialized at all. Thanks for pointing that out, will fix it!

src/AdePTTrackingManager.cc
```diff
@@ -46,13 +47,26 @@ void AdePTTrackingManager::InitializeAdePT()
...
+  // At the same time, all other G4 workers must wait for the common initialization to be finished, before they are
+  // allowed to continue
+  static std::once_flag onceFlag;
+  static std::mutex initMutex;
+  static std::condition_variable initCV;
+  static bool commonInitDone = false;
+
+  // Global initialization: only done once by the first worker thread
+  std::call_once(onceFlag, [&]() {
+    std::unique_lock<std::mutex> lock(initMutex);
```
JuanGonzalezCaminero, line 67: As we don't modify the condition variable, and this is run by the first thread only, is it necessary to acquire this lock here? I think that doing so means that the other G4 workers will be waiting at the lock rather than the condition variable (Making the cv redundant as `wait` will only be called when `commonInitDone` is true)

JuanGonzalezCaminero (approved):
Looks good to me, thanks for the changes!

PR #389, opened 2025-05-05
Temporary injection bug fix

JuanGonzalezCaminero:
This adresses the long-standing particle injection bug where some particles may be injected after their events are finished.

This bug only appears in specific runs, such as running the `testEm3` geometry with large events and a large number of leaks or oversubscribing the CPU threads. This PR solves it by synchronizing the particle injection stream at the end of every iteration.

While this fix solves the issue, it is not adressing the root  cause of the bug (As the logic doesn't require a synchronization), and should be removed once the underlying bug is understood and fixed.

This temporary fix doesn't cause any measurable slowdown in runs using the `CMS2018` geometry with ttbar events.


agheata (approved):
It looks to me that the state machine is a bit flawed, in the sense that the end-of-event signal must be conditioned by the buffer being actually flushed.

PR #391, opened 2025-05-20
Add nuclear processes in AdePT

SeverinDiederichs:
This PR adds the nuclear processes in AdePT.

This PR introduces a `LeakStatus`, which describes for which reason a particle is leaked (e.g., because it hit GammaNuclear or because it left the GPU region). Then, in the Geant4IntegrationLayer, when leaked tracks are handled and given back to Geant4, the nuclear processes are invoked. 
Therefore, the AdePTGeant4Integration layer now holds a pointer (not-owned) to the specialized `G4HepEmTrackingManager`, that is a member of the `AdePTTrackingManager`. The `G4HepEmTrackingManager` holds a pointer to the **thread-local** hadronic processes. Since there is only a single AdePTTransport instance in async AdePT, each thread must initialize its own Integration layer (this needed to be outsourced), such that it points to the correct `G4HepEmTrackingManager` (with the corresponding pointer to the thread-local hadronic process). 

Physics validation for async AdePT:
Screenshot from 2025-05-22 09-32-01

Physics validation for sync AdePT:
Screenshot from 2025-05-22 09-32-15

Performance measurement using 32 threads for 2000 events à 100 e- with 10 GeV being shot at testEm3:
| Mode   | with nuclear | without nuclear | cost (%) |
|--------|----------------|------------|-------------|
| Sync  AdePT | 233.403 s       | 218.304    | -6.92%      |
| Async AdePT | 163.869 s      | 163.406    | -0.28%      |
| G4HepEm | 377.132 s       | 377.184    | +0.01%      |

Note: Cannot be merged until we upgrade the CI to the latest version of G4HepEm, CI will fail at the moment.

To Do: 
- [x] Run high statistics validation for sync and async (1e7)
- [x] Update CI baseline file to include nuclear processes
- [x] performance measurements
- [x] do validation when not using nuclear processes in physics list
- [x] update CI to use latest G4HepEm, just waiting for the nightlies to push the new version


agheata (approved):
Nice, thanks for this important feature!

PR #387, opened 2025-04-16
Adopt cuda_rdc

agheata:
This overseeds https://github.com/apt-sim/AdePT/pull/376 rebasing and adding to the changes in the master

PR #393, opened 2025-05-22
Added non-consecutive reproducible track id

agheata:
A `uint_64` number is generated from the track seed at construction time. As long as the seed sequence is reproducible and non-clashing, so are the track id's.

To enable debugging of track N in the step range MIN-MAX: configure with ADEPT_TRACK_DEBUG=1; run like: ADEPT_DEBUG_TRACK=N ADEPT_DEBUG_MINSTEP=MIN ADEPT_DEBUG_MAXSTEP=MAX <executable> [args]

This has to be merged AFTER https://github.com/apt-sim/AdePT/pull/387


SeverinDiederichs (approved):
Great!

PR #398, opened 2025-06-12
Field cleanup

agheata:
This is an initial cleanup of the RK pusher. It introduces a couple of fixes and adds comments to the algorithm:
- Crossing a boundary in the first step now checks and detects back-scattering, in which case it returns rather than reducing the integration step.
- The last linear step was moving the track with the corrected arc fraction matching the chord intersection. This approximates better the length of the final arc step, but puts the point not on the boundary and biases the next step in the next volume
- The new direction after the RK advance was not normalized, generating errors downstream navigation

Note: there is also a printout improvement commit - if we want to keep this separate we can rebase instead of squash


SeverinDiederichs (approved):
Thanks for the fixes and cleaning!

PR #400, opened 2025-06-19
clean kernel and fix vecgeom version check

SeverinDiederichs:
This PR cleans / fixes 3 things:

1. a missing include of `<optional>` which was discovered with a different compiler
2. it cleans the macros for the surface model for the safety calculation (which were actually only needed to ensure the correct vecgeom version) in the kernels, which can now be done that VecGeom is tagged and the surface model merged.
3. It fixes the CMake version check logic. Due to string comparisons, all kinds of ugly misses happened. Now, it correctly identifies and rejects older versions and accepts correct versions.


agheata (approved):
Thanks for the cleanup!

PR #402, opened 2025-06-25
Tweaks for navigation in field.

agheata:
- The last exited state must be reset in the field propagator, if not done by `ComputeStepAndNextVolume` when finding the first parent not containing the point. This is needed because the track may return to the same volume due to BS or curvature.

- Added detection of a first step being zero in magnetic field, which can be used in future correlated with normal computation to signal back-scattering. Currently, the returned value is not used.

- Cleanup of pushes for readability in the field propagator.

- Tweaked some debug printouts.

Note: This PR removes the vast majority of tracks that get stuck at boundaries when used with VecGeom 2.0.0-rc5, but is also compatible with 2.0.0-rc4


JuanGonzalezCaminero:
The split kernels are currently not tested in CI when using a Bfield, but these changes should be implemented there too! (Only the parts outside `#if ADEPT_DEBUG_TRACK > 0` as the debug code is not implemented in the split kernels)

SeverinDiederichs (approved):
Thanks for the fixes!

PR #403, opened 2025-06-26
Improve RNG state

SeverinDiederichs:
This PR improves the RNG handling in 3 ways:

1. the initial seed was depending only on the AdePT initial seed, the event id, and the track id. Thus, if a particle left the GPU region and was injected again, it would have received the same seed. This is fixed by generating the seed by scrambling the original seed with the kinetic energy and the global time.

2. When the new ID was generated from the rng state without advancing it, there must have been some truncation because multiple collisions in the trackIds were observed at a rate way too high. This is solved by advancing the state.

3. When a positron fell below the tracking cut, it would annihilate to two gammas. However, this could happen after the first interaction, meaning the newRNG was already used on the first generated secondary. This would lead to collisions of state. 

There are still too many collisions observed, which are currently under investigation.

The failing of the CI was just a random fluctuation, the accuracy is not impacted by this PR.

### DIFF
--- a/include/AdePT/copcore/Ranluxpp.h
+++ b/include/AdePT/copcore/Ranluxpp.h
@@ -83,21 +83,6 @@ public:
     return bits;
   }
 
-  /// Return random bits matching the current state without advancing it
-  __host__ __device__ uint64_t GetRandomBits() const
-  {
-    int idx     = fPosition / 64;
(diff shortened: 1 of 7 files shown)


agheata (approved):
Thanks for this important fix/feature

PR #409, opened 2025-07-07
Add CPU-only track information

SeverinDiederichs:
This PR proposes to split the G4 track data into a GPU-track part that is handled by AdePT and a CPU-track part that is kept on the CPU. This way, we can handle all kinds of custom user information, as well as reduce our track size.

This works as follows:

Whenever a particle is send from G4 to GPU, it creates an entry in the `HostTrackDataMapper` where it stores information such as the vertex origin, creator process, or G4VUserTrackInformation. There is a lockup map to map from the uint_64 AdePT ID to an index for vector containing the data. This vector then holds all relevant information and is easily extendable. 
Whenever a secondary is created, the initial information of the new secondary must be send back to the CPU (this is also done with the RecordHit interface). Then, in the `FillG4Step` function, it is checked whether it was an initializing step (with stepCounter = 0) or a later step of the track. In the first case, a new entry is created in the `HostTrackDataMapper` and set with all the relevant information. The PreTrackingAction is called and the user defined G4VUserTrackInformation is stored in the hosttrackdata. For later steps, then this can be looked up.

This requires a strict ordering of the steps: Whenever a track returns not the first step or has a leaked track, the HostTrackData **must** already exist. If it does not, it is a bug. In all currently tested settings, this works.
Then, to keep the track data structure small (we only need to store particles in flight), we delete the entry if it was the last step of the track. This requires that the initializing step of any daughter track is handled **before** the final step of the parent track.

This new approach requires that all the initializing steps must be returned to the CPU. Thus, as soon as a secondary is created, recordHit is called. Note that this significantly increases the run time for settings where particles do very few steps, i.e., if a particle does one step and then dies, it doubles the amount of data that must be send back to the CPU, which has a significant performance impact on simple examples like testEm3. Therefore, this is guarded and done only if the G4TrackingAction is called (i.e., if the flag `/adept/CallUserTrackingAction true`) is used.

Otherwise, this approach allows the reduce the track size (things like vertex information etc do not need to be stored on the GPU anymore) and the information that is send back to the CPU per step (everything that is in the CPU-only track data).

The run time impact on testEm3 looks like this:
| Configuration        | NEW (s) | OLD (s) |
|----------------------|---------|---------|
| Not sensitive        | 10.0291 | 13.743  |
| All sensitive        | 14.5    | 20.1437 |
| Call tracking action | 33.5978 |   –     |

For 128 ttbar events with 16 threads with a B field in CMS2018, the slowdown of using the full track information is ~8% (1074.5 s vs  989.6 s)

Note that the impact of sending back the initializing step is hardly visible in a more complicated setting, i.e., a proper geometry such as CMS or ATLAS and a B field, as then the copy can be overlapped with the compute.

NOTE: the new implementation has problems if track IDs collide. Also, it allows to check for collisions:
Before creating a new entry, if the ID already has an entry, there must be a collision.
With this, it was found that the previous hashing to generate the uint64 track id was not sufficient. Additionally, the way that the AdePT uint64 track id is generated was changed, to truly generate a uint64 track id and not a 48 bit one via ranlux++, was it was done before. 

Next steps that are left for future PRs: there are a few things that can be improved now:
1. the `originNavState` stored on the track can be moved to the CPU-only data. However, storing the smart pointer of the touchable handle takes a bit of effort
2. The creatorProcess stored on the device should be the step-defining process, as that one is needed for CMS. Then, for the initializing step, the step-defining process is the creator process and for other steps the step-defining process.
3. The RecordHit function has many arguments and is called many times, this can be improved by giving the full track as an argument.

In this PR, the ADEPT_TRACK_DEBUG is also enabled for sync AdePT. 
Furthermore, a bug is fixed that caused undefined behavior on the GPU, when the boolean input flag `returnLastStep` was changed (but by each thread individually).

### DIFF
--- a/include/AdePT/core/AdePTTransport.cuh
+++ b/include/AdePT/core/AdePTTransport.cuh
@@ -364,8 +364,10 @@ GPUstate *InitializeGPU(adeptint::TrackBuffer &buffer, int capacity, int maxbatc
   COPCORE_CUDA_CHECK(cudaMalloc(&gpuState.stats_dev, sizeof(Stats)));
   COPCORE_CUDA_CHECK(cudaMallocHost(&gpuState.stats, sizeof(Stats)));
 
+#if ADEPT_DEBUG_TRACK > 0
   // initialize track debugging
   InitializeTrackDebug();
+#endif
 
   // initialize buffers of tracks on device
   COPCORE_CUDA_CHECK(cudaMalloc(&gpuState.toDevice_dev, maxbatch * sizeof(TrackData)));
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -749,8 +749,10 @@ std::unique_ptr<GPUstate, GPUstateDeleter> InitializeGPU(int trackCapacity, int 
   // init scoring structures
   gpuMalloc(gpuState.fScoring_dev, numThreads);
 
+#if ADEPT_DEBUG_TRACK > 0
   // initialize track debugging
   InitializeTrackDebug();
+#endif
 
   scoring.clear();
   scoring.reserve(numThreads);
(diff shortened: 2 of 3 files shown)


JuanGonzalezCaminero (approved):
Looks good to me, thanks for the PR and detailed explanation!

PR #410, opened 2025-07-09
only thread that allocated may deallocate common data in Sync AdePT

SeverinDiederichs:
This fixes an issue in Sync AdePT that probably stems from #386 but could be even older:

In Sync AdePT, one thread initializes the common data on the GPU, however, all threads would try to delete them in the destructor, causing double frees and crashes.

This is fixed by flagging which thread did the common init and only that thread deallocates the common data in the destructor.


JuanGonzalezCaminero (approved):
Looks good to me!

PR #413, opened 2025-07-10
Add Step-limiting process

SeverinDiederichs:
This adds the step-limiting process to the G4 step information when re-creating the G4 steps to call the SD code.

The creator process ID could be removed from the track.

Now, the step-limiting process is returned with recordHit. If it is an initializing (0th) step, it is the creator process, if it is a later step, it is the step-limiting process.


JuanGonzalezCaminero (approved):
Thanks for the changes!

PR #416, opened 2025-07-15
cleaner TrackingManager initialization via constructor

SeverinDiederichs:
In Athena, an issue was observed that the fAdePTConfiguration was not correctly set. The issue was tracked down to the fAdePTConfiguration being a garbage pointer despite the setter being called right after the constructor. The problem was solved by setting the fAdePTConfiguration directly in the constructor.


JuanGonzalezCaminero (approved):
Looks good to me, thanks!

PR #418, opened 2025-07-16
Reduce block size for transport kernels

JuanGonzalezCaminero:
Reducing the block size from the previous 256 to 32 provides a speedup in Bfield runs, while not affecting runs without a field. This has been tested on two different GPUs:

```
Nvidia RTX 4080
CMS 2018, 8 Threads x 32 TTbar events
```

**No field**
| Kernels    | Time 256 | Time 32 |
|------------|----------|---------|
| Monolithic | 177      | 175     |
| Split      | 182      | 181     |

**3.8 T Field**
| Kernels    | Time 256 | Time 32 | Speedup |
|------------|----------|---------|---------|
| Monolithic | 339      | 292     | 1.16    |
| Split      | 348      | 296     | 1.17    |

```
Nvidia RTX 4090
CMS 2018, 16 Threads x 128 TTbar events
```
**3.8 T Field**
| Kernels    | Time 256 | Time 32 | Speedup |
|------------|----------|---------|---------|
| Monolithic | 856      | 600     | 1.42    |
| Split      | 928      | 718     | 1.29    |

### DIFF
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -903,14 +903,16 @@ void TransportLoop(int trackCapacity, int leakCapacity, int scoringCapacity, int
                                   std::ref(cvG4Workers), std::ref(debugLevel)};
 
   auto computeThreadsAndBlocks = [](unsigned int nParticles) -> std::pair<unsigned int, unsigned int> {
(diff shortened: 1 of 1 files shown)


SeverinDiederichs (approved):
Good finding! It would be nice to also test it on an A100 or H100 at some point.

PR #419, opened 2025-07-17
adjust B field parameters to G4

SeverinDiederichs:
This PR renames a variable in AdePT so that it has the same name as the one in G4. gEpsilonDeflect -> deltaChord.

Also, this PR adds some more explanations to the variables and sets the same defaults as G4 (they were more strict in AdePT before)


JuanGonzalezCaminero (approved):
Thank you for these changes!

PR #421, opened 2025-07-28
Cleaning macros

SeverinDiederichs:
This PR cleans the macros that we generate in example1. The previous distinction between the "large_stack" and the normal file was not helpful, as the normal file was not able to run without changing the stack size in it. So this PR proposes to just delete the large stack files and use the larger stack in the normal file right away.

It also deletes a CI test which only tested that example1 runs. However, since our reproducibility test already tests that, it was redundant. Arguably, the same is true for the example1-g4vg test, I just kept it in so we have one test that runs example1 and one that runs the integrationBenchmark. 

Also, I noticed some deprecated, outcommented piece of code that was just deleted.


agheata (approved):
Thanks for the cleanup!

PR #422, opened 2025-07-28
Enable mixed precision in Bfield integration

Krafpy:
- Renamed ADEPT_USE_SURF_SINGLE to ADEPT_MIXED_PRECISION flag
- Added precision type for RK integration based on flag
- Hardcoded some vecgeom::Precision variables to double type


JuanGonzalezCaminero:
Trying to compile locally, I get an error since this is called with Real_t = float, but RkDriver_t has been instantiated with Real_t = double (Vecgeom::Precision):

https://github.com/apt-sim/AdePT/pull/422/files#diff-dfecf5ea9ab1a85844010e343a742bfe03917ac79ea4779bc1477ef53773a185R247

The call to `Advance` then fails because `dydx_end` is a `float[]` but it expects `double[]`

https://github.com/apt-sim/AdePT/blob/master/include/AdePT/magneticfield/fieldPropagatorRungeKutta.h#L176

SeverinDiederichs:
> Trying to compile locally, I get an error since this is called with Real_t = float, but RkDriver_t has been instantiated with Real_t = double (Vecgeom::Precision):
> 
> https://github.com/apt-sim/AdePT/pull/422/files#diff-dfecf5ea9ab1a85844010e343a742bfe03917ac79ea4779bc1477ef53773a185R247
> 
> The call to `Advance` then fails because `dydx_end` is a `float[]` but it expects `double[]`
> 
> https://github.com/apt-sim/AdePT/blob/master/include/AdePT/magneticfield/fieldPropagatorRungeKutta.h#L176

Note that this happens when compiling the `test_magfieldRK.cpp`.

Also, the real CI is still not running, this seems to be a permission issue, which we need to fix.

SeverinDiederichs (approved):
Thank you for this addition!

A few new stuck tracks were observed in Athena when running with `-DADEPT_MIXED_PRECISION=ON`, those will be debugged separately.

PR #427, opened 2025-08-13
Enable using fNumThreads from AdePTconfig

SeverinDiederichs:
G4 applications don't necessary need to use a G4RunManager. In that case, AdePT can retrieve the number of threads via the AdePTConfig, which is passed to the constructor of the AdePTTrackingManager. Then, in the physicslist, the number of threads must be passed via the AdePTConfig to the AdePTTrackingManager. Needed in CMSSW.

This PR adds a safe guard for the possible nullptr access for the G4RunManager, sets a default value for fNumThreads and deletes an used variable.


agheata (approved):
Thanks for the improvement

PR #429, opened 2025-08-18
Refactoring the monolithic kernels for potential stepping action

SeverinDiederichs:
This PR refactors the monolithic kernels to enable a central spot for implementing a custom stepping actions for experiments.

As the experiment's frameworks have their own custom stepping actions where they decide to kill unwanted tracks, it would be beneficial if AdePT supports implementing such a stepping action in a straightforward way.

Currently, the `survive()` function was called in multiple places in the monolithic kernels. Therefore, this PR suggests to refactor the kernels such that only a surviveFlag is set to true if a track survives such that before the actual `survive` call at the end of the kernel, a stepping action could be added.

Note that this change originally did affect the run time as it forced some synchronization under the hood. After adding a few `__syncwarp()` after divergent parts of the code, the run time is recovered.

For 32 threads and 64 ttbar events, the run time for master was `126.095 s` and for this PR `124.882 s`. The results are unchanged, as the RNG or physics are not touched in this refactoring.

Note, the PR should be reviewed with `hide white space = on` as the indentation in the gamma kernel needed to be changed, so it looks like more changes than there actually are.

It has to be seen if this is feasible also in the split kernels, but seems unlikely.

### DIFF
--- a/include/AdePT/kernels/electrons.cuh
+++ b/include/AdePT/kernels/electrons.cuh
@@ -367,6 +367,8 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
       currentTrack.zeroStepCounter = 0;
     }
 
+    __syncwarp(); // was found to be beneficial after divergent calls
+
 #if ADEPT_DEBUG_TRACK > 0
     if (verbose) {
       printf("| geometryStepLength %g | propagated_pos {%.19f, %.19f, %.19f} dir {%.19f, %.19f, %.19f}",
@@ -375,8 +377,6 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
       nextState.Print();
     }
 #endif
-
-    __syncwarp(); // was found to be beneficial after divergent calls
 
     // punish miniscule steps by increasing the looperCounter by 10
     if (geometryStepLength < 100 * vecgeom::kTolerance) currentTrack.looperCounter += 10;
--- a/include/AdePT/kernels/electrons.cuh
+++ b/include/AdePT/kernels/electrons.cuh
@@ -132,6 +132,8 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
 
 #endif
     bool isLastStep                          = returnLastStep;
+    bool surviveFlag                         = false;
+    LeakStatus leakReason                    = LeakStatus::NoLeak;
     constexpr Precision kPushStuck           = 100 * vecgeom::kTolerance;
     constexpr unsigned short kStepsStuckPush = 5;
     constexpr unsigned short kStepsStuckKill = 25;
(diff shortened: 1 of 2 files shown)


include/AdePT/kernels/electrons.cuh
```diff
@@ -175,7 +171,7 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
       currentTrack.globalTime = globalTime;
       currentTrack.localTime  = localTime;
       currentTrack.properTime = properTime;
-      currentTrack.navState   = navState;
+      currentTrack.navState   = nextState;
```
JuanGonzalezCaminero, line 174: Why does this change?

include/AdePT/kernels/electrons.cuh
```diff
@@ -390,6 +376,8 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
     }
 #endif
 
+    __syncwarp(); // was found to be beneficial after divergent calls
```
JuanGonzalezCaminero, line 379: Since this sync is related to the push, I think it should be moved before the printout

include/AdePT/kernels/electrons.cuh
```diff
@@ -976,14 +947,64 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
...
+                 "geoStepLength=%E "
+                 "physicsStepLength=%E "
+                 "safety=%E\n",
+                 eKin, currentTrack.eventId, currentTrack.trackId, currentTrack.looperCounter, energyDeposit,
+                 geometryStepLength, geometricalStepLengthFromPhysics, safety);
+        surviveFlag = false;
+      }
+    }
+
+    if (surviveFlag) {
```
JuanGonzalezCaminero, line 970: I think all of these should go inside a single `if (surviveFlag)`, including the code inside the `#ifdef ASYNC_MODE`

include/AdePT/kernels/electrons.cuh
```diff
@@ -1003,30 +1024,6 @@ static __device__ __forceinline__ void TransportElectrons(adept::TrackManager<Tr
                                isLastStep,                                  // whether this was the last step
                                currentTrack.stepCounter);                   // whether this was the first step
     }
-    if (cross_boundary) {
-      // Move to the next boundary now that the Step is recorded
```
JuanGonzalezCaminero, line 1007: Why is this code removed? Is there no issue with moving to the next boundary before recording the step?
In any case, think the code is cleaner if we do the change of `navState = nextState` here, instead of inside the `survive` function, since it is not obvious that `survive` should have that functionality.
Is it not possible to keep this code and call `survive` at the end of the kernel?

include/AdePT/kernels/electrons.cuh
SeverinDiederichs: Discussed offline

PR #430, opened 2025-08-21
Fix access to HostTrackData

SeverinDiederichs:
This PR fixes a bug in the access of the HostTrackData, so only a local copy was modified and not the stored data.
This led to many tracks using the default constructed HostTrackData. The fix does not change the run time. This was found during the LHCb validation, as the joining of hits with their parents did not work:

Screenshot from 2025-08-21 14-01-14

Now it does:
Screenshot from 2025-08-21 13-59-57

- [x] test in LHCb


JuanGonzalezCaminero (approved):
Looks good to me!

PR #435, opened 2025-09-04
Replace overlooked vecgeom::Precision with rk_integration

JuanGonzalezCaminero:
An overlooked `vecgeom::Precision` prevented compilation of the split kernels in mixed-precision mode in some machines.


SeverinDiederichs (approved):
Thanks for the fix!

PR #441, opened 2025-09-25
[mini] Fix proper time

SeverinDiederichs:
Reported by @dkonst13 :

The proper time was not calculated correctly in AdePT:

It should be  $t_{proper} += \Delta_t * m_e / E_{tot}$ with $E_{tot} = m_e + E_{kin}$.
See also in G4 [here](https://gitlab.cern.ch/geant4/geant4/-/blob/master/source/processes/transportation/src/G4Transportation.cc?ref_type=heads#L553)


agheata (approved):
Nasty one, good catch!

PR #442, opened 2025-09-26
[mini] fix track data == operator

SeverinDiederichs:
Found thanks to static analysis by @dkonst13:

The == operator in the `TrackData` was bugged:

`!(a < b && b < a)` returns always true. In case of `a < b` (or `b < a`), it yields `!(true && false) -> !(false) -> true`

Correct is: `!(a < b) && !(b < a);`, which yields true only if `a == b`.


JuanGonzalezCaminero (approved):

agheata (approved):
Good catch

PR #443, opened 2025-10-02
Consumption benchmarking

JuanGonzalezCaminero:
Adds the `ENABLE_POWER_METER` cmake option in order to do power consumption measurements.

If the option is enabled, we use the [power meter](https://github.com/apt-sim/power_meter) library to launch a background thread taking measurements from CPU and GPU. The monitoring is started once AdePT initialization is finished, and ends when shutting down the application.

Note that if the option is enabled, simulations need to be run with root access, which is required in order to read the CPU's MSR registers.

Prior to Geant4 11.3 the specialized tracking managers were not deleted by Geant4. Since the power meter is stopped in the tracking manager destructor, this will lead to an error message in older geant4 versions when the program ends without joining the power meter thread. This should not cause a memory leak, since `terminate()` will be called on the unjoined thread.

### DIFF
--- a/cmake/AdePTConfig.cmake.in
+++ b/cmake/AdePTConfig.cmake.in
@@ -38,6 +38,10 @@ if(@ADEPT_USE_EXT_BFIELD@)
   find_dependency(covfie REQUIRED)
 endif()
 
+if(@ADEPT_ENABLE_POWER_METER@)
+  find_dependency(Power_meter REQUIRED)
+endif()
+
 # Record whether we support G4VTrackingManager.
 # Note that this implies the re-found Geant4 below is the exact same as we
 # were built with, especially if G4VTrackingManager has been backported to < 11.0


src/AdePTTrackingManager.cc
```diff
@@ -40,6 +44,13 @@ AdePTTrackingManager::AdePTTrackingManager(AdePTConfiguration *config, int verbo
 
 AdePTTrackingManager::~AdePTTrackingManager()
 {
+  auto tid = G4Threading::G4GetThreadId();
```
SeverinDiederichs, line 47:
```suggestion
```

Unused variable that caused a warning

SeverinDiederichs (commented):
Very nice!

See minor comment to remove unused variable warning.

Can merge after we successfully ran it in the frameworks

PR #444, opened 2025-10-03
[mini] initialize (empty) secondary vector

SeverinDiederichs:
This initializes an empty secondary vector when returning a GPU step to the CPU. This is needed such that the access to the secondary does not crash:
```
  const std::vector<const G4Track*>  *secondaryVector = aStep->GetSecondaryInCurrentStep();
  secondaryVector->size(); <- currently crashes as it is a nullPtr
```
With this PR the secondary vector is available, just empty.

This PR is needed to run Athena with the `Sim_tf` transform.


SeverinDiederichs:
With the latest Athena version, this does not seem to be needed for now, but I'll keep it open for a bit to confirm with the other experiment frameworks that indeed it is no problem.

With this PR, the check for secondaries in a step would not be a nullptr access anymore, preventing potential crashes.

PR #448, opened 2025-10-21
Fix for the surface model in the field propagator

JuanGonzalezCaminero:
The field propagator uses the `GetLastExited()` method from the navigation state. This doesn't work with the surface model as this method returns a PlacedVolume. With this fix the correct method is called, which also avoids a different call for solids and surfaces.


agheata (approved):
Thanks!

PR #450, opened 2025-10-28
[mini] move position when step size advance fails

SeverinDiederichs:
This mini PR fixes a bug reported by code analysis from @dkonst13:

The the loop of reducing the step size when stuck on the boundary, the `move` was adjusted, but the position was not changed. As the `move` affects the remains and the stepDone, the position must be also adjusted here, otherwise it is not consistent.


SeverinDiederichs:
CI failed in reproducibility in sync AdePT, a longer standing and rare issue. Just restarted it.

agheata (approved):
OK, it looks like the addition is correct and may prevent some undesired lock on boundary. Thanks for the fix!

PR #452, opened 2025-10-29
Removed leftover setting of last exited state affecting relocation.

agheata:
This removes a line probably left over during back-scattering debugging. I see no more stuck electrons in EMEC in a small test.


SeverinDiederichs (approved):
Thanks for fixing this bug!

PR #453, opened 2025-10-31
Add extendable, experiment-specific SteppingAction

SeverinDiederichs:
This PR adds extendable, experiment-specific SteppingActions.

They can be enabled via `-DADEPT_STEPPINGACTION=LHCb|CMS`. So far, only the SteppingAction for LHCb and CMS are implemented, but further actions could be added.

A struct with parameters for the stepping action is passed to the kernel (it is specific to the experiment), and the kernels are templated over the SteppingAction.
Then, the kernel calls the correct stepping action.

No run time difference was observed in the LHCb minbias runs.

Note that so far this is only for the monolithic kernels, the split kernels would first need a setup PR similar to #429

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -459,6 +459,11 @@ if(ADEPT_USE_EXT_BFIELD)
   )
 endif()
 
+# must be defined when compiling AdePT_G4_integration_objects' source files into the objects lib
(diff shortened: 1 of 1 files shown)


JuanGonzalezCaminero (approved):
Looks good as discussed, thank you for this important addition!

PR #459, opened 2025-11-21
[mini] clean FillFromDeviceBuffer

SeverinDiederichs:
This PR cleans `FillFromDeviceBuffer` kernel:

When looping over the particles to be copied back, the index starts at `nAlreadyTransfered` and was going to  `nAlreadyTransferred + maxFromDeviceBuffer`. Via an if statement, out of bounds access was prevented. 
This PR proposes to directly fix the range correctly in the loop, enabling to remove the out of bounds check later. Thus, this is some minor cleaning PR

Found via code analysis by @dkonst13.


SeverinDiederichs:
CI is failing because the shipped VecGeom misses some files like `VecGeomConfigVersion.cmake`. Under investigation

JuanGonzalezCaminero:
I think this is not an issue because the index is checked before accessing the buffer:
https://github.com/apt-sim/AdePT/blob/master/include/AdePT/core/AsyncAdePTTransport.cuh#L278

Otherwise it would be crashing

PR #464, opened 2025-12-16
[mini] remove passing of variables that are already captured

SeverinDiederichs:
This small PR cleans the `survive` function in the monolithic kernels.

Variables were passed to `survive` but since it captures variables of the same name via `[&]` it was shadowing.
Before, it was functionally correct, but not clear. Now, just the captured variables are used.


agheata (approved):
Thanks for the cleanup. We should consider having a method `Track::Update(ekin, pos, dir, ...)`

PR #469, opened 2026-01-19
Cleaning in split kernels as preparation for SteppingAction

SeverinDiederichs:
This PR does some minor cleaning and a few fixes in the split kernels as a preparation for enabling the SteppingAction in the split kernels as well:

- added the missing freeing of slots when particles are killed due to loopers or stuck tracks
- use `currentTrack.nextState` instead of `currentTrack.navState` to get the `nextlvolID`. This is clearer as then it does not depend on the code ordering whether `navState` was already updated or not.
- remove the template `IsElectron` from `PerformStoppedAnnihilation` as this is done only for positrons. Move the `MarkSlotForFreeing` outside of the function, such that it is called by electrons and positrons alike (as electrons won't call it anymore).
- remove the setting of the (at this point) unused `currentTrack.stopped = true` (it is only used in the setup of the interactions).
- remove unused `survive` function definitions
- use `continue` instead of `break`: as there is no switch statement (unlike the monolithic kernels), this would have caused the full loop over tracks to exit, which would be a serious bug. It was unnoticed, as the low energy if-statements are for safety only and practically never triggered.

### DIFF
--- a/include/AdePT/kernels/electrons_split.cuh
+++ b/include/AdePT/kernels/electrons_split.cuh
@@ -824,7 +824,7 @@ __global__ void ElectronIonization(G4HepEmElectronTrack *hepEMTracks, ParticleMa
       if (IsElectron) {
         energyDeposit += currentTrack.eKin;
       }
-      if (IsElectron) {
+      if (!IsElectron) {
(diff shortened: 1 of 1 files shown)


JuanGonzalezCaminero (approved):
Looks good to me, thanks for the cleanup!

PR #484, opened 2026-02-06
add Woodcock tracking to split kernels

SeverinDiederichs:
This PR adds the Woodcock tracking also to the split kernels.

- A new kernel `GammaWoodcock` is added. It is called for the gamma woodcock queue instead of `GammaHowFar` and `GammaPropagation`. However, it does feed into `GammaSetupInteractions`. Note: it is ensured that all gammas out of `GammaWoodcock` hit one of the interactions. In case of a delta interaction, they are enqueued again into the gamma Woodcock queue, in case of a relocation, the relocation is done and they are relocated into the normal gamma queue. Thus, in the `GammaSetupInteractions`, they should never be queued for `GammaRelocation`
- A new check is introduced: if no Woodcock tracking is enabled, the Woodcock tracking kernels are not called.
- The CI tests for split kernels are extended to test also for Woodcock tracking. Note that the reproducibility tests are enabled in general
- There is very good physics agreement between Woodcock tracking with monolithic kernels and Woodcock tracking with the split kernels:
 
Screenshot from 2026-02-09 10-27-55
Screenshot from 2026-02-09 10-27-48

It was verified that this PR
- [ ] Changes physics results
- [x] Does not change physics results. Clarification: this does not change the physics results running with split kernels and without Woodcock tracking

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -139,7 +139,7 @@ def buildAndTest() {
     ctest -V --output-on-failure --timeout 2400 -S AdePT/jenkins/adept-ctest.cmake,$MODEL
     export CMAKE_BINARY_DIR=BUILD_SPLIT_ON
     export ExtraCMakeOptions="-DADEPT_USE_SPLIT_KERNELS=ON -DADEPT_BUILD_TESTING=ON"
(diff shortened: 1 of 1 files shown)


JuanGonzalezCaminero (approved):
Looks good to me, thank you!

PR #494, opened 2026-02-26
add regression test vs master

SeverinDiederichs:
This PR adds a regression test suite against the current master branch. In these short tests, 20 10 GeV electrons are hitting the testEm3  geometry and the physics output is compared against that of the current master branch (with 0 tolerance on any changes).

Similar to the validation, there are 4 variations:
The default setting, using Woodcock tracking, using the GPU only in certain regions, and (freshly added here), turning on a constant B field.

These 4 tests are run for all 3 build configurations: monolithic kernels, split kernels, and monolithic + mixed precision, thus this PR adds 12 tests in total.

The tests are rather short:

```
    Start 17: physics_drift_default
1/4 Test #17: physics_drift_default ............   Passed   10.87 sec
    Start 18: physics_drift_regions
2/4 Test #18: physics_drift_regions ............   Passed   13.32 sec
    Start 19: physics_drift_wdt
3/4 Test #19: physics_drift_wdt ................   Passed   11.54 sec
    Start 20: physics_drift_bfield
4/4 Test #20: physics_drift_bfield .............   Passed   10.75 sec
```

In case a test fails, this does not invalidate the PR or the CI. This simply means that the physics output were changed in that PR. Therefore, the result will be marked as unstable, and the full validation suite will be performed. If the exact physics output passes, the full physics validation is skipped.
E.g., advancing the RNG in the Woodcock tracking kernel will lead to:

```
3/4 Test #19: physics_drift_wdt ................***Failed   11.58 sec
Cleaning up temporary files...
Step 'compare_master_pr_outputs_wdt' failed. Command output:
The results are not reproducible. Differences found
Column ' 13': File1 = 160.525274, File2 = 156.058061
Column ' 14': File1 = 71.48516, File2 = 74.402084
```

and the result is marked as:

```

physics_drift mismatches detected for MIXED_PRECISION
physics_drift detected differences in at least one configuration.
[Pipeline] unstable
WARNING: physics_drift differences detected against master (non-blocking)
[Pipeline] End of Pipeline
Finished: UNSTABLE
```

This should provide a solid protection against unwanted changes in the physics output.

With this new simple drift tests for exact physics reproducibility w.r.t to the master branch in place, the validation vs the benchmark and longer reproducibility tests are only performed when the fast drift tests return as unstable (i.e., changed physics output). 
The full validation and reproducibility tests are still always performed during the nightlies.

It was verified that this PR
- [ ] Changes physics results

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -247,12 +247,9 @@ def checkoutReferenceSource() {
       $class: 'GitSCM',
       branches: [[name: '*/master']],
       doGenerateSubmoduleConfigurations: false,
-      extensions: [
-        [$class: 'CleanBeforeCheckout'],
-        [$class: 'SubmoduleOption', parentCredentials: true, recursiveSubmodules: true]
-      ],
       userRemoteConfigs: [remoteConfig]
     ])
+    sh 'git submodule update --init'
   }
 }
 


jenkins/adept-ctest-build.cmake
```diff
@@ -108,34 +87,27 @@ foreach(v
...
 
 include("${CTEST_SOURCE_DIRECTORY}/CTestConfig.cmake")
 
-# ctest_empty_binary_directory--------------------------------------------------
+# clean build tree for this dashboard slot--------------------------------------
 file(REMOVE_RECURSE ${CTEST_BINARY_DIRECTORY})
 
-# Execute all the steps---------------------------------------------------------
 ctest_start(${CTEST_MODEL} TRACK ${CTEST_MODEL})
-ctest_update(SOURCE ${CTEST_SOURCE_DIRECTORY})
```
JuanGonzalezCaminero: Why was the update step removed?

JuanGonzalezCaminero (approved):
Thank you for the changes, looks good to me!

PR #497, opened 2026-03-02
Clean AdePTPhysics definition

SeverinDiederichs:
This improves and clarifies the `AdePTPhysics`:

Previously, the `AdePTPhysics` were added as an additional physics constructor to the PhysicsList, right after the G4EmStandard physics. However, this gave the impression that one could add them similarly to other G4Em physics lists, but that is not the case, as the `AdePTPhysics` constructor was calling this:
```c++
  G4EmParameters *param = G4EmParameters::Instance();
  param->SetDefaults();
  param->SetVerbose(1);
```
resetting all parameters that other options might have set.

Thus, the clean way is to add the `AdePTTrackingManager` directly inside an `G4EmStandardPhysics_AdePT` and `_HepEm`. This can be done in a relatively clean way by deriving directly from the `G4EmStandardPhysics`.

Now, it is clear that the AdePTPhysics list only replaces the G4EmStandardPhysics and is not equivalent with any other physics constructors.
This could be extended to other physics constructors as well. 

NOTE: This PR **does** change the physics output and therefore is marked as unstable in the drift test.
In fact, the new implementation uncovers the issues of the original approach:

In `G4EmStandardPhysics` the following flags are set:
```c++
UseGeneralProcess = true
FluctuationType = Urban
```
however, the `  param->SetDefaults();` from the previous  `AdePTPhysics` sets it to 
```c++
UseGeneralProcess = false
FluctuationType = Universal
```
The drift test passes again if those flags are enforced as a regression via:
```
/process/em/UseGeneralProcess false
/process/eLoss/setFluctModel Universal
```
in the macro.

It was verified that this PR
- [x] Changes physics results
- [ ] Does not change physics results

### DIFF
--- a/src/G4EmStandardPhysics_AdePT.cc
+++ b/src/G4EmStandardPhysics_AdePT.cc
@@ -26,6 +26,7 @@ void G4EmStandardPhysics_AdePT::ConstructProcess()
   // First register the standard Geant4 EM processes for this constructor.
   G4EmStandardPhysics::ConstructProcess();
 
+  // Register custom tracking manager for e-/e+ and gammas.
   fTrackingManager = new AdePTTrackingManager(fAdePTConfiguration, /*verbosity=*/0);
 
   auto g4hepemconfig = fTrackingManager->GetG4HepEmConfig();
@@ -33,9 +34,11 @@ void G4EmStandardPhysics_AdePT::ConstructProcess()
       fAdePTConfiguration->GetMultipleStepsInMSCWithTransportation());
   g4hepemconfig->SetEnergyLossFluctuation(fAdePTConfiguration->GetEnergyLossFluctuation());
 
+  // Loop over all configured Woodcock regions and register them
   for (const auto &regionName : fAdePTConfiguration->GetWDTRegionNames()) {
     g4hepemconfig->SetWoodcockTrackingRegion(regionName);
   }
+  // set Woodcock tracking energy limit
   g4hepemconfig->SetWDTEnergyLimit(fAdePTConfiguration->GetWDTKineticEnergyLimit());
 
   G4Electron::Definition()->SetTrackingManager(fTrackingManager);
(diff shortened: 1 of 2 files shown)


src/G4EmStandardPhysics_AdePT.cc
```diff
@@ -1,56 +1,41 @@
...
+  // Keep Geant4-owned cleanup behavior for tracking manager deletion.
 }
 
-void AdePTPhysics::ConstructProcess()
+void G4EmStandardPhysics_AdePT::ConstructProcess()
 {
-  // Register custom tracking manager for e-/e+ and gammas.
+  // First register the standard Geant4 EM processes for this constructor.
+  G4EmStandardPhysics::ConstructProcess();
+
```
SeverinDiederichs:
```suggestion
  // First register the standard Geant4 EM processes for this constructor.
  G4EmStandardPhysics::ConstructProcess();

  // Register custom tracking manager for e-/e+ and gammas.
```

src/G4EmStandardPhysics_AdePT.cc
```diff
@@ -1,56 +1,41 @@
...
   g4hepemconfig->SetMultipleStepsInMSCWithTransportation(
       fAdePTConfiguration->GetMultipleStepsInMSCWithTransportation());
   g4hepemconfig->SetEnergyLossFluctuation(fAdePTConfiguration->GetEnergyLossFluctuation());
 
-  // Loop over all configured Woodcock regions and register them
   for (const auto &regionName : fAdePTConfiguration->GetWDTRegionNames()) {
     g4hepemconfig->SetWoodcockTrackingRegion(regionName);
   }
-  // set Woodcock tracking energy limit
   g4hepemconfig->SetWDTEnergyLimit(fAdePTConfiguration->GetWDTKineticEnergyLimit());
```
SeverinDiederichs:
```suggestion
  // set Woodcock tracking energy limit
  g4hepemconfig->SetWDTEnergyLimit(fAdePTConfiguration->GetWDTKineticEnergyLimit());
```

PR #506, opened 2026-03-12
fix doxygen warnings

SeverinDiederichs:
This PR fixes the doxygen warnings that appeared in building the documentation.

It was verified that this PR
- [ ] Changes physics results


JuanGonzalezCaminero (approved):
Looks good to me

PR #507, opened 2026-03-12
dont create hostTrackData if advanced MC truth is not enabled

SeverinDiederichs:
This PR fixes an issue with the HostTrackData mapper:

When running without advanced MCTruth info, the hostTrackData was still generated for particles that were being offloaded to the GPU. That was only deleted at the end of the event, causing potential crashes in ttbar events in CMS example1.

The fix is to not create the hostTrackData mapper if no advanced MCTruth info is used, as it won't be deleted when the track finishes.

Note: currently, the advanced MC truth info + hostTrackData is entangled with `/adept/CallUserTrackingAction` or `/adept/CallUserSteppingAction`. This will be separated for clarity in a separate PR.

It was verified that this PR
- [ ] Changes physics results


JuanGonzalezCaminero (approved):
Thanks for the fix!

PR #512, opened 2026-03-14
Enhance drift test to test the full MC truth chain

SeverinDiederichs:
This PR drastically upgrades the current drift tests. Instead of just recording the energy deposition, it records all available MC truth info that AdePT provides, including creator process, step-defined process, and a custom G4VUserTrackInfo, which is implemented to track the ancestor primary ID and the generation.

The ROOT part is enabled if ROOT is discoverable when the integrationBenchmark is compiled.
In the CI drift test, the ROOT backend is used as long as it is available in both the master and the PRs executable (i.e., for this PR, the previous simple drift test was used, from the next on, it will be the new ROOT-based drift test).

A few difficulties were encountered:
- to ensure reproducibility when merging the results from multiple events, the energy deposition in double is stored in the 64 bits as uint64, such that the addition is deterministic
- due to the fact that the `ProcessGPUStep` is used to link between the parent and the secondaries the MC truth but the nuclear processes are only invoked in `ReturnTrack`, the linkage is broken for the secondaries generated by the nuclear processes. Therefore, testing of the lineage is deactivated for now, as it is not reproducible (as the G4 id is not reproducible, only the GPU trackID is).

With this PR, good insights from the shower profile can be extracted:

generation_population
vertex_x
step_defining_process_counts
creator_process_counts

For the comparison of the ROOT histograms, the `check_root_histograms.py` can be used. For plotting, a simple script to plot all histograms was added as `plot_truth_histograms.py`

It was verified that this PR
- [ ] Changes physics results

### DIFF
--- a/test/regression/IntegrationTest/include/TruthHistogrammer.hh
+++ b/test/regression/IntegrationTest/include/TruthHistogrammer.hh
@@ -15,7 +15,7 @@ class G4Track;
  * @brief Collects aggregated truth observables for the ROOT-based drift test.
  *
  * The collector stores categorical histograms as label -> count maps and
- * continuous observables as exact floating-point bit patterns -> count maps.
+ * continuous observables as exact floating-point values -> count maps.
  * The master thread later serializes these maps into ROOT histograms in a
  * deterministic order for exact comparison against the reference file.
  *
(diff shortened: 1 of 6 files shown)


test/regression/IntegrationTest/include/TruthHistogrammer.hh
```diff
@@ -0,0 +1,80 @@
...
+  /// Histogram with stable string labels and integer counts.
+  using CountHistogram = std::map<std::string, std::uint64_t>;
+  /**
+   * @brief Histogram with exact `double` bit patterns and integer counts.
+   *
+   * This is the key trick that makes the ROOT truth path reproducible in MT:
+   * floating-point values are first bucketed by their exact IEEE-754 bit
+   * pattern, and only the integer populations are merged across workers.
+   */
+  using ValueHistogram = std::map<std::uint64_t, std::uint64_t>;
```
agheata: Why not just summing double and accepting the rounding error at the end? Well, yes, if you want this to be pure reproducible, but the sum is then meaningless as a number, in general. I mean, this trick is a weird way to deal with the non-associativity

test/regression/IntegrationTest/src/TruthHistogrammer.cc
```diff
@@ -0,0 +1,464 @@
...
+    hist.SetBinContent(1, 0.0);
+    WriteLabelMetadata(file, "edep_by_volume", {});
+  } else {
+    std::vector<std::string> labels;
+    labels.reserve(energyDepositByVolume.size());
+    int bin = 1;
+    for (const auto &[volumeId, entry] : energyDepositByVolume) {
+      std::ostringstream label;
+      label << volumeId << ":" << entry.label;
+      hist.SetBinContent(bin, CanonicalWeightedSum(entry.contributions));
```
agheata: The values in this histogram are meaningless as energy deposits, I guess

PR #515, opened 2026-03-17
reduce drift test statistics

SeverinDiederichs:
The new drift test with MC truth introduced in #512 was taking only a bit longer on my local machine but much longer on the CI machine. Therefore, this PR is reducing the statistics such that the drift test is fast again (the new MC truth test is heavier, but also had much higher statistics). This PR puts back the old level of statistics.

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
LGTM

PR #528, opened 2026-03-27
Make Geant4 a strict dependency

SeverinDiederichs:
Geant4 was not required in some standalone examples in AdePT, which don't exist anymore. 

Therefore, Geant4 is required now, and now this is enforced correctly in CMake

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
Thanks for the fix!

Issue #516, opened 2026-03-18
Refactor AdePT into transport library and G4Integration library


SeverinDiederichs:
The goal is to refactor AdePT such that that the transport and the Geant4 integration are cleanly separated.
Today, AdePT is build as a single library, with the CUDA implementation living in headers, leading to very few translation units that suck in a huge amount of code. This original design was done to avoid cuda linking issues. Since the introduction of cuda_rdc into AdePT, this is not needed anymore and two clean and separated libraries can be created.

The core problems are that
`AdePTTrackingManager`, `AdePTGeant4Integration`, and `AsyncAdePTTransport` mix many responsibilities:
- The `AsyncAdePTTransport` depends on the `AdePTGeant4Integration` leading to the G4 dependency of the transport.
- initialization and upload to the GPU are interleaved
- the GPU step processing is not as clearly separated as it should

This makes the code hard to read and hard to test in isolation, as nothing can be initialized on device without the full G4 initialization. This leads to code duplication when trying to test certain parts, see the `testField`.

The long-term target is to have two libraries:
- `AdePT_Transport`
  - owns GPU steering and kernels, device upload, retrieving of scoring and hits
  - depends on CUDA, VecGeom, and G4HepEm
  - must not depend on Geant4 tracking/reconstruction APIs
- `AdePT_G4_Integration` (although `AdePT_G4Integration` would probably be a better name)
  - owns `AdePTTrackingManager`, `AdePTGeant4Integration`, geometry preparation, Geant4 reconstruction, and treatment of returned tracks
  - depends on Geant4 and `AdePT_Transport`

Therefore, the refactor is planned as follows: 

- Move per-thread `AdePTGeant4Integration` ownership to `AdePTTrackingManager`
- Move Geant4 logic out of `AsyncAdePTTransport`
- Split geometry into:
  - metadata build on the Geant4 side
  - plain shared data objects
  - device upload on the core side
- Replace large responsibilities (e.g., `AdePTGeant4Integration `or `PerEventScoringImpl.cuh` with smaller, targeted pieces.

The overall goal is to be able to test many parts of the code in isolation.

A series of PRs will be created and linked to the issue.

PR #530, opened 2026-03-28
Refactor AdePT into 2 libraries 11:  move packing of Woodcock data to GeometryBridge

SeverinDiederichs:
This PR belongs to the refactor of AdePT described in https://github.com/apt-sim/AdePT/issues/516.

This small PR fixes a leftover that is needed to split AdePT into two libraries:

The initialization of the Woodcock tracking geometry data was done in the `CommonStruct`, which will belong to the `AdePT::core`, but it was relying on the `G4RegionStore`. This PR moves it into the GeometryBridge, which will live in the g4integration.

It was verified that this PR
- [ ] Changes physics results

### DIFF
--- a/include/AdePT/core/CommonStruct.h
+++ b/include/AdePT/core/CommonStruct.h
@@ -12,9 +12,6 @@
 
 #include <G4HepEmRandomEngine.hh>
 
-#include "G4RegionStore.hh"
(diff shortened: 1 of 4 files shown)


agheata (approved):
LGTM

PR #537, opened 2026-04-07
Inherit static/shared build from G4

SeverinDiederichs:
This PR allows to build AdePT both as a static or as a shared library. Similarly to what G4HepEm does, the build type is inherited from the provided Geant4 installation. 
However, there is a difference: in G4HepEm, if both a static and a shared G4 are provided, G4HepEm is also build both as a static and as a shared library. To avoid the complications with the static helper library generated for the cuda_rdc, AdePT takes a preference to the static build:
- if only shared G4 is provided, AdePT is build as a shared library
- if only static G4 is provided, AdePT is build as a static library
- if both are provided, AdePT is build as a static library

This will simplify some of the integration logic in Athena and CMSSW, as both require a static build of AdePT, which was currently done via a patch.

It was verified that this PR
- [ ] Changes physics results

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -294,12 +294,17 @@ endif()
 find_package(Geant4 REQUIRED)
 message(STATUS "Using Geant4 version ${Cyan}${Geant4_VERSION}${ColorReset} from ${Geant4_INCLUDE_DIRS}")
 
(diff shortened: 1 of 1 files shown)


hahnjo (approved):
LGTM!

PR #540, opened 2026-04-09
Remove leaking of tracks and replace it by construction from steps

SeverinDiederichs:
This PR is a follow up to #539 and should not be reviewed before that one is merged.

This PR extends the idea to use the GPUHits to create the tracks for nuclear reactions to use it for all leaked tracks.
This has the following consequences:

- all leaked tracks related machinery is removed, drastically simplifying the state machine, the asynchronous GPU transport loop, the interfaces
- the deferred nuclear steps and leaked tracks can now be treated as the same again and the sorting is simplified

Regarding performance:
- sending back leaked tracks is cheaper than steps. Therefore, this new implementation can be a) cheaper if a step + a track were sent back, b) more expensive before only a track was sent back, as the step is bigger. 
- simplifying the state machine and removing synchronizations should not really make a difference, as the leaks were usually not the problem.

### Performance results in Athena:
Run on the RTX4090

Note that there are quite the fluctuations between runs, larger than the std can show, as some results deviated quite a bit when run again on another day.
| Config | Mean Real [s] | Std [s] | Mean wall [min] |
|---|---:|---:|---:|
| `Mono_Master` | 860.3 | 11.6 | 14.338 |
| `Mono_PR` | 850.4 | 9.13 | 14.173 |
| `Split_Master` | 855.1 | 4.9 | 14.252 |
| `Split_PR` | 863.2 | 6.4 |14.386 |
| `G4HepEm` | 1149.9 | 2.2 | 19.165 |

### Changes in physics:
This PR does change the physics behavior: first, before #539 is merged, it contains the changes regarding the MC truth that are explained in that PR, which fixes previously incorrect behavior for gamma / lepton nuclear.

Then, this PR intentionally changes the local time and the proper time histograms. Now the `GPUHit` transfers both as `float` to the CPU, where they are changed to the G4double (note that the GPU track itself also stores them in float). This is different w.r.t. how the leaked tracks were handling both (they were returned in double). Furthermore, the proper time was not set in the G4Track when being returned. This is now also fixed. 

Note that commit `d2e170c34c53e9e71ca0d1106920c6cf4b61501e` intentionally recreates the exact behavior as before, so it passes the drift test cleanly, before in the final commit, the drift test behavior changes, but is correct.

As this PR superseded #539, it resolves #533

### DIFF
--- a/include/AdePT/core/AdePTConfiguration.hh
+++ b/include/AdePT/core/AdePTConfiguration.hh
@@ -25,7 +25,6 @@ public:
   void AddWDTRegionName(std::string name) { fWDTRegionNames.push_back(name); }
   void SetVerbosity(int verbosity) { fVerbosity = verbosity; };
   void SetMillionsOfTrackSlots(double millionSlots) { fMillionsOfTrackSlots = millionSlots; }
-  void SetMillionsOfLeakSlots(double millionSlots) { fMillionsOfLeakSlots = millionSlots; }
   void SetMillionsOfHitSlots(double millionSlots) { fMillionsOfHitSlots = millionSlots; }
   void SetHitBufferFlushThreshold(float threshold) { fHitBufferFlushThreshold = threshold; }
   void SetCPUCapacityFactor(float CPUCapacityFactor) { fCPUCapacityFactor = CPUCapacityFactor; }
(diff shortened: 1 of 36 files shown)


agheata (approved):
Thanks @SeverinDiederichs for this nice cleanup/improvement. LGTM!

SeverinDiederichs:
@codex review

PR #538, opened 2026-04-07
List all failing ROOT histograms in drift test check

SeverinDiederichs:
This PR improves the output of the ROOT histogram checker:

Before, it would stop and fail at the first mismatch, e.g.:

```
Histogram 'creator_process_counts' differs: bin count 6 != 11
```

Now, it will always compare all histograms and lists the differences:

```
ROOT histogram comparison found mismatches: 35 failing checks across 42 common histograms.
- Histogram 'creator_process_counts' differs: bin count 6 != 11
- Histogram 'edep_by_volume' differs at bin 2 ('1:G4_lA'): file1=40.780903516103535 file2=40.78090351610354
- Histogram 'final_dir_x' differs: bin count 87046 != 87435
- Histogram 'final_dir_y' differs: bin count 87046 != 87435
- Histogram 'final_dir_z' differs: bin count 87046 != 87435
- Histogram 'final_ekin' differs: bin count 2516 != 2573
- Histogram 'final_global_time' differs: bin count 87046 != 87435
- Histogram 'final_local_time' differs: bin count 87042 != 87431
- Histogram 'final_num_secondaries' differs: bin count 3 != 12
- Histogram 'final_proper_time' differs: bin count 1 != 110
- Histogram 'final_step_length' differs: bin count 87042 != 87431
- Histogram 'final_total_edep' differs: bin count 74387 != 74688
- Histogram 'final_volume_counts' differs at bin 1 ('0:G4_Pb'): file1=49.0 file2=53.0
- Histogram 'final_x' differs: bin count 86614 != 86980
- Histogram 'final_y' differs: bin count 86949 != 87323
- Histogram 'final_z' differs: bin count 86951 != 87331
- Histogram 'generation_population' differs at bin 5 ('bin=5'): file1=4302.0 file2=4357.0
``` 

This is important to understand differences in the drift test, as one can see if only one histogram (e.g., the time at vertex) differs or the full physics output changed (e.g., also the energy deposition and all kinds of other histograms).

It was verified that this PR
- [ ] Changes physics results
- [x] Does not change physics results

PR #542, opened 2026-04-10
integrationBenchmark: use GetSecondaryInCurrentStep in MC truth linking

SeverinDiederichs:
This PR fixes to use the `GetSecondaryInCurrentStep` instead of the `GetSecondary`. In principle, the `GetSecondary` gives all secondaries of the full track, not just of the current step. However, AdePT never provides those, it only provides the secondaries in the current step.

Thus, although equivalent here, the correct function should be used to avoid any confusion.

It was verified that this PR
- [ ] Changes physics results
- [x] Does not change physics results. Note:
This does change the histogram of the `final_num_secondaries` for the tracks that are tracked by G4, which are the hadrons generated by nuclear reactions. This was confirmed by running this PR on top of #538, to confirm that only this expected histogram fails.


agheata (approved):
Looks good and matches the intent better, LGTM!

PR #545, opened 2026-04-13
Enable Gamma Russian Roulette as ATLAS SteppingAction

### LINKED ISSUE #532: Add Photon Russian Roulette for ATLAS
ATLAS currently uses Russian Roulette for photons (PRR) to accelerate the simulation. In AdePT, we currently simulate all photons. As photons are the bottleneck, it would be good to reduce their amount by also supporting Russian Roulette for them.

This requires
- using the `AuxVolData` to identify the volumes where PRR is applied
- a compile time hook similar to the SteppingActions for ATLAS to apply PRR in the kernels.

Note: currently the originTouchable is used to identify the correct volumes for PRR on the host. As PRR is not applied for the GPU, the originTouchable does not need to be provided. This be removed in AdePT, as it is quite expensive to generate the originTouchable.

SeverinDiederichs:
This PR adds the gamma Russian Roulette for ATLAS and resolves #532.

The Russian Roulette is implemented as a SteppingAction: if active (`-DADEPT_STEPPINGACTION=ATLAS`), before a gamma is created, it will undergo RR. The settings for the energy threshold and probability are hardcoded for now, as the default parameters in Athena. One important difference: Athena checks for the placed volume prefix "LAr" while AdePT does this at the logical volume level. A safety check is added, and it does not trigger in Athena, in case a placed volume would be created without the LAr prefix, while the logical volume had one.

### Notes:
- example1 was not using the weight in the energy deposition (integrationBenchmark was), this is fixed
- A new constructor for the track is needed, to assign the weight not directly from the parent pass it as a parameter

### Performance results:
Run on the RTX4090

| Config | Mean Real [s] | Std [s] | Mean wall [min] |
|---|---:|---:|---:|
| `Mono_Master` | 891.0 | 11.5 | 14.850 |
| `Mono_NoRR` | 909.9 | 3.0 | 15.165 |
| `Mono_GammaRR` | 860.3 | 11.6 | 14.338 |
| `Split_Master` | 898.0 | 8.6 | 14.967 |
| `Split_NoRR` | 904.6 | 2.0 | 15.077 |
| `Split_GammaRR` | 855.1 | 4.9 | 14.252 |
| `G4HepEm` | 1149.9 | 2.2 | 19.165 |

Validation: 
The results in Athena with both monolithic and split kernels look good with the Russian Roulette turned on.
Screenshot from 2026-04-20 08-54-38

### To do:
- [x] update Readme and documentation to describe the new option
- [x] do performance tests with Athena
- [x] do physics validation in Athena

For an upcoming PR: 
- Add CI test. The reason to not do it in this PR, is that it requires changing the drift test to use a different compilation flag. Doing in a separate PR will allow to show that the current test passes cleanly and then a separate PR will be made to change the CI test to include a case with Russian Roulette.


agheata (approved):
Thanks for this important addition. As discussed offline, the new photon roulette only preserves expectation for unit-weight parents, but this is how it is done in ATLAS so it has to match, propose a fix in Athena and only after update in AdePT. LGTM

PR #547, opened 2026-04-14
Trigger jenkins via GitHub actions pipeline

SeverinDiederichs:
This PR fixes the issue that the jenkins pipeline was not triggered by PRs from external contributors even if the workflow was approved.

Now, the Jenkins pipline is triggered as soon as the github workflow finishes instead of a webhook.

For this, the three previous small checks (docs, clangformat, reuse compliance) are merged into one.
As the `reuse.yml` was the last file to use the GPL-3.0 license, the file is removed.

It was verified that this PR
- [ ] Changes physics results
- [x] Does not change physics results

PR #549, opened 2026-04-14
switch to use Jenkins user for trigger

SeverinDiederichs:
This is a follow up to #547.

Using the tokens as in #547 triggers the CERN SSO, so it does not run the job. 
This is (hopefully) fixed by providing the jenkins user and jenkins API token instead (both stored in github as a secret)

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
LGTM

PR #566, opened 2026-04-15
Remove legacy testing

SeverinDiederichs:
This removes the legacy testing suite, which is not used and is also not functional anymore.
It was used for the `integrationBenchmark` application, when it was supporting `--do-validation` and `---do-benchmark`. The benchmark has been removed and the validation has been changed, so these scripts are not functional and can be removed.

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
Thanks for the massive cleanup!

PR #568, opened 2026-04-15
add precommit as an easier user entry point for clangformat

SeverinDiederichs:
This PR adds pre-commit to the repository. It allows for an easier user experience to be aware of clang formatting already at the stage of committing and not at the stage of CI when the PR is open

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
Looks good, LGTM!

PR #569, opened 2026-04-16
[mini] fix TruthHistogrammer warning

SeverinDiederichs:
When the integration benchmark was build without ROOT support, these warnings appeared

```
/home/ndiederi/software/AdePT/test/regression/IntegrationTest/src/TruthHistogrammer.cc:139:6: warning: ‘bool {anonymous}::ValueKeyLess(double, double)’ defined but not used [-Wunused-function]
  139 | bool ValueKeyLess(double lhs, double rhs)
      |      ^~~~~~~~~~~~
/home/ndiederi/software/AdePT/test/regression/IntegrationTest/src/TruthHistogrammer.cc:129:13: warning: ‘std::string {anonymous}::ExactValueLabel(double)’ defined but not used [-Wunused-function]
  129 | std::string ExactValueLabel(double value)
      |             ^~~~~~~~~~~~~~~
```

because the guarding of the ROOT support was not scoping all it should have.

It was verified that this PR
- [ ] Changes physics results

### DIFF
--- a/test/regression/IntegrationTest/src/TruthHistogrammer.cc
+++ b/test/regression/IntegrationTest/src/TruthHistogrammer.cc
@@ -126,6 +126,7 @@ std::string ParticleName(const G4Track *track)
   return track && track->GetParticleDefinition() ? track->GetParticleDefinition()->GetParticleName() : "none";
 }
 
+#ifdef ADEPT_INTEGRATIONTEST_HAS_ROOT
 std::string ExactValueLabel(double value)
 {
   if (std::isnan(value)) return "nan";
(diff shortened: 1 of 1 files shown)


agheata:
@SeverinDiederichs I see that tests may self-cancel based on some priority ?

SeverinDiederichs:
> @SeverinDiederichs I see that tests may self-cancel based on some priority ?

I just saw as well, I am trying to stress the system (the PR would not require any tests). This time, I started the `run-test` before the other prechecks were finished, and they seem to cancel the others, so that is something I need to investigate.

SeverinDiederichs:
@agheata ok, it was a real bug, fixed in #570.

SeverinDiederichs:
A random comment to see if it still cancels selfhosted CI

agheata (approved):
LGTM

PR #571, opened 2026-04-16
Move PreuserTrackingAction into UseUserAction guard

SeverinDiederichs:
Previously, the PreUserTrackingAction was always called and potential custom UserTrackInfo was attached. However, if the CallUserTrackingAction in AdePT was off, it would have been attached to a dummy HostData, leading to a memory leak.

This is fixed by simply moving the block into the part that is guarded by CallUserTrackingAction

Reported by @dkonst13

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
Looks good to me. The moved block now matches the callUserActions lifetime/ownership model.
This also means PreUserTrackingAction side effects no longer happen when user-action handling is disabled, but that seems consistent with the flag semantics.

PR #572, opened 2026-04-17
update PR template

SeverinDiederichs:
This small PR updates the PR template.
One does not have to set the checks whether the PR changed the physics by hand anymore, as the CI test will clearly show whether the physics drifted and the validation was run or not.

It was verified that this PR
- [ ] Changes physics results


agheata (approved):
Indeed, that was clearly a temporary solution, CI should not function based on trust ;-)
LGTM

PR #573, opened 2026-04-17
Add selfhosted nightly

SeverinDiederichs:
This PR adds a nightly workflow on the self-hosted machine. It does the same as Jenkins is currently doing:

Running the unit tests on the monolithic build, and the validation on the monolithic and the split build. The drift tests are not run, as they only make sense in comparing two branches.

A badge is added to the readme to show the nightly status.
If a nightly fails, an issue is automatically created. Further failures will be added to the same issue. The issue is closed automatically when the nightlies pass.

When this PR is merged, AdePT does not require any use of Jenkins anymore.


agheata (approved):
Nice workflow, let's try it!

PR #579, opened 2026-04-20
check VecGeom has CUDA enabled

### LINKED ISSUE #184: Check if VecGeom was built with CUDA support
I just tried setting up AdePT and it nicely told me that I needed to have an installation of VecGeom with GDML enabled. I fixed that and then I got these errors:
```
CMake Error at test/CMakeLists.txt:8 (add_executable):
  Target "test_sparsevector_cpu" links to target
  "VecGeom::vecgeomcuda_static" but the target was not found.  Perhaps a
  find_package() call is missing for an IMPORTED target, or an ALIAS target
  is missing?
Call Stack (most recent call first):
```
It seems I also need to enable CUDA when building/installing VecGeom. It would be nice if there was a similar error/warning as with GDML support :)

SeverinDiederichs:
This PR resolves #184. It had been fixed in some examples but was not enforced in the library version of AdePT that we have now.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -198,6 +198,14 @@ find_package(XercesC REQUIRED)
 # Find VecGeom geometry headers library
 find_package(VecGeom REQUIRED)
 
+if(ADEPT_USE_CUDA AND NOT VecGeom_CUDA_FOUND)
+  message(FATAL_ERROR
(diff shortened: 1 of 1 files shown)


agheata (approved):
Thanks for cleaning this up!

Issue #541, opened 2026-04-09
Enable ApplyCuts in nuclear reactions


SeverinDiederichs:
G4HepEm allows for a flag to use ApplyCuts also in the nuclear reactions in the `PerformNuclear` call. AdePT previously couldn't use that, as the nuclear reactions were performed via leaked tracks, so there was no G4Step anymore, or no SD code that could record the energy deposition.

Following #539, this will be done correctly. Now, `ApplyCuts` could be used on the `PerformNuclear` call, making the code more efficient, as the full secondaries could actually be cut and the energy deposition passed to the SD code.

SeverinDiederichs:
This issue was resolved by #589

PR #589, opened 2026-04-22
enable apply cuts for PerformNuclear

SeverinDiederichs:
Before the major cleanup in #540, the `PerformNuclear` was called when the leaked tracks where returned, not when the SD code was invoked. Therefore, ApplyCuts could not be used. Now, as the `PerformNuclear` is called directly before the SD code is invoked, we can apply the cuts also to `PerformNuclear` like G4HepEm and cut low energy secondaries.

This solves #541 

### Physics behavior changes:

This PR naturally changes all drift tests, as apply cuts is used in the CI, so edep and the full history is changed.

### DIFF
--- a/src/AdePTGeant4Integration.cpp
+++ b/src/AdePTGeant4Integration.cpp
@@ -16,6 +16,8 @@
 #include <G4StepStatus.hh>
 
 #include <G4HepEmNoProcess.hh>
+#include <G4HepEmConfig.hh>
+#include <G4HepEmParameters.hh>
 
 #include "G4Electron.hh"
 #include "G4Positron.hh"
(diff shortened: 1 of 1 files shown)


SeverinDiederichs:
This PR fails the validation ever so slightly: 
```
Layer 48: Relative Error = 1.009609%, File1 = 14.551780, File2 = 14.406333
```

In fact, this is caused by pure bad luck.
When running with 10 different seeds, this PR fails the CI 2 times, just because the bar of 1% is relatively strict given the random fluctuations.
```
1234567: pass, 0.411247%
1234568: fail, 1.076262%
1234569: pass, 0.570841%
1234570: pass, 0.560200%
1234571: pass, 0.496087%
1234572: pass, 0.500329%
1234573: pass, 0.940292%
1234574: pass, 0.650887%
1234575: pass, 0.408893%
1234576: fail, 1.369465%
```

However, on master it is very similar:
```
1234567: pass, 0.407735%
1234568: fail, 1.038702%
1234569: pass, 0.561926%
1234570: pass, 0.556683%
1234571: pass, 0.509937%
1234572: pass, 0.500730%
1234573: pass, 0.944030%
1234574: pass, 0.642265%
1234575: pass, 0.410258%
1234576: fail, 1.347790%
```

Therefore, this PR is not causing a physics regression. The CI must be updated to run with higher statistics.

agheata (approved):
Thanks for the fix, LGTM

PR #596, opened 2026-04-28
Separate GPU track into neutral and charged track

SeverinDiederichs:
This PR reduces the memory footprint of the tracks by splitting the track into a neutral track and a charged track (similarly to what G4HepEm does).

As the track size depends also on the NavigationState size, this is the table for before and after:

| Nav mode | NavigationState size | Mode | Master Track | PR NeutralTrack | PR ChargedTrack |
|---|---:|---|---:|---:|---:|
| NavIndex | 12 B | monolithic | 248 B | 200 B | 240 B |
| NavIndex | 12 B | split | 312 B | 288 B | 312 B |
| NavTuple depth 3 | 36 B | monolithic | 272 B | 224 B | 264 B |
| NavTuple depth 3 | 36 B | split | 360 B | 336 B | 360 B |
| NavTuple depth 4 | 44 B | monolithic | 280 B | 232 B | 272 B |
| NavTuple depth 4 | 44 B | split | 376 B | 352 B | 376 B |

While ncu shows some lower memory traffic, it did not translate into a measurable speedup (tested with ttbar events in example1 with the LHCb geometry).

Note: Now there is no array over the tracks possible anymore, as they differ, so the tracksAndSlots struct was changed as well as some loops were replaced by lambda function calls.

### DIFF
--- a/include/AdePT/core/AsyncAdePTTransport.cuh
+++ b/include/AdePT/core/AsyncAdePTTransport.cuh
@@ -170,6 +170,23 @@ __device__ inline uint64_t GenerateSeedFromTrackInfo(const AsyncAdePT::TrackData
   return fnv1a_hash64(input, 12);
 }
 
+template <typename SpeciesManagerT>
+__device__ inline void InitTrackToQueue(SpeciesManagerT &speciesTM, const AsyncAdePT::TrackDataWithIDs &trackInfo,
+                                        short queueIndex, adept::MParrayT<QueueIndexPair> *toBeEnqueued,
+                                        uint64_t initialSeed)
+{
+  // TODO: Delay when not enough slots?
+  const auto slot = speciesTM.NextSlot();
+  // Scramble the initial seed with track data so a particle returning from the
+  // device and being injected again does not collide with its old RNG stream.
+  auto seed = GenerateSeedFromTrackInfo(trackInfo, initialSeed);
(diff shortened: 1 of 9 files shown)


hahnjo:
> these members [...] are still in the neutral track, to avoid an increase of size due to trailing padding

I didn't understand this at first, but it's because `ChargedTrack` inherits from `NeutralTrack`. An alternative may be to have a common `Track` as base class for both that is as tightly packed as possible. For example, `NeutralTrack::numIALeft` is not used / hidden by `ChargedTrack::numIALeft`. For split kernels, this member is omitted, but then there is padding of the same size before `vecgeom::Vector3D<double> pos`...

SeverinDiederichs:
> I didn't understand this at first, but it's because `ChargedTrack` inherits from `NeutralTrack`. An alternative may be to have a common `Track` as base class for both that is as tightly packed as possible. For example, `NeutralTrack::numIALeft` is not used / hidden by `ChargedTrack::numIALeft`. For split kernels, this member is omitted, but then there is padding of the same size before `vecgeom::Vector3D<double> pos`...

Indeed, you are correct. I named it `TrackBase`  so it is clear that it is not a full usable track itself (as it was before), and packing it more tightly, more memory could be saved and it is cleaner.

agheata (approved):
Good, now the track info seems to be optimally packed, and I don't see any regressions from different perspectives. Just a minor note on a stale comment in `include/AdePT/core/Track.cuh:53` that refers to `NeutralTrack` while it should refer to `TrackBase`

PR #597, opened 2026-04-29
add CMS stepping action for dead regions

SeverinDiederichs:
This PR adds the CMS-specific stepping action which kills tracks in so-called dead regions.
Similarly to the ATLAS Russian Roulette, the VolAuxData gets an additional flag for the volumes in that region and based on that flag tracks are killed.

This changes the interface to pass the full volAuxData to the SteppingAction instead of just the `volAuxData.fMCIndex`.

Note: additionally, a small bug was fixed:
In the monolithic kernels, the volAuxData of the initial volume was used. However, in case of the relocation, this is not correct, so the volAuxData must be updated.

The dead regions can be configured in the AdePTConfiguration (currently, for CMS only), via
```c++
fAdePTConfiguration->AddDeadRegionName("QuadRegion");
fAdePTConfiguration->AddDeadRegionName("InterimRegion");
```


agheata (approved):
Thanks for adding the dead region mechanism in this way. Even if now it applies only to CMS, we can generalize it in future, as discussed offline. LGTM!

Issue #578, opened 2026-04-20
Improve Russian Roulette


SeverinDiederichs:
Two issues were left open by #545:

- [x] Currently, it does exactly what Athena does. However, there is a potential issue in case the neutron Russian Roulette and gamma RR have different weights. In that case, the RR could be applied twice (and then, incorrectly) to particles.
The behaviour in Athena will be fixed with [this MR](https://gitlab.cern.ch/atlas/athena/-/merge_requests/87611).
Solved in AdePT by #598
- [x] A CI test should be put in place to test the RR. This can be done by adding a testEm3 like geometry, which has the same naming scheme as ATLAS uses to check for RR, with a "LAr" prefix.
Solved in AdePT by #599

PR #598, opened 2026-04-29
use simple unity check in ATLAS Russian Roulette

SeverinDiederichs:
This PR improves the Russian Roulette as discussed in #578:

Instead of doing a check against the weight (which only works in athena as the weight for neutrons and gammas is the same, [see this MR](https://gitlab.cern.ch/atlas/athena/-/merge_requests/87611)), we do a simple check against unity here.

Note that the float comparison is fine here, as the weight is assigned to 1.0f and not accumulated.

This PR does not change any behavior, it simply fortifies AdePT against any possible future changes of the weights in Athena.

Note, this PR should be reviewed after #599 is merged, so the new RR CI can be triggered on this PR.


agheata (approved):
Nice, good that they are following our suggestions! LGTM

PR #599, opened 2026-04-29
add full CI test and drift test for Russian Roulette

SeverinDiederichs:
This PR partially addresses #578.

A validation test and a drift test are added for the Russian Roulette.

To avoid additional builds (which take most of the CI time), all CI tests are now build with the ATLAS stepping action flag (which enables RR for volumes which begin with the name `LAr`). Then, an additional testEm3 gdml file is added, where the volumes with LAr are renamed such that they trigger the naming scheme in athena.

Then, a drift test and a validation test are added


agheata (approved):
This looks good overall, thanks! LGTM

PR #600, opened 2026-04-29
Split AdePTTransportConfig from Geant4UI configuration

SeverinDiederichs:
This PR is a leftover from #516:

It splits an `AdePTTransportConfig` used by `AsyncAdePTTransport` from the user-facing `AdePTConfiguration`.

Before,  the `AdePTConfiguration` was passed to the AdePTTransport, which pulls in the Geant4 UI messenger. Therefore, the AdePT::Transport library still used implicitly some Geant4 part, which it should not.

This if fixed by creating a small clean struct `AdePTTransportConfig`, which contains exactly the data needed by the transport which is captured from the user via the general `AdePTConfiguration`. 

Further changes:
- quite a few getters were constified.
- Some region and UserAction flags were moved from the Transport to the G4 integration, where they belong
- Moved `AdePTConfiguration` constructor/destructor out of the header, so the messenger is still created automatically but no longer included directly from the AdePTConfiguration header.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -489,6 +489,7 @@ set(ADEPT_CORE_SRCS
 )
 
 set(ADEPT_G4_INTEGRATION_SRCS
+  src/AdePTConfiguration.cc
   src/AdePTGeometryBridge.cpp
   src/G4HepEmTrackingManagerSpecialized.cc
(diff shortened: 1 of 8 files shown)


agheata (approved):
Thanks for the separation!

PR #601, opened 2026-04-30
[mini] Allow exact-fit host buffer transfers

SeverinDiederichs:
Previously, the host-device transfer was rejected in case it was the exact free space, which should be allowed.

This mini PR fixes that


agheata (approved):
Yes, changing the condition makes TransferHitsToHost consistent with the buffer manager contract. Thanks for the quick fix!

PR #602, opened 2026-04-30
Fix CPU-finish event thread id & factor out GetThreadId

SeverinDiederichs:
The check to finish an event on CPU was incorrect:
It was comparing > 0, (the default is -1 if the event is not finished on CPU), so event 0 could have not been finished on CPU.

Furthermore, the access to the vector via `G4Threading::G4GetThreadId();` was unsafe, as in sequential mode it can return a negative number. This was fixed  in #546 in most places of the code. Here, the AdePT helper `GetThreadId` is factored out and also used in the CPU event finish.


agheata (approved):
Thanks for the fixes!

PR #603, opened 2026-04-30
Cleaning TrackData

SeverinDiederichs:
This PR cleans some long standing leftovers:

Previously, there was a split between `TrackData` and `TrackDataWithIDs` where the former was used by Sync AdePT and the latter by Async AdePT, as it added the required thread ID and event ID to the `TrackData`. As now only async AdePT is used, the event and thread ID can be directly put into `TrackData` and `TrackDataWithIDs` can be deleted.

Furthermore, there was a sorting operator for the tracks being returned to the CPU. As now only steps are returned to the CPU (and sorted for reproducibility) but not `TrackData` objects, the operator can be deleted.


agheata (approved):
Good to go, thanks for the cleanup!

PR #604, opened 2026-04-30
Rename returned GPU hits to GPU steps

SeverinDiederichs:
This PR cleans up the internal naming around the GPU steps returned.

Previously, they were called GPUHits, as there was scoring done on the GPU, or only the SD code was invoked.

Now, as UserActions etc are invoked, and steps can be returned even if no SD is attached, we are really returning GPU steps.
The GPU records step information, transfers it back to the host, and the Geant4 integration then reconstructs a `G4Step` and calls the sensitive detector on the CPU. So the internal names should say `step`, not `hit` or `scoring`.

The renames are:

- ScoringCommons.hh -> GPUStep.hh
- GPUHit -> GPUStep
- FillHit -> FillGPUStep
- RecordHit -> RecordGPUStep
- adept_scoring -> adept_step_recording
- HitScoringBuffer -> DeviceStepBufferView
- gHitScoringBuffer_dev -> gDeviceStepBuffer
- HitQueueItem -> GPUStepBatch
- HitScoring -> GPUStepTransferManager
- HandleReturnedGPUHitBatchesWith -> HandleGPUStepBatchesWith
- GetGPUHitsFromBuffer -> GetGPUStepBatchFromBuffer
- ProcessReturnedGPUHits -> ProcessReturnedGPUSteps
- IsHitsFlushed -> AreReturnedStepsFlushed
- RequestHitFlush -> RequestStepFlush
- SwappingHitBuffers -> SwappingStepBuffers
- FlushingHits -> FlushingSteps
- HitsFlushed -> StepsFlushed
- hitBufferOccupancy -> stepBufferOccupancy
- scoringCapacity / fScoringCapacity -> stepCapacity / fStepCapacity
- hitBufferSafetyFactor / fHitBufferSafetyFactor -> stepBufferSafetyFactor / fStepBufferSafetyFactor

The existing UI command names are kept unchanged in this PR,  like `/adept/setMillionsOfHitSlots`, since those are user-facing and would break all current input scripts.

No behavior change is intended here. This is only the vocabulary cleanup before splitting the returned-step transfer/buffer code out of `PerEventScoringImpl.cuh`.

### DIFF
--- a/include/AdePT/core/AdePTTransportConfig.hh
+++ b/include/AdePT/core/AdePTTransportConfig.hh
@@ -12,7 +12,7 @@ struct AdePTTransportConfig {
   uint64_t adeptSeed{1234567};
   unsigned short numThreads{0};
   unsigned int trackCapacity{0};
-  unsigned int scoringCapacity{0};
+  unsigned int stepCapacity{0};
   int debugLevel{0};
   int cudaStackLimit{0};
   int cudaHeapLimit{0};
@@ -23,7 +23,7 @@ struct AdePTTransportConfig {
   std::string bfieldFile{};
   double cpuCapacityFactor{2.5};
   double cpuCopyFraction{0.5};
-  double hitBufferSafetyFactor{1.5};
+  double stepBufferSafetyFactor{1.5};
 };
 
 #endif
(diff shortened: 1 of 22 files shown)


agheata (approved):
Thanks for the more consistent naming!

PR #606, opened 2026-04-30
Extract host circular buffer

SeverinDiederichs:
This PR extracts the host circular buffer for the GPUSteps from the `PerEventScoringImpl.cuh`

It is moved into its .cpp files, as it is pure CPU code. A unit test is added, testing for multi threaded access and other functionalities.

Note: this PR does not change any behavior, it is on purpose split into a pure extraction.


agheata (approved):
Thanks for this cleanup!

PR #607, opened 2026-04-30
Clean up HostCircularBuffer

SeverinDiederichs:
Clean up the `HostCircularBuffer` API after the extraction in the previous PR.

The buffer now works in terms of offsets instead of raw `GPUStep*` pointers. This makes the class independent of the GPU step type and keeps the pointer arithmetic where the actual host buffer is.

This is a pure cleanup and does not change the implementation itself


agheata (approved):
Thanks for the clean-up, LGTM

PR #608, opened 2026-05-04
Split PerEventScoringImpl and name correctly

SeverinDiederichs:
Split the returned GPU-step transfer code into smaller headers with clearer responsibilities.

This moves the device buffer view and recording helper out of the transfer-manager implementation, and renames the remaining header from the old scoring-oriented name to `GPUStepTransferManager.cuh`.

Changes:
- add `DeviceStepBuffer.cuh` for the device-side step buffer view and comparison helper
- add `GPUStepRecording.cuh` for `RecordGPUStep`
- rename `PerEventScoringImpl.cuh` to `GPUStepTransferManager.cuh`

This is intended as a cleanup only. The transfer behavior is not changed.


agheata (approved):
Thanks for the separation and renaming, looks good!

Issue #609, opened 2026-05-04
Reorganize file structure by AdePT::Transport and AdePT:G4Integration libraries


SeverinDiederichs:
Following #516, where AdePT was restructured into two libraries, this issue documents how to leverage the library split into a cleaner file tree:

The final file tree could look like this:

```
AdePT/
  include/
    AdePT/

     ### NOTE: core/AdePTConfiguration.hh and integration/AdePTTrackingManager.hh
     # are temporary to maintain backward compatibility with the user integration for a bit.
      core/
        AdePTConfiguration.hh              # temporary compatibility forwarding header
      integration/
        AdePTTrackingManager.hh             # temporary compatibility forwarding header
        G4EmStandardPhysics_AdePT.hh   # temporary compatibility forwarding header
        G4EmStandardPhysics_HepEm.hh # temporary compatibility forwarding header

     ### Actual structure starts here: ###

      g4integration/
        AdePTConfiguration.hh
        AdePTTrackingManager.hh

      transport/
        config/
          AdePTTransportConfig.hh

        support/
          Macros.h
          Global.h
          Portability.hh
          ResourceManagement.hh
          ResourceManagement.cuh
          PhysicalConstants.h
          SystemOfUnits.h
          AdePTPrecision.hh

        containers/
          Atomic.h
          BlockData.h
          MParray.h
          MParrayT.h
          SlotManager.cuh
          mpmc_bounded_queue.h
          VariableSizeObj.h
          VariableSizeObjAllocator.h

        random/
          Ranluxpp.h
          G4HepEmRandomEngineDeviceImpl.hh
          ranluxpp/
            helpers.h
            mulmod.h
            ranlux_lcg.h

        g4hepem/
          AdePTG4HepEmState.hh

        tracks/
          TrackData.h
          Track.cuh
          ParticleTypes.hh

        queues/
          TrackBuffer.hh
          ParticleQueues.cuh
          ParticleManager.cuh

        steps/
          GPUStep.hh
          HostCircularBuffer.hh
          DeviceStepBuffer.cuh
          GPUStepRecording.cuh
          GPUStepTransferManager.cuh

        geometry/
          GeometryAuxData.hh
          WDTData.hh

        state/
          EventState.hh
          GPUState.cuh
          TransportStats.hh
          DeviceGlobals.cuh

        navigation/
          AdePTNavigator.h
          BVHNavigator.h
          LoopNavigator.h
          SurfNavigator.h

        magneticfield/
[...]

PR #610, opened 2026-05-04
Move library sources into target folders

SeverinDiederichs:
This PR is the first step towards the reorganization described in #609.

It only moves the source files, which currently live all in `/src/` even if they belong to different libraries.


agheata (approved):
Perfect, LGTM!

PR #612, opened 2026-05-04
Move headers into transport and g4integration layout

SeverinDiederichs:
Following #609, this PR reorganizes the headers into the transport and g4integration libraries.

After the sources have been moved in #610, this moves the headers.

To allow for a compatibility period with users, the publicly-used headers of the `AdePTConfiguration`, `AdePTTrackingManager`, and the physics constructors, are kept as a forwarding layer to the new design.

They will be deleted as soon as LHCb, ATLAS, CMS, and dd4hep are moved to the new design.

In the restructure, `copcore` was dissolved and the components moved to the correct part. Note that the `copcore` namespace was kept, this should be cleaned later.

The documentation was updated: The integration of AdePT is better described and it uses the new path for the headers.


agheata (approved):
No pending issues, thanks for the big restructuring step!

PR #614, opened 2026-05-05
Split AsyncAdePTTransportStruct headers into topical headers

SeverinDiederichs:
Split the old `AsyncAdePTTransportStruct.*` headers into smaller topical headers.

This is intended as a cleanup only: the transport data structures are grouped by responsibility, while runtime behavior should remain unchanged. The split makes the async transport state easier to navigate and reduces the amount of unrelated CUDA-side state pulled in through one broad header.

The following files were removed / renamed
- `include/AdePT/transport/AsyncAdePTTransportStruct.cuh`
- `include/AdePT/transport/AsyncAdePTTransportStruct.hh`
into
- `include/AdePT/transport/queues/ParticleManager.cuh`
- `include/AdePT/transport/queues/ParticleQueues.cuh`
- `include/AdePT/transport/state/DeviceGlobals.cuh`
- `include/AdePT/transport/state/GPUState.cuh`
- `include/AdePT/transport/state/TransportStats.hh`

Further cleanup:

As part of the split, avoid defining reusable symbols directly in topical headers:

- `gMagneticField` is now declared in `DeviceGlobals.cuh` and defined once beside the other device globals in `AsyncAdePTTransport.cuh`.
- `GPUstateDeleter::operator()` is now defined once in `src/transport/AsyncAdePTTransport.cu`.

This keeps the new headers safer to reuse from additional translation units.


agheata (approved):
Looks very good!

PR #615, opened 2026-05-05
Add unit tests for external fields with covfie

SeverinDiederichs:
This PR switches the `physics_drift_bfield` test to use an external covfie field map for the test, instead of using a constant B field, because the external fields are used in most production cases.

For this, a covfie B field is generated on the fly.

Furthermore, two unit tests are added:

1. `test_covfie_field` checks the Geant4-side field wrapper from `examples/common/include/MagneticFields.hh`.
It creates a tiny constant covfie `.cvf` file, loads it through CovfieField, and compares its returned values against UniformField at:
- the origin
- a far outside point, to exercise the clamped CPU Covfie field path

2. test_general_magnetic_field tests the GPU field in AdePT.
It creates a tiny constant Covfie .cvf file, loads it through `GeneralMagneticField`, copies that to the device, and launches a CUDA kernel that evaluates it at several positions. In the same kernel it evaluates a `UniformMagneticField` with the same stored float value, then compares the returned components exactly on the host.


agheata (approved):
Very good, LGTM!

PR #616, opened 2026-05-05
Split GeometryAuxData and Woodcock tracking data

SeverinDiederichs:
Following #609, this splits the GeometryAuxData and the Woodcock tracking data.


agheata (approved):
All good, LGTM!

PR #622, opened 2026-05-06
Modernize include guards

SeverinDiederichs:
This PR cleans the include guards and modernizes them:

Instead of using `#ifndef FILENAME_HH` it uses `#pragma once`

It is a pure mechanical cleanup.


agheata (approved):
Looks clean, good to go!

PR #626, opened 2026-05-07
Remove Async from AdePTTransport naming

SeverinDiederichs:
As there is no synchronous AdePT backend anymore, all the AsyncAdePT namings were redundant, as described in #609.

This PR cleans the names by removing all the `Async` from the file names. The `AsyncAdePT` namespace was replaced by `adept::transport` and the `async_adept_impl` was replaced by `adept::transport::detail`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -491,8 +491,8 @@ endif()
 #----------------------------------------------------------------------------#
 set(ADEPT_CORE_SRCS
   src/transport/g4hepem/AdePTG4HepEmState.cpp
-  src/transport/AsyncAdePTTransport.cc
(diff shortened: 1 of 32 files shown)


agheata (approved):
Looks consistent, there are just a couple of commented includes of `electrons_async.cuh` and `gammas_async.cuh` leftover in `AdePTTransport.cuh`, but those can be removed in a different PR. LGTM

SeverinDiederichs:
> Looks consistent, there are just a couple of commented includes of `electrons_async.cuh` and `gammas_async.cuh` leftover in `AdePTTransport.cuh`, but those can be removed in a different PR. LGTM

Good point, those files don't exist for long time anymore.. Will do a separate cleanup, thanks!

PR #627, opened 2026-05-08
cleanup leftover of async naming

SeverinDiederichs:
This PR follows #626 and cleans the remaining usage of the `async` naming in AdePT:

- Some commented-out includes of non-existing asynchronous kernels were deleted
- The CI modes were named `async | split | mixed`, although all of them where async. This is changed to `mono | split | mixed`. Note that mixed is also using the monolithic kernels but mixed precision.


agheata (approved):
LGTM

PR #629, opened 2026-05-12
Set unit test CMake targets to corresponding AdePT library

SeverinDiederichs:
This PR cleans the unit tests  so tests link only what they actually need.
There by, the unit tests are grouped into header-only, `AdePT::Transport`, and `AdePT::G4Integration`.
The artificial `AdePT::Core` was removed.


SeverinDiederichs:
@codex review

agheata (approved):
Looks good!

PR #631, opened 2026-05-13
Add EM3 validation test for FinishParticlesOnCPU option

SeverinDiederichs:
This PR adds an em3 validation test for the FinishParticlesOnCPU option.

It is run only in the `run-test-full` and the `nightlies`, as a drift test does not make sense, as the option is not physics reproducible.


SeverinDiederichs:
@codex review

PR #632, opened 2026-05-13
fix bug for returning FinishOnCPU steps

SeverinDiederichs:
This PR fixes a bug reported by @codex review in #540:

For the FinishOnCPU option, the step defining process is set after the real interaction. That means, steps that return with FinishOnCPU may have energy deposition or secondaries attached. In that case, the step replay must not be skipped.


SeverinDiederichs:
@codex review

agheata (approved):
Thanks codex ;-)

PR #633, opened 2026-05-13
set RNG state in constructor

SeverinDiederichs:
This small PR fixes a double `SetSeed` call in Ranlux++, which is very expensive:

Before, the rngState was default constructed (using the default seed: ` RanluxppDouble(uint64_t seed = 314159265) { this->SetSeed(seed); }`, and then the `SetSeed` was called again. 

This is in fact the most compute heavy part of the `InitTracks` kernel, which takes in Athena ~5-8% of all GPU compute time.
With this fix, the time of the kernel is roughly halved (at least when it is called with a proper number of blocks).


agheata (approved):
LGTM

agheata:
good catch!

PR #635, opened 2026-05-15
Overlap physics kernels with relocation

SeverinDiederichs:
This PR improves the handling of streams for the split kernels.

Currently, all split kernels run back to back: `ElectronsHowFar`, `ElectronsPropagation`, `ElectronMSC`, `ElectronRelocation`,  `ElectronIonization`, `ElectronBremsstrahlung`, (`PositronAnnihilation`), (`PositronStoppedAnnihilation`).

With this PR, the physics kernels for electrons, positrons, and gammas, run in a separate stream in parallel to the `Relocation` kernel, as particles either undergo an interaction or relocate.

This change is clearly visible with nsys, as the kernels now run in parallel.

As the physics kernels typically take only a small fraction of the compute time, only a modest speedup of 1-2% was observed, which coincides with the fraction of the physics kernels for electrons, which are the overall slowest particle species. (They do take a larger fraction for the gamma kernels, but as overall electrons are dominating (at least in ATLAS), the faster gamma propagation is completely hidden, as electrons and gammas run in parallel).
adept_profile_summary_wdt_on_woodcock_separate

### DIFF
--- a/include/AdePT/transport/AdePTTransport.cuh
+++ b/include/AdePT/transport/AdePTTransport.cuh
@@ -648,6 +648,12 @@ std::unique_ptr<GPUstate, GPUstateDeleter> InitializeGPU(int trackCapacity, int 
 
     ADEPT_DEVICE_API_CALL(StreamCreate(&particleType.stream));
     ADEPT_DEVICE_API_CALL(EventCreate(&particleType.event));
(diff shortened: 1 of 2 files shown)


SeverinDiederichs:
@codex review

PR #636, opened 2026-05-15
add nsys profiling hooks and documentation

SeverinDiederichs:
This PR adds optional CUDA profiler API hooks for profiling the transport loop with Nsight Systems, plus documentation for nsys and ncu workflows.

As the large production work flows can be difficult to profile, nsys hooks are introduced to only profile the transport loop and avoid geometry startup.

To enable it, compile with `-DADEPT_ENABLE_NSYS_PROFILING=ON` and run with `ADEPT_NSYS_CAPTURE_TRANSPORT`.
The variables `ADEPT_NSYS_CAPTURE_START_AFTER_ITERATIONS` and `ADEPT_NSYS_CAPTURE_STOP_AFTER_ITERATIONS` can be used to scope the profiling as otherwise the profiler can crash with too large events and >100k iterations.

The full behaviour is documented in the RTD documentation. Tested both in Athena and Gauss.


SeverinDiederichs:
@codex review

PR #637, opened 2026-05-15
Run Woodcock kernels in parallel

SeverinDiederichs:
This PR runs the WDT kernels in parallel to the normal gamma kernels, both for the split kernels and the monolithic kernels.

For this, the `physicsStream` from #635 is used to also cover the WDT kernels. As the name was not appropriate anymore, it was changed to `auxilaryStream`.
Additionally, the queues were renamed:
The split-kernel extra queues are now named:

- ParticleQueues::splitQueues
- SplitQueues
- ParticleQueues::numSplitQueues

Queue slots now have explicit names:

- gammaConversion
- gammaCompton
- gammaPhotoelectric
- gammaWoodcock
- chargedIonization
- chargedBremsstrahlung
- positronAnnihilation
- positronStoppedAnnihilation
- relocation

Furthermore, now the WDT kernels are launched with the correct size, and so are the other gamma kernels.

This PR accelerates LHCb by ~10% for the split kernels, for ATLAS it does not matter, as the WDT kernels run in parallel to the bottlenecking electron kernels.


SeverinDiederichs:
@codex review

PR #641, opened 2026-05-19
record steps for tracks that are finished on CPU first

SeverinDiederichs:
This fixes a crash observed in CMSSW with the FinishOnCPU option.
The bug is the following in the monolithic kernels:

As the FinishOnCPU is evaluated in the SteppingAction at the end of the kernel (as opposed the beginning in the split kernels), a step flagged with FinishOnCPU may have generated a secondary. Then, as returned tracks are deferred to the end for reproducibility, the step where the secondary is generated, and hence, where its HostTrackData is created, is processed very late. Until then, the secondary that was generated on the GPU in the last step, may have done another step itself. When this step is processed, the HostTrackData is still missing (the step to generate it, is in the DeferredQueue), leading to the crash.

The correct fix is to check for FinishOnCPU in the beginning of the kernel, such that a deferred step cannot have a secondary attached.

Furthermore, the per kernel printouts of finishing particles on CPU are removed, now a single printout on CPU per event is written, guarded by the AdePT verbosity.

### DIFF
--- a/include/AdePT/transport/kernels/electrons.cuh
+++ b/include/AdePT/transport/kernels/electrons.cuh
@@ -125,6 +125,46 @@ static __device__ __forceinline__ void TransportElectrons(ParticleManager &parti
     printErrors = !gTrackDebug.active || verbose;
 #endif
 
+    // Finish-on-CPU must happen before any interaction can create secondaries on the GPU.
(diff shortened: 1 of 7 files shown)


SeverinDiederichs:
@codex review

agheata (approved):
The fix is OK, LGTM!

PR #642, opened 2026-05-20
Remove static usage of kMaxThreads

SeverinDiederichs:
This PR removes the fixed `kMaxThreads` dependency from the transport loop and makes the relevant counters runtime-sized from the Geant4 worker count.

This PR should fix the nightlies, which ran out of stack in debug mode when the kMaxThreads was increased to 512 to accommodate some larger runs on H100 GPUs.

The fixed-size per-event arrays in `Stats` are replaced with device pointers sized at GPU initialization.
Thereby, the population counting and reset kernels are updated to use `numThreads` and dynamic shared memory.
Note that now the flags for the FinishOnCPU must be copied and cannot be passed as a fixed size array anymore. This copy should be negligible.


SeverinDiederichs:
@codex review

agheata (approved):
Looks good, LGTM!

PR #645, opened 2026-05-27
Include VecGeom CudaManager where used

agheata:
This MR adds direct `VecGeom/management/CudaManager.h` includes in CUDA translation units that use `vecgeom::cxx::CudaManager`.

This avoids relying on transitive VecGeom includes, which changed in recent VecGeom development branches.
This must have zero impact on physics, since it is just a header include.

---


SeverinDiederichs (approved):
Thanks for the cleanup!

PR #646, opened 2026-06-02
remove redundant path for track lineage info

SeverinDiederichs:
This PR cleans up a redundant path:
The lineage info is assigned once, in the step of the primary that creates the secondaries. There should be no track info attached at this point, therefore an error is thrown in case it is already attached.

Found by @WitekPokorski


SeverinDiederichs:
@codex review

WitekPokorski (approved):
Looks good to me. Thanks!

PR #647, opened 2026-06-09
remove geometry-related pushes in AdePT, keep one for the GPU-CPU handoff

### LINKED ISSUE #543: Remove all geometry-related pushes in AdePT
There are still some leftovers where tracks were pushed "to avoid issues on boundaries when the tracks are located within Geant4". However, as no tracks are located in Geant4 anymore, but the navigation states are always cleanly converted from VecGeom to Geant4 and vice versa, none of these pushes are needed anymore.

Therefore, all pushes should be removed.

SeverinDiederichs:
This PR resolves #543. 

In AdePT, previously some pushes were used to 
a) help VecGeom avoiding stuck tracks
b) when handing tracks from GPU in VecGeom back to G4 on CPU, to avoid that particles get located in a GPU region by G4, leading to a infinite loop of particles bouncing between GPU and CPU.

For a) in fact it is harmful, because the push provided via the interface: https://gitlab.cern.ch/VecGeom/VecGeom/-/blob/master/VecGeom/navigation/BVHNavigator.h?ref_type=heads#L572
is always applied and not only when the particle is on boundary. This could lead to an error, because if the push was leading to a step of 0, (because the track was close to boundary), then the position was not updated by the push, despite it was the push that led to the step of 0 (see [here](https://gitlab.cern.ch/VecGeom/VecGeom/-/blob/master/VecGeom/navigation/BVHNavigator.h?ref_type=heads#L599-L602)). This was a problem, as then for larger pushes (AdePT was using `1000 *kTolerance`), the particle could still be outside of the next volume, despite being relocated to that one.

b) in fact, the track is still located within G4 when it is initialized in G4HepEm [here](https://github.com/mnovak42/g4hepem/blob/master/G4HepEm/G4HepEm/src/G4HepEmTrackingManager.cc#L276). UPDATE:
Therefore, these pushes are unfortunately necessary:
As it was found, there are some rare, but occurring cases, where a track is located in G4 (correctly), but then it does a 0 step and hits the GPU volume again, as when the `G4Navigator` is freshly initialized, it does not have a LastExitedVolume state set, leading to the bouncing between G4 on CPU and VecGeom on GPU. Without being able to set the LastExitedVolume in the `G4Navigator` a push is needed when handing off a track from GPU to CPU. Here, we push by `100 * kTolerance`. 

Due to the actual changes in b), this needs thorough testing in Athena, CMSSW, and Gauss, before this PR should be merged:
Tested in:

- [x] Athena
- [x] CMSSW
- [x] Gauss

### DIFF
--- a/include/AdePT/transport/AdePTTransport.cuh
+++ b/include/AdePT/transport/AdePTTransport.cuh
@@ -193,7 +193,6 @@ __global__ void InitTracks(adeptint::TrackData *trackinfo, int ntracks, Particle
                            const vecgeom::VPlacedVolume *world, adept::MParrayT<QueueIndexPair> *toBeEnqueued,
                            uint64_t initialSeed)
 {
-  // constexpr double tolerance = 10. * vecgeom::kTolerance;
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < ntracks; i += blockDim.x * gridDim.x) {
     const auto &trackInfo = trackinfo[i];
 
--- a/include/AdePT/transport/kernels/electrons.cuh
+++ b/include/AdePT/transport/kernels/electrons.cuh
@@ -59,7 +59,6 @@ static __device__ __forceinline__ void TransportElectrons(ParticleManager &parti
                                                           AllowFinishOffEventArray allowFinishOffEvent,
                                                           const bool returnAllSteps, const bool returnLastStep)
 {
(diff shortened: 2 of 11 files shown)


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
Looks good!

PR #648, opened 2026-06-10
set nSteps correctly when tracks from GPU

SeverinDiederichs:
This PR fixes the step number of the G4Track when a track is returned from the GPU to the CPU.

It enables to either increment the StepNumber in a bulk (available in this MR in G4: https://gitlab.cern.ch/geant4/geant4-dev/-/merge_requests/5768) or falls back to a simple for loop.

When testing, the compiler actually optimizes away the simple for loop in release builds, so that no performance penalty is there, even for the older fall-back. 

Note that this is an important fix, because in case particles get stuck in a loop between CPU to GPU, before, the stepNumber was always reset to one when returned from GPU. This way, an infinite loop can occur, which is otherwise prevented, when particles are getting killed after N number of steps. With the fix, these infinite loops should not be possible anymore, as particles should get killed eventually, when their step number gets too high.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -372,6 +372,20 @@ block()
       return 0;
     }
     ]] AdePT_HAS_G4VTRACKINGMANAGER)
+
+  # Check whether G4Track can increment the current step number by nSteps
+  check_cxx_source_compiles([[
+    #include "G4Track.hh"
+
+    int main() {
+      auto increment = static_cast<void (G4Track::*)(G4int)>(&G4Track::IncrementCurrentStepNumber);
+      (void)increment;
+      return 0;
+    }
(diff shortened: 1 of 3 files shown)


SeverinDiederichs:
@codex review

SeverinDiederichs:
Note that the failing of the drift test for the regions is expected, as this PR fixes the number of steps when being returned to the CPU, and when particles go back to the GPU, [their step number is used to seed the RNG](https://github.com/apt-sim/AdePT/blob/master/include/AdePT/transport/AdePTTransport.cuh#L162), so a change in all histograms is expected.

agheata (approved):
Looks good, no findings, LGTM!

PR #649, opened 2026-06-11
fix minor errors in the B field, replace old unit test with proper one

SeverinDiederichs:
This PR does two things:

- it replaces the previous B field unit test, which was not really strict (it returned true for many parts regardless of the result), with a new, more consistent B field unit test. The new B field unit test uses GoogleTest, in a similar style to how VecGeom adopted it.
- In the generation of the unit tests, three small bugs were found in the B field, which are fixed, in order to pass the new B field unit test.

The new unit tests now covers:
- UniformMagneticField
- MagneticFieldEquation
- RkIntegrationDriver
- float/double RK integration vs analytic helix
- fieldPropagatorRungeKutta failure and navigation handoff behavior

The bugs were the following:
1. Previously, when IntegrateStep rejected a substep, it did:

`xCurrent = 0.0;`
That is wrong after any earlier accepted substep. The state vector `yStart` had already advanced, but the cumulative distance was reset. This could desynchronize the accepted state from the progress bookkeeping, causing wrong retry sizing and possible overshoot.

Fix: rejected substeps now only compute a smaller `hnext`; accepted cumulative progress is preserved.

2. RK failure could normalize a zero chord, leading to a division by 0:
`fieldPropagatorRungeKutta` called `RkDriver_t::Advance`, ignored its return value, then computed and normalized:

```c++
chordDir = endPosition - position;
chordLen = chordDir.Length();
chordDir *= 1.0 / chordLen;
```
If RK failed all trials, `endPosition` could remain equal to position, giving `chordLen == 0`. Normalizing then injected NaNs.

This is fixed by checking before the normalization.

3. Partial RK progress was treated as a completed chord

`RkIntegrationDriver::Advance` could exhaust its trial budget after accepting one or more valid substeps. In that case the returned endpoint was a valid partially advanced state, but `fieldPropagatorRungeKutta` had no way to know how much arc length had actually been accepted.

As a result, the propagator could use the partially advanced endpoint but still bookkeep the move as the full requested `safeArc`:

```cpp
move = safeArc;
remains -= move;
```
This could make propagation skip the remaining part of the curved arc and advance the transport bookkeeping farther than the RK integration had actually progressed.
This is fixed such that the Advance method now returns the actual arcAdvance and using that variable to progress.

Last, a minor underflow check was removed, which should be totally irrelevant and in any case, was not handled correctly.


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
Thanks for the tests, green from my side, but we should not forget to disallow the broken ADEPT_USE_SURF=ON in a separate PR; this won't even compile with it

SeverinDiederichs:
> Thanks for the tests, green from my side, but we should not forget to disallow the broken ADEPT_USE_SURF=ON in a separate PR; this won't even compile with it

You are right, I added the surface model overloads. This way we are consistent with the rest of the code and the tests compile, despite the surface model currently not being in a production state.

PR #650, opened 2026-06-12
Make Woodcock tracking compiletime option

SeverinDiederichs:
This PR makes the Woodcock tracking a compile time option (default on), as it currently is not compatible with the VecGeom surface model, which does not support certain required features in the navigator.

A little bit of cleanup is done by further using the `enterWDTRegion = ShouldUseWDT(nextState, eKin);` in the kernels.

Additionally, the surface model is compiled in the CI (at least for `/run-test-full` and the nightlies), and the bfield test is run.


SeverinDiederichs:
@codex review

agheata (approved):
Looks good from my side, LGTM

PR #652, opened 2026-06-15
update G4VG to 1.0.7

SeverinDiederichs:
This updates the G4VG to 1.0.7 which includes a fix on the tessellated solid. With this and the latest VecGeom version, tessellated solids work in AdePT.


agheata (approved):
LGTM

PR #653, opened 2026-06-15
retry CI in case of bad connection error from github

SeverinDiederichs:
In case of a bad connection, the self-hosted CI runner would just fail, as it did here: https://github.com/apt-sim/AdePT/actions/runs/27555965206

This PR strengthens the setup by retrying up to 5 times with increased time intervals before failing.


SeverinDiederichs:
@codex review

agheata (approved):
LGTM

PR #655, opened 2026-06-16
improve handling of safety in B field propagation

SeverinDiederichs:
This PR improves the handling of the safety in the B field propagation.

- It makes the names intuitive (`safetyAtOrigin`, `safetyAtChordStart`), and removes a redundant variable (`newSafety`)
- Before, it was overly conservative, as the `currentSafety` (now `safetyAtChordStart`) was actually calculated with the `EndPosition`, which makes it significantly smaller than it should be, so a safety had to be `2* chordlen` to get away without geometry checks. To clear this up further, the calculation of the `safetyAtChordStart` is moved to the beginning of the loop.
- A unit test is added to prevent the case above
- all unit tests for the B field have a better description.
- The logic of recomputing the safety is simplified.


SeverinDiederichs:
@codex review

agheata (approved):
Looks clean!

PR #656, opened 2026-06-16
cleanup unused code in FieldPropagatorRungeKutta

SeverinDiederichs:
This removes some unused code for alternative treatments at the boundary. As this might end up with steps that don't end up exactly on boundary, this is dangerous and fragile, so this PR cleans this unused code.


agheata (approved):
Good to go!

PR #657, opened 2026-06-16
fix safety in split kernels

SeverinDiederichs:
This fixes a bug regarding the safety in the split kernels.

In the split kernels, the field propagator gets directly the `currentTrack.safety` value as an input. However, in the setup of `ElectronHowFar` the safety value is not always refreshed based on the current position (it is only refreshed in the track if it is recomputed). The fix is straightforward to always write back the safety to the `currentTrack` in the split kernels, such that the field propagator always gets a safety / position that is in sync.


SeverinDiederichs:
@codex review

agheata (approved):
Looks good, but we should add a test in a future PR

PR #659, opened 2026-06-18
Fix 0 step B field loop

SeverinDiederichs:
This PR fixes a zero-step B-field navigation failure where the RK chord could report an immediate boundary crossing while the track direction (not the chord direction) still belonged to the current volume. This happened for tiny first steps at volume boundaries and could make tracks bounce between parent/daughter boundary states until killed by `zeroStepCounter`.

The fix classifies the ambiguous first-step boundary case more carefully:
- only applies the special handling when both current and chord states are on a boundary, but increases the tolerance to trigger.
- probes the physical direction for tiny boundary distances
- reduces chord-only boundary artifacts instead of accepting the wrong boundary state
- preserves genuine immediate boundary relocations/backscatter as zero-length crossings
- keeps finite tiny crossings at the navigator-reported distance (so far, this does not seem to be triggered)

This also removes the unused `zero_first_step` output and moves the RK/navigation handoff coverage into `test_bfield_navigation.cpp`.
Further unit tests are added to make sure the behaviour cannot be reintroduced.

In running 1k ttbar events in CMSSW, the zero-step stuck tracks went down from 77 to 0 with no performance difference.
Running 5k min bias events in Gauss, it went down from 72 stuck tracks to 0, a performance regression in the noise (<1%)


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
Excellent outcome! Good to go. Funny that we did not spot this after several passes on the field code.

PR #660, opened 2026-06-19
remove legacy testField, move tests to new unit tests

SeverinDiederichs:
This PR removes the legacy testField CI test, which was still originating from the time when AdePT was example based. Hence, it came with its own kernels and full GPU loop.

None of this is needed for the B field testing anymore and creates only legacy code, which has to be maintained when interfaces change.

Therefore, all of it is removed and the small part of relevant testing that was not yet covered, is moved to the new B field tests.


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
Good cleanup!

PR #662, opened 2026-06-19
add looper count as UI command, move bools into small struct

SeverinDiederichs:
Previously loopers in the B field were killed at a hardcoded count. This was obtained from using ATLAS simulations. However, this should be a UI command, as e.g., plenty of particles are killed in CMSSW (mostly low energy particles). 
Therefore it is important that this parameter is tunable, as in CMSSW a higher limit can be used, which has no performance penalty, but drastically reduces the amount of killed particles.

To clean the implementation a bit, the booleans are now passed as a small struct to the kernels, instead of invidivually.

### DIFF
--- a/include/AdePT/g4integration/AdePTConfiguration.hh
+++ b/include/AdePT/g4integration/AdePTConfiguration.hh
@@ -4,7 +4,9 @@
 #pragma once
 
 #include <cstdint>
+#include <limits>
 #include <memory>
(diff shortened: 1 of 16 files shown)


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
Thanks for the cleanup!

PR #663, opened 2026-06-22
Reduce reserve HostTrackDataMapper, move unit test to GTest

SeverinDiederichs:
Previously, the HostTrackDataMapper was reserving for 1'000'000 expected tracks, which was humongous. In fact, for 64 threads, that would lead to reserving ~10 GB of RAM (as this is done per thread).

Recent profiling has shown that in most production setups in LHCb, ATLAS, CMS, most events stay below 30k tracks in flight and rarely exceed that (but then sometimes they do so heavily). 
Therefore, this PR reduces the expected size to 30k. In case a large event came through, the reserved size is reduced to currently 4 times the initial 30k. This drastically reduces the memory footprint where it usually was not needed. 

Note that this high number of tracks can easily occur when benchmarking testEm3 or other setups, therefore this is a relevant use case, it should just not reserve so much memory per default.

Two more unit tests are added and the HostTrackData mapper unit test is moved to google test.

### DIFF
--- a/include/AdePT/g4integration/returned_steps/HostTrackDataMapper.hh
+++ b/include/AdePT/g4integration/returned_steps/HostTrackDataMapper.hh
@@ -49,6 +49,9 @@ struct HostTrackData {
 // such as the pointer to the creator process, the G4 primary particle, and the G4VUserTrackInformation
 class HostTrackDataMapper {
 public:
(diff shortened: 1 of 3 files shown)


SeverinDiederichs:
@codex review

agheata (approved):
Looks good!

PR #666, opened 2026-06-25
remove particle type from HostTrackData

SeverinDiederichs:
As the particle type is already returned for each step from the GPU, there is no need to store it in the HostTrackData.
Originally, it was the idea to return it only for the creation step, but as the size of the returned step does not change, it is better to keep it in the returned step, as otherwise it might not be available if the HostTrackData is not available. Since this is a very basic parameter, it should always be available.


SeverinDiederichs:
@codex review

agheata (approved):
The change looks consistent, LGTM!

PR #669, opened 2026-07-06
fix bug with replicated volumes in gdml path

SeverinDiederichs:
In case of the gdml path (not via G4VG), replicated volumes were not visited correctly in AdePT.
This bug is fixed and a unit test is added.

Later, it was found that the inverse was also not set correctly, so that the VecGeom volumes were not pointing to the correct copy number of the replica.

More unit tests were added to also cover this case.
Most of the lines added in this PR are unit tests.

This was not causing issues so far, as none of the consumers use replica volumes and read from gdml.


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

agheata (approved):
AI's/humans agree on this being OK to merge. LGTM!

PR #672, opened 2026-07-10
Fix PostUserTrackingAction in GPU handover

SeverinDiederichs:
When a track is given to the GPU after being tracked in G4HepEm, the PostUserTrackingAction should not be called. As the G4VUserTrackInfo is by now propagated with the HostTrackData, the PostUserTrackingAction will be called when the track is finished on the GPU with the returned step (or finally by the CPU, when the track returns and is finished on the CPU).

Thus, the call to the PostUserTrackInfo in the CheckEarlyTrackingExit must be removed to avoid a double call. 

Note that this PR changes the drift test output:

The PostUserTrackingAction in the drift test records the entry:
`truthHistogrammer->RecordFinalTrack(track);`
Therefore, the entries differ, as this was called multiple times per track.


SeverinDiederichs:
@codex review

agheata (approved):
Looks good!

PR #673, opened 2026-07-20
use correct minimum length for displacement

SeverinDiederichs:
Despite the comment saying 0.05nm and the [original implementation](https://github.com/mnovak42/g4hepem/blob/master/G4HepEm/G4HepEm/src/G4HepEmTrackingManager.cc#L560) using 0.05nm, AdePT used 5nm for the minimal displacement distance under which no displacement was applied.

That is fixed in this PR.

Drift test is expected to fail, as physics output changes with this PR.


agheata (approved):
Good catch, do you expect a big impact on ATLAS?

SeverinDiederichs:
> Good catch, do you expect a big impact on ATLAS?

No, 5nm was also very small

PR #674, opened 2026-07-20
clear SlotCounter and FreeCounter only in block 0

SeverinDiederichs:
Technically speaking, there is a data race in the `SlotManager::Clear()` call, as in every block, the threadidx 0 clears the SlotCounter and the FreeCounter. The function is called with 80 blocks [here](https://github.com/apt-sim/AdePT/blob/master/include/AdePT/transport/AdePTTransport.cuh#L847).

Therefore, this PR fixes it, and only the first thread in the first block clears the counters.


SeverinDiederichs:
@codex review

agheata (approved):
This was in practice  more resets of counters, but harmless, right?

SeverinDiederichs:
> This was in practice more resets of counters, but harmless, right?

Yes, it was a race condition which block sets it to 0 first, so harmless but should be fixed

PR #675, opened 2026-07-20
avoid binding a reference to nullptr

SeverinDiederichs:
This PR fixes a small bug: In the non-B field case, `gMagneticField` is a nullptr, thus the following line is undefined behavior:
```c++
auto &magneticField = *gMagneticField;
```
This is fixed by using the pointer directly.


SeverinDiederichs:
@codex review

agheata (approved):
Good cleanup, LGTM

PR #676, opened 2026-07-24
[mini] improve profiling plots

SeverinDiederichs:
This PR improves the visuals of the profiling plots.
Before, the colors between the same kernels were not the same for electrons/positrons and gamma. Also, the order was not the order they are called in. The labels are increased.

Before:
595624555-7523b439-e85b-409e-9049-481563a59536

After:
split_capped_species_fractions


agheata (approved):
Nice!

PR #677, opened 2026-07-24
allow returnOfTrack UI commands to be callable from G4_Idle

SeverinDiederichs:
This fixes the UI commands to also be callable from G4 idle. As we do the late initialization, these commands did not run in Gauss anymore, where the UI commands are invoked only later.


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

PR #678, opened 2026-07-24
[mini] fix caching of safety

SeverinDiederichs:
This PR fixes the caching of the safety:

1. when the safety is needed for the `HowFarToMSC` it is recomputed if it is below the `physicalStepLength`. However, it was stored back in the track independently whether it actually had been recomputed or not. In case it had not been recomputed, this re-centered the safety at the current position, causing the cached safety to shrink—or, at best, remain unchanged.

2. Something similar applied to the safety that comes out of the B field integration: Before, it was also shortened by re-centering. When passing the safety cache of the track directly to the integrator, it will already update the safety and the safety center to the correct point.

A physics drift is expected, as this will change the safety, and consequently the step length and everything else


SeverinDiederichs:
@codex review

PR #679, opened 2026-07-27
set gammas eKin to 0 after photoelectric effect

SeverinDiederichs:
The kinetic energy after the photoelectric absorption is 0, however it was only set to 0 if the secondary was cut due to ApplyCuts.

This could lead to an incorrect postStepPoint kinetic energy of the parent gamma.

As expected, the drift test fails, but only the single entry that was changed:
```
ROOT histogram comparison found mismatches: 1 failing checks across 42 common histograms.
- Histogram 'final_ekin' differs: bin count 2025 != 1089
```


SeverinDiederichs:
@codex review

PR #680, opened 2026-07-27
reduce inputs to RecordGPUStep

SeverinDiederichs:
This reduces the footprint of the RecordGPUStep function signature by taking the track as an input, which saves duplication of multiple parameters that are taken from the track anyway.

Note that in the split kernels it could be reduced significantly more, but to keep uniformity between monolithic and split kernels, it is not further reduced to also include the stepPoints etc, which are only member of the split kernel track structure, not in the monolithic kernel track structure.


SeverinDiederichs:
@codex review

Issue #681, opened 2026-07-27
Improve initialization of tracks


SeverinDiederichs:
Profiling has shown that a significant fraction of the compute time is spend in the Electron and Positron HowFar kernels, where they calculate the safety:

<img width="2972" height="1420" alt="Image" src="https://github.com/user-attachments/assets/8bafe0a0-9fbb-4523-b735-a8d48d982242" />

However, here it should be recalculated only if the safety is smaller than the physics step.

There are multiple ways to improve this:

1. Particles that are being offloaded from the CPU are initialized without safety. The `InitTracks` kernel is run fully asynchronously in the background, so this one could already calculate the safety of the track. This would reduce the calls to safety in the HowFar kernel.
2. Secondaries that are generated could also benefit from directly getting the safety: secondaries from charged particles via ionization could get the safety directly from their parents, but unfortunately, the large fraction of electrons from gamma compton scattering or photoelectric could not get it from their parents.
3. In the split kernels, one could have a separate HowFar for initializing particles. This kernel could then always calculate the safety, so there is no divergence. In this kernel, the initialization of the HepEm track could also happen, so this divergence:
```c++
    if (!currentTrack.hepEmTrackExists) {
      // Init a track with the needed data to call into G4HepEm.
      elTrack.ReSet();
      theTrack->SetEKin(currentTrack.eKin);
      theTrack->SetMCIndex(auxData.fMCIndex);
      theTrack->SetOnBoundary(currentTrack.navState.IsOnBoundary());
      theTrack->SetCharge(Charge);
      mscData->ReSet();
      currentTrack.hepEmTrackExists = true;
    } else {
      theTrack->SetEnergyDeposit(0);
      mscData->fIsFirstStep = false;
    }
```
Could also be removed. Then the InitTracks kernel would also have to do the hepemTrack initialization.

PR #682, opened 2026-07-27
initialize safety in InitTracks

SeverinDiederichs:
This PR implements the first step of #681: 

When initializing the tracks that come from the CPU on the device asynchronously, their safety is already calculated, thus reducing the amount of safety calculation in HowFar. Note that in this initial approach in a bad case when the new calculated safety is still smaller than the physics limit, the safety would be calculated both in the InitTracks and HowFar, but as the init tracks is fully hidden, it is considered negligible for this first implementation, as this will be changed in the subsequent steps of #681.

### DIFF
--- a/include/AdePT/transport/AdePTTransport.cuh
+++ b/include/AdePT/transport/AdePTTransport.cuh
@@ -33,7 +33,7 @@
 
 #include <AdePT/transport/tracks/TrackDebug.cuh>
 
-#include <AdePT/transport/navigation/BVHNavigator.h>
+#include <AdePT/transport/navigation/AdePTNavigator.h>
 
 #include <AdePT/transport/kernels/AdePTSteppingActionSelector.cuh>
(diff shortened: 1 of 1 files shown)


SeverinDiederichs:
@codex review

SeverinDiederichs:
The failure in the CI is bad luck:

We currently running 60k primaries with a threshold of rejection of 1% between the result of the PR and the high statistics result stored from G4HepEm.
Changing the seed can cause the setup to be unlucky to fail the 1%, despite being as accurate as before.
This PR shifts the physics output to this unlucky case, as the following table shows:

| Result                        | Master | This PR |
| ----------------------------- | -----: | ------------: |
| Validation passes             |  19/20 |         19/20 |
| Validation failures           |   1/20 |          1/20 |
| Aggregate maximum layer error | 0.244% |        0.219% |
| Mean per-run maximum error    | 0.574% |        0.585% |

Therefore, the number of primaries should be increased to make these failures appear less often

PR #683, opened 2026-07-27
move stepping action to beginning of kernel

SeverinDiederichs:
Previously, the SteppingAction in the monolithic kernels was done at the end of the step, while in the split kernels it was done at the beginning.

This caused an issue in the physics validation in LHCb when Woodcock tracking is turned on:

As even low energy gammas can travel a significant distance, crossing volume boundaries, when the MeV tracking cut in LHCb was applied only after the first step, this could lead to particles depositing energy in the ECAL that previously did not.

This is shown in the following plot: 
ecal_energy_weighted_subhits_before_after

Before this PR, there were many more low energy hits in AdePT. After, it agrees perfectly well with G4HepEm.

Note that this behavior of the large cuts most likely causes the differences in the first bin when comparing Gauss + G4 with its custom tracking cuts vs G4HepEm or AdePT:
energy_weighted_subhits_ecal


SeverinDiederichs:
@codex review

PR #684, opened 2026-07-27
[mini] Improve documentation for nsys usage

SeverinDiederichs:
When running with nsys, stalling was observed in running more sophisticated runs.
They come from the cuda events. Disabling the tracing of the cuda events fixes the problem, hence the documentation should reflect that.


SeverinDiederichs:
@codex review

agheata (approved):

PR #686, opened 2026-07-29
Set Velocity and Status in G4Step information

SeverinDiederichs:
This PR sets the velocity and the G4Status of the returned GPU steps in the persistent G4 steps correctly.

The issue was that the G4Status could be set by the NuclearReplay to fStopAndKill, which was then seen by the next track, causing issues in the MC truth in ATLAS.

Similarly for the velocity, particles undergoing nuclear replay had a velocity set, the other particles did not. This is now added for consistency.


SeverinDiederichs:
@codex review

PR #687, opened 2026-07-30
inherit safety from charged tracks

SeverinDiederichs:
In ATLAS, roughly 45% of all electrons are produced from electrons/positrons via ionization.
Previously, they were initialized with safety 0, meaning that the safety had to be recalculated.

Instead, the full safety cache, including the safetyOrigin point and the safety itself can be inherited, such that no re-calculation of safety is needed for those particles.

This improves the HowFar kernel by 7% in the full ATLAS geometry.


SeverinDiederichs:
@codex review

PR #689, opened 2026-08-13
fix returned process id

SeverinDiederichs:
Previously, the returned process ID was incorrect when the particle was stopped due to continuous energy loss.
This does not change physics itself, only the information on the step limiting process (so the drift test fails only in one histogram).

However, this could lead to a falsely classified process in the step reconstruction, which would again not change the physics, but it allows a stronger and correct assumption that particles that undergo nuclear reactions do not have any secondaries.

That avoids the previous additional (and costly) lookup
```
    HostTrackData &parentTDataAfterSecondaries = useHostData ? fHostTrackDataMapper->get(parentStep.fTrackID) : dummy;
```
because this code path was run by stopped positrons which secondaries, so the `parentTData` was invalidated due to the added secondaries and a new lookup was needed.
Fixing the return type, the additional lookup can be removed, as the original `parentTData` cannot be validated by adding new secondaries to the lookup map.


SeverinDiederichs:
@codex review

agheata (approved):
Came out green, LGTM!

PR #690, opened 2026-08-14
fix potential undefined behavior in hostCircularBuffer

SeverinDiederichs:
This PR fixes two bugs in the HostCircularBuffer:

1. There was potentially undefined behavior: It was possible that the iterator returned the `end()` and `erase(iterator)` was called. This is fixed by ensuring that the segment is only erased if it is found.
2. If the forward space + the wraparound space were smaller than the transfersize, the free space could have an unsigned underflow, returning a very large number, instead of 0.

Two unit tests are added to test this for the future.


SeverinDiederichs:
@codex review

agheata (approved):
LGTM!

PR #691, opened 2026-08-14
fix memory leak of VolAuxData on host

SeverinDiederichs:
The volume auxiliary data was allocated as a raw pointer and never freed in `InitializeSharedAdePTTransport()`:
```c++
  auto *auxData = new adeptint::VolAuxData[vecgeom::GeoManager::Instance().GetRegisteredVolumesCount()];
```

This is fixed by using a smart pointer on the host, and the view is cleared when the device is freed.


SeverinDiederichs:
@codex review

agheata (approved):
The ownership and teardown sequence is coherent: the transport owns the host array, the GPU worker is joined before device cleanup, the constant device pointer is cleared, and the singleton’s non-owning views are reset.

PR #692, opened 2026-08-14
fix minor bugs in data containers, add unit tests

SeverinDiederichs:
Multiple minor bugs in the data containers are fixed in this PR:

| Defect | Fix |
|---|---|
| Shrinking `VariableSizeObj` copy |Copy `min(new_size, other.fN)` elements. Leave initialization of additional capacity to the derived container. ) |
| `SizeOf(0)` | Avoid `(nvalues - 1)` wraparound and allocate at least enough storage for a complete `Cont` object. Containers such as the bounded queue should continue rejecting unsupported zero capacity. |
| Missing atomic assignment return |Add `return *this;`.  |
| Empty `MParrayT::back()` | Add `empty()`, assert the nonempty precondition, and return `*(end() - 1)`. The current zero-capacity alternative also returns an invalid reference. |
| `SlotManager` move assignment | Handle self-move, release the destination’s existing allocation before stealing, and reset both source pointers and its sizes/counters.|

Unit tests are moved to GTest and are added for the MPArray and SlotManager


SeverinDiederichs:
@codex review

agheata (approved):
LGTM

PR #695, opened 2026-09-03
make trackInAllRegions and removeGPURegion consistent

SeverinDiederichs:
There was a bug with the removeGPURegion and trackInAllRegion flags: if trackInAllRegions was used, the removeGPURegion would not work correctly in the offloading of the TrackingManager.

This PR fixes that and makes the implementation more robust, by keeping only one list of GPU regions and no possibly conflicting flags. Furthermore, as soon as the GPU regions are initialized on GPU, the flags are locked and one cannot change them with subcommands anymore, to avoid any stale configurations.

Unit tests are added to cover the cases.

### DIFF
--- a/include/AdePT/g4integration/AdePTConfiguration.hh
+++ b/include/AdePT/g4integration/AdePTConfiguration.hh
@@ -3,11 +3,13 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cstdint>
 #include <limits>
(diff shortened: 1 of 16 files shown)


SeverinDiederichs:
@codex review

agheata (approved):
No high-confidence issues, LGTM!

PR #697, opened 2026-09-09
Strengthen handling of secondaries from nuclear steps 

SeverinDiederichs:
This PR fixes the handling of secondaries from nuclear steps.

Before, a separate G4Track vector was populated by the secondaries from nuclear reactions, which was then put on the stack. However, in the UserSteppingAction, tracks can be deleted by the user (i.e., as they do in the Russian Roulette in CMSSW). In that case, the vector would point to invalid track objects.

This is fixed by stacking directly the vector that is also passed to the UserActions, thus, any changes during the user actions to the vector are passed correctly to Geant4. 
As nuclear reactions cannot have any other secondaries, it is safe to stack the entire secondary vector in case of a nuclear reaction.


SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

SeverinDiederichs:
@codex review

PR #698, opened 2026-09-11
[mini] disentangle UserStepping and UserTrackingAction

SeverinDiederichs:
This small PR disentangles a leftover from #665. 
Currently, the UserTrackingAction is called when eiter the CallUserTrackingAction or CallUserSteppingAction flag is set.

As the returning of steps was already disentangled, so should be the user actions.

### DIFF
--- a/src/g4integration/tracking_managers/AdePTTrackingManager.cc
+++ b/src/g4integration/tracking_managers/AdePTTrackingManager.cc
@@ -77,11 +77,6 @@ bool CallUserTrackingAction(const AdePTConfiguration &configuration)
   return configuration.GetCallUserTrackingAction();
 }
 
-bool CallUserActions(const AdePTConfiguration &configuration)
-{
-  return CallUserSteppingAction(configuration) || CallUserTrackingAction(configuration);
-}
-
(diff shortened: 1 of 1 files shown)


SeverinDiederichs:
@codex review

agheata (approved):
**Review by: Claude Sonnet 5**

Small, self-contained 13-line change in AdePTTrackingManager.cc: it removes a dead free function CallUserActions() (no other call sites, no header declaration) and renames callUserActions → callUserTrackingAction, tightening the gate on PreUserTrackingAction so it no longer fires when only stepping-action callbacks were requested.

I traced the full ProcessTrack function and the GetCallUserSteppingAction()/GetCallUserTrackingAction() OR-logic in AdePTConfiguration.hh — the already-disentangled stepping-action path (ProcessGPUStep/ProcessReturnedGPUSteps) is untouched, and the combined-flag case is unaffected; only the previously-buggy stepping-only case changes, matching the PR's stated intent. Docs and messenger text remain accurate.

No findings. This is a clean, correctly-scoped fix with no correctness, cleanup, or documentation issues.

